PR #28, opened 2015-03-16
Implement tessel init

HarleyKwyn:
Need example files so I can load them into the root for auto init but this should work now. :)

- [x] if there is no package.json it will make one with tessel as a keyword
- [x] installs any dependencies in your package.json
- [x] if there is no index.js, makes one with blinky in it


HarleyKwyn:
Fixes #28

PR #31, opened 2015-03-17
require fs properly

Frijol:
@johnnyman727 should I also add back that passkey line?

### DIFF
--- a/lib/deploy.js
+++ b/lib/deploy.js
@@ -52,7 +52,7 @@ function sftpDeploy(opts) {
             });
           });
         });
-      }); 
+      });
     });
   });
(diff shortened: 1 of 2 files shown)


johnnyman727:
Yes, this will have to pass in the provided passphrase to the `connect` method inside `ssh-deploy.js`

Frijol:
Is it not? What needs to change?

johnnyman727:
My mistake- you have it in there. Can you confirm that it works for keys without passphrases as well? Once you do that I think it's good to merge.

Frijol:
Works great! Merging.

PR #35, opened 2015-03-18
WIP Auto-authenticate Tessel with your computer

Frijol:


kevinmehall:
I'm not especially inclined to trust the `keypair` library after a quick read of the source. Incredibly [overcomplicated ways of doing important things like obtaining random numbers](https://github.com/juliangruber/keypair/blob/master/index.js#L758-1107) that can be done with [simple Node APIs](https://nodejs.org/api/crypto.html#crypto_crypto_randombytes_size_callback), and in general doing too much easy-to-subtly-break crypto in unaudited JS.

Most of the other libraries just shell out to `ssh-keygen`, which won't work on Windows, though...

package.json
```diff
@@ -1,41 +0,0 @@
-{
-  "name": "prime-cli",
```
johnnyman727: @Frijol Why are you deleting this file?

johnnyman727:
It seems like this should be broken up into two components before being merged:

1. Using a [postInstall](https://docs.npmjs.com/misc/scripts) script tag in the package.json, create `HOME/.tessel` if it doesn't exist and place a new SSH key there. That way, it's automatically created when the tessel package in installed.

2. Running `tessel setup --name OPTIONAL_NAME` will transfer that SSH key over to the Tessel and optionally set the hostname of the device.

@kevinmehall @Frijol thoughts?

Frijol:
@johnnyman727 that sounds appropriate

Frijol:
@johnnyman727 What's going on with this file organization? Should I move init and setup into the tessel folder? 
screen shot 2015-03-27 at 12 03 52 pm

johnnyman727:
I placed tessel specific commands in the `tessel` folder. Higher level CLI commands can be in the `lib` folder. Setup.js can stay where it is.

lib/setup.js
```diff
@@ -0,0 +1,59 @@
...
+function authTessel (pubKey) {
+  console.log('Authenticating Tessel with public key...');
+
+  var sourceFile = filepath + '.pub';
+  var authFile = '/etc/dropbear/authorized_keys';
+
+  // Copy pubKey into authFile
+  tessel.connection.exec(commands.appendFile(sourceFile, authFile), function (err, streams) {
+    err && console.log(err);
+    streams.stderr && console.log(streams.stderr);
```
johnnyman727: I think what you actually want to do here is pipe stdout and stderr to the console:
```.js
streams.stderr.pipe(process.stderr);
streams.stdout.pipe(process.stdout);
```
And then wait for the stream to close to know that the process has completed:
```
// Commands finished executing
 streams.stdin.once('close', function(code, signal) {
  // End the connection
  self.connection.end();
  // Exit the process
   process.exit(1);
});
```

johnnyman727: And just for some more information, the reason you don't want to do
```
streams.stderr && console.log(streams.stderr);
```
is because that will check if `streams.stderr` is truthy (which it always will be b/c it's an object) and then print out that object. Instead, we want to stream data to the console while the command is running and close the connection when the stream tells us it's done.

lib/setup.js
```diff
@@ -0,0 +1,59 @@
...
+    if(!err && !streams.stderr) {
+      console.log('Tessel authenticated with public key.');
+    }
+  });
+}
+
+// Sets the hostname of the Tessel
+function setHostname (name) {
+  // TODO: Any security we need to do surrounding hostnames?
```
johnnyman727: Security here is handled by SSH. You can only change the hostname if you have your SSH key on the device, or else you can't connect.

resources/postinstall.js
```diff
@@ -0,0 +1,31 @@
+var osenv = require('osenv');
+var fs = require('fs-extra');
+var keypair = require('keypair');
+
+var dir = osenv.home() + '/.tessel';
+var filename = 'id_rsa';
+var filepath = dir + '/' + filename;
+
+// If keypair doesn't already exist.
+// TODO: fs.exists is going to be deprecated, what's a better way to do this?
```
johnnyman727: I think you mean `fs.existsSync` is going to be deprecated. You can replace it with

```
fs.exists(filename, callback(exists))
```

resources/postinstall.js
```diff
@@ -0,0 +1,31 @@
+var osenv = require('osenv');
+var fs = require('fs-extra');
+var keypair = require('keypair');
+
+var dir = osenv.home() + '/.tessel';
+var filename = 'id_rsa';
```
johnnyman727: You might consider changing the name of this file so it's not the default name for an SSH key. It's a very minor point but it could be confusing if users are, for some reason, looking at two keys with the same name in different folders.

resources/postinstall.js
```diff
@@ -0,0 +1,31 @@
...
+    err && console.log(err);
+
+    // Put SSH keys for Tessel in that folder
+    fs.writeFile(filepath + '.pub', pair.public, function (err) {
+      err && console.log(err);
+      console.log('Public key written.');
+    });
+    fs.writeFile(filepath, pair.private, function (err) {
+      err && console.log(err);
+      console.log('Private key written.');
```
johnnyman727: We've been using ```tessel.logs.info``` to print to the console in other files. It adds color formatting and prepends "INFO" to the beginning of the string for a consistent user experience. 

You'll have to `require('tessel');` first.

README.md
```diff
@@ -29,9 +27,10 @@ The commands below are planned to be implemented by product ship date.
   * `up` start an access point with previous credentials
   * `down` stop an access point
 * `tessel logs` lists the recent activity, console.logs of running program from flash or RAM
+* `tessel setup` copies your RSA public key onto the Tessel so that you can SSH into that tessel
+  * `[--name NAME]` sets the hostname of the Tessel
```
johnnyman727: I would change the language so it's less Linux-y since we're abstracting away the OS. Something like:

```give the Tessel a unique name```

File: lib/setup.js [hunk 8]
```diff
@@ -0,0 +1,59 @@
...
+    err && console.log(err);
+    streams.stderr && console.log(streams.stderr);
+    if(!err && !streams.stderr) {
+      console.log('Tessel authenticated with public key.');
+    }
+  });
+}
+
+// Sets the hostname of the Tessel
+function setHostname (name) {
```
johnnyman727: In terms of code organization, I would add this function as a method of a Tessel object (so you can do `tessel.setName()` because it's an action that you would only take on a single Tessel. (see `deploy.js` or `erase.js` for how to do that).

Also, I'm curious what you think about exporting this functionality and adding it to the command line so someone could just change the name of their Tessel without calling `setup`? Something like `tessel christen NAME`. My thinking is that people may want to change the name of the Tessel more often than when they initially set it up. I suppose it doesn't do a whole lot different from `setup` if `setup` doesn't recreate the `.tessel` folder if it already exists.

kevinmehall:
It would be better to create `.tessel` and the SSH key on `tessel setup` if they don't exist, rather than on postinstall. Postinstall will lead to permissions issues if people `sudo` to install globally, or if we choose to install the package with a system package manager.

johnnyman727:
@kevinmehall why would you run into permissions issues on `postinstall` but not `tessel setup`?

johnnyman727:
Just for the record, we talking about this IRL and @kevinmehall explained that the issue would arise when someone runs `sudo npm install -g tessel` which sets root permissions on the file that gets created (and which users do all too often). By not using `postInstall`, we can avoid permission issues (unless someone does `sudo tessel setup`, at which point, they've probably got bigger issues).

lib/setup.js
Frijol: streams.stdin is the only one that needs to close? Not one each for stdout and stderr?

lib/setup.js
Frijol: More worried about name collisions or special characters or anything like that. Can the hostname be just any string?

resources/postinstall.js
Frijol: they are both going to be deprecated.

resources/postinstall.js
Frijol: Do you have a recommendation? tesselkey?

resources/postinstall.js
Frijol: Will this be the same tessel that we get back for the purposes of tessel.connection?

Frijol:
Replaced by https://github.com/tessel/t2-cli/pull/58 and https://github.com/tessel/t2-cli/pull/59

PR #37, opened 2015-03-19
Refactors Communication into Connection Abstraction

johnnyman727:
hierarchy:
`controller.js` - logic that manages connections to one or more Tessels over either USB or SSH
`tessel/` - code for all common Tessel interactions like scanning for networks and deploying code
`lan_connection.js` - exposes interface for executing commands and returning stdin, stdout, and stderr streams over SSH
`usb_connection.js` - exposes interface for executing commands and returning stdin, stdout, and stderr streams over USB
`commands.js` - returns the bash commands needed to control openWRT on T2

What this PR doesn't do: break bash commands args into an array


bin/prime.js
```diff
@@ -24,7 +21,7 @@ parser.command('run')
 
 parser.command('push')
   .callback(function(opts) {
-    deploy.sftpDeploy(opts, true); // true: push=true
+    controller.runScript(opts, true); // true: push=true
```
Frijol: Can the function be called deploy? Match the script name

Frijol:
Would much prefer the individual functions be separated out into their own files in lib. This promotes modularity & should make it easier to find functions.

bin/prime.js
```diff
@@ -49,55 +46,11 @@ parser.command('erase')
   })
   .help('Erase pushed code from Tessel filesystem.');
 
-parser.command('root')
```
Frijol: Why is root gone?

bin/prime.js
```diff
@@ -49,55 +46,11 @@ parser.command('erase')
...
-  .option('keyPath', {
-    abbr: 'k',
-    help: 'The path to your SSH Key'
-  })
-  .option('keyPassphrase', {
-    abbr: ''
-  })
-  .help('Access the terminal of your remote Tessel');
-
-parser.command('wifi')
```
Frijol: Why is WiFi gone?

lib/tessel.js
```diff
@@ -0,0 +1,141 @@
...
+  Fetch the human friendly name of this Tessel.
+    param callback: called upon completion in the form (err)
+*/
+Tessel.prototype.getName = function(callback) {
+  // TODO: actually fetch a name
+  // this.write(commands.getIdentification, function(err, readStream) {
+
+  // });
+  // This is for testing purposes only
+  this.name = 'Jon\'s Tessel'
```
Frijol: Hardcoded?

bin/prime.js
johnnyman727: Most of it will be added back in once we have USB comms but it's not that useful otherwise and I wanted to get this PR out quickly.

bin/prime.js
johnnyman727: It was different functionality than the spec documents.

PR #47, opened 2015-03-27
Basic WiFi interactions

johnnyman727:
- Scan for networks
- Fetch connected Tessel's IP Address
- Connect to different networks.


bin/tessel-2.js
```diff
@@ -73,6 +73,47 @@ parser.command('init')
...
+    }
+    else if (opts.ssid && opts.password) {
+      controller.connectToNetwork(opts, function(err) {
+        if (err) throw err;
+        process.exit(1);
+      })
+    }
+  })
+  .help('Configure the wireless connection');
+
```
HarleyKwyn: Why did you make this an if else chain here instead of using options in a handler in the controller? I can see arguments for both but I'd like to know your reasoning for this. It's opposite what I did for tessel init and I just want to be consistent.

lib/tessel/wifi.js
```diff
@@ -0,0 +1,126 @@
...
+  self.connection.exec(commands.connectToNetwork(ssid, password), function(err, streams) {
+    if (err) return callback && callback(err);
+
+    streams.stdin.once('close', function() {
+      tessel.logs.info("Credentials set!");
+      self.connection.end();
+      return callback && callback();
+    });
+  });
+}
```
HarleyKwyn: Little pedantic but make sure to add a newline at the end of your files. 
http://sublime-text-unofficial-documentation.readthedocs.org/en/latest/reference/settings.html#automatic-behavior

HarleyKwyn:
When you have a moment if you could show me how to use this correctly, I couldn't get it working on the tessel wifi.

bin/tessel-2.js
johnnyman727: @HarleyKwyn I believe this command uses the same format as the others but it's the only one that has different options for what the command should do based on flags. I still use the `parser.command('wifi')` but I also need to check the flags to figure out which function to call.

johnnyman727:
Closing in favor of #49

PR #48, opened 2015-03-27
Promises, promises everywhere

### LINKED ISSUE #45: Improve user experience of `tessel list`
Currently, the `tessel list` command waits an arbitrary number of seconds while searching for Tessels on the LAN. This should be changed so that USB Tessels are returned immediately and asynchronously from the LAN Tessels and the LAN Tessels should be returned faster than 2 seconds.

HarleyKwyn:
Let's hope they deliver :package:  :P 

Fixes #45


lib/usb_connection.js
```diff
@@ -41,55 +42,65 @@ USBConnection.prototype._read = function(num) {
...
+          .map(function(dev) {
+            if ((dev.deviceDescriptor.idVendor == TESSEL_VID) && (dev.deviceDescriptor.idProduct == TESSEL_PID)) {
+              return new USBConnection(dev);
+            }
+          })
+          .filter(function(x) {return x});
+      resolve(connections);
+    })
+    .map(function(conn){
+      console.log(typeof conn);
```
johnnyman727: Delete

lib/lan_connection.js
```diff
@@ -38,80 +40,71 @@ LANConnection.prototype.end = function() {
...
-        conn.initError = err;
-        cb(err);
       });
-    }, function(err) {
-      callback(err, connections)
-    });
-  }
+    })
+    // Create new LAN connections for each tessel.
+    .map(function(tessel){
```
johnnyman727: Previously, the "preferred" Tessel was the first Tessel in the returned Tessel list. Admittedly, there is a better way to handle that, but for now, this logic places the Tessel selected by the `config.env` somewhere randomly in the list. It needs to be first.

lib/lan_connection.js
```diff
@@ -38,80 +40,71 @@ LANConnection.prototype.end = function() {
...
+      if (tessel.host === lanConfig.host){
+        return new LANConnection(lanConfig);
+      }
+      // Add credentials before testing?
+      return new LANConnection(tessel);
+    }, {concurrency: 6})
```
johnnyman727: Out of curiosity, what is this concurrency arg? Why is it set to 6?

lib/tessel/mdns-list.js
```diff
@@ -0,0 +1,58 @@
+var mdns = require('mdns-js')
```
johnnyman727: In my opinion, this file shouldn't be in the `tessel` folder because it's not an action that's taken by a single Tessel (unlike erase, deploy, etc.). I think it should be in the `lib` folder.

lib/lan_connection.js
```diff
@@ -38,80 +40,71 @@ LANConnection.prototype.end = function() {
...
+    // Test the connections.
+    // Not sure if we want to throw errors here or somewhere else.
+    .map(function(connection){
+      return connection.open();
+    }, {concurrency: 6});
 }
 
-// TODO: Actually find Tessels on the network using mDNS
 function findmDNSConnections(callback) {
```
johnnyman727: Can we remove (refactor) this function? It's only got one line and it just returns another function.

lib/lan_connection.js
HarleyKwyn: Arbitrary number, but allows six of these functions to run in parallel.

lib/tessel/mdns-list.js
HarleyKwyn: totally, didn't know where to put it and was following old architecture. We should have a conversation about what the file structure is.

lib/lan_connection.js
johnnyman727: Is it required or can it be removed? *fears magic numbers*

HarleyKwyn: 666 (yes, it can be removed, it's only a small performance increase)

lib/lan_connection.js
HarleyKwyn: Will in the future impliment this to be controlled by flags.

PR #61, opened 2015-04-24
Enables connecting to a local VM.

tcr:
This fixes two minor bugs: the inability to detect "localhost" if specified, and a missing timeout parameter.

### DIFF
--- a/lib/lan_connection.js
+++ b/lib/lan_connection.js
@@ -82,9 +82,16 @@ function findConnections(timeout) {
             lanConfig = {
               host: envConfig.host,
               username: 'root',
-              port: 22,
+              port: envConfig.port || 22,
               privateKey: fs.readFileSync(envConfig.keyPath),
               passphrase: envConfig.keyPassphrase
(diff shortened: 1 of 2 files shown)


lib/lan_connection.js
```diff
@@ -96,11 +103,18 @@ function findConnections(timeout) {
     })
     // Create new LAN connections for each tessel.
     .map(function(tessel){
-      if (tessel.host === lanConfig.host){
-        return new LANConnection(lanConfig);
+      try {
+        if (tessel.host === lanConfig.host){
+          return new LANConnection(lanConfig);
+        }
+        // Add credentials before testing?
```
johnnyman727: What is this comment about? Do we need to add an Issue?

johnnyman727:
It would be nice if we could pick up this vm using `t2 list`. Currently, the logic for fetching Tessels to list and getting the anticipated Tessels are separate to the changes you made won't work for `t2 list`.

PR #73, opened 2015-05-02
Update setup instructions to use `tessel setup`

johnnyman727:


johnnyman727:
@Frijol how does this look?

Frijol:
Changed it around.. what do you think of this?

johnnyman727:
looks good.

PR #89, opened 2015-05-04
Adds the ability to set remote devices manually

johnnyman727:


johnnyman727:
@tcr can you test this out and tell me what you think?

To add the default VM, you would run:
```
t2 remote add localhost -k PATH_TO_YOUR_VM/.vagrant/machines/default/virtualbox/private_key -p 2222 -w tessel2
```

Make sure to insert your path to your VM.

johnnyman727:
Closing this since we're taking a different approach.

Folks will run
```
tessel key generate
```

prior to using the VM, if they haven't deploy to hardware before.

Then the public key will automatically be pulled into the VM as the Vagrantfile dictates.

The VM will be able to be picked up using mDNS.

PR #97, opened 2015-05-06
Unbreak nomnom help

Frijol:
Before this change:

```sh
➜  t2-cli git:(master) t2 wifi -h
Usage: t2 <command>
```

After this change:
```
➜  t2-cli git:(master) t2 wifi -h

Usage: /usr/local/bin/iojs t2 wifi [options]

Options:
   --name NAME      the name of the tessel on which the command will be executed.
   -l, --list       List available Wifi networks
   -n, --ssid       Set the SSID of the network to connect to
   -p, --password   Set the password of the network to connect to

Configure the wireless connection
```

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -184,6 +184,4 @@ parser.command('wifi')
   })
   .help('Configure the wireless connection');
 
-parser.usage('Usage: t2 <command>');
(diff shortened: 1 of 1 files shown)


johnnyman727:
r+

Issue #99, opened 2015-05-06
t2 key


johnnyman727:
We'd like a way to create an SSH key and without having to transfer it to a USB-connected device like `tessel provision` does. This is immediately useful for creating a simple and functional VM story but will ultimately be useful for managing the lifecycle of the key. 

It will tie into most of the functionality already created by #64.

```
$ t2 key generate
> Creating public and private keys for Tessel authentication...
> Private key written.
> Public key written.

$ t2 key generate
> Key already exists. Run `key delete` to delete it first before creating a new key.

$ t2 key display-public
> Displaying public key...
> <dd dd dd..... >

$ t2 key delete
> Deleting key...
> Key deleted.
```

- [x] key generate
- [ ] key display-public
- [ ] key remove

tcr:
`delete` instead of `remove` (so that `add` and `remove` have pairity, for specifying named things, versus `generate` and `delete` which refer to one thing)

Also `display` is ambiguous, does that just mean the public key? (I imagine) Then we might make that `t2 key display-public` or something similar.

johnnyman727:
+1 to `delete`

I changed `display` to `display-public` to make it more clear.

johnnyman727:
`generate` added with #100.

kevinmehall:
Why would you want to `delete` (`tessel lock-me-out-of-all-my-tessels`), and do we care to support that vs just tell someone to  `rm -rf ~/.tessel`?

johnnyman727:
I don't really have an issue with having a `delete` command (it should at least prompt you to make sure you know what you're doing) but I also don't really see the use case besides building tests for the CLI.

johnnyman727:
I'm going to go ahead and close this issue for now because I haven't heard a single person request it.

PR #100, opened 2015-05-06
Adds key generation command

johnnyman727:
Lets you run `t2 key generate` without having to provision a physical piece of hardware. Useful for VMs. Working towards #99.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 var parser = require("nomnom")
   , controller = require('../lib/controller')
+  , key = require('../lib/key')
   , init = require('../lib/init')
   , tessel = require('tessel')
   ;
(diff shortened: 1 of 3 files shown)


bin/tessel-2.js
```diff
@@ -187,4 +188,22 @@ parser.command('wifi')
...
+parser.command('key')
+  .option('method', {
+    position: 1,
+    required: true,
+    choices: ['generate'],
+  })
+  .callback(function(opts) {
+    key(opts)
+    .then(function(){
+      console.log('then called!');
```
tcr: I got "then called!" and an error code of 1 which signifies error, though it was successful?

bin/tessel-2.js
```diff
@@ -187,4 +188,21 @@ parser.command('wifi')
...
+    choices: ['generate'],
+  })
+  .callback(function(opts) {
+    key(opts)
+    .then(function(){
+      process.exit(0);
+    })
+    .catch(function (err) {
+      tessel.logs.warn(err);
+      process.exit(0);
```
tcr: Should be exit(1) ?

tcr:
lgtm

Issue #86, opened 2015-05-04
t2 list


Frijol:
Current status:

* prints out USB devices and LAN devices
* takes a timeout flag
* shows authorization status of tessels
* doesn't show which tessel is selected

Spec: [T2 CLI spec](https://forums.tessel.io/t/t2-cli-spec-comments-requested/1743)
**tessel list**

Lists all connected tessels. Shows which tessel is selected and authorization status of each tessel.

Find default per https://github.com/tessel/t2-cli/issues/93
```sh
# Help command example
$ tessel list --help
> Lists all connected tessels. Shows which tessel is selected and authorization status of each tessel.
$ 

# Single Tessel found example
$ tessel list
> Searching for nearby Tessels...
>           Jigglypuff    USB
$ 

# Same Tessel with both USB and LAN connections
$ tessel list
> Searching for nearby Tessels...
>           Jigglypuff    USB
>           Jigglypuff    LAN
$ 

# Many Tessels available
$ tessel list
> Searching for nearby Tessels…
>            Rafael        LAN
>            Bulbasaur     USB       (run `tessel provision` to authorize)
>            Mewcat        LAN       (USB connect and run `tessel provision` to authorize)
>            Jigglypuff    USB
>            Jigglypuff    LAN
>
> Multiple Tessels found. Will default to Bulbasaur.
> Set default Tessel with environment variable (e.g. export TESSEL=Bulbasaur) or use the --name flag.
$ 

# No Tessels available
$ tessel list
> Searching for nearby tessels...
> ERR! No Tessels Found
$
```

Option: tessel list -t to set a timeout

To do:
* [x] Change functionality to display both authorized and unauthorized devices
* [ ] Change printout style to match spec (layout, connection, authorization, hostname)
* [ ] Remove timeout flag
* [ ] Show which tessel is selected
* [ ] Make help flag match spec

nplus11:
I don't think the most important piece is the connection type. The name of the device should come first.

Also it should probably mention that `tessel provision` should be used via USB. First thing user is going to run after reading this is something like `tessel provision Mewcat`

Frijol:
Sure, switching the order seems like a good idea. Maybe even:

```sh
[selected] Rafael       LAN
           Bulbasaur    USB       (run `tessel provision` to authorize)
           Mewcat       LAN       (USB connect and run `tessel provision` to authorize)
           Jigglypuff   USB/LAN
```

nplus11:
[...]

Issue #70, opened 2015-05-02
How does t2-cli work with hardcoded IPs of Tessel devices?


tcr:
This should be a general solution. Ideally: I can add several hardcoded IPs and ports for remote tessels, or a local Tessel VM operating on port 222, and have them show up on `tessel list` and `tessel select`.

tcr:
Relevant: #67

johnnyman727:
Can we just use a config file for this? 

We can have the CLI list the devices you've specified in the config file so that you can use the --name or --ip flags to interact with them.

johnnyman727:
I think a command to add specific ip addresses should be easy to add. We'll just keep a config file in `.tessel` with a list of these specified Tessels and expose them to the rest of the API.

```
tessel remote:add ip name
tessel remote:add 192.168.1.1 vm

tessel remote:remove name OR ip
```

What do you think @tcr?

johnnyman727:
Alternatives are:
- just add optional flags to this command structure for every possible config setting
- have users create a JSON configuration file that specifies the config settings and register that file.

Frijol:
This is a cool idea, but wondering how strong of an analogy we want to make to a git remote

johnnyman727:
Working on this [here](https://github.com/tessel/t2-cli/pull/89).

johnnyman727:
Here is the current design of the API
```
tessel remote <method> <host>

<method> is one of add, remove, or get. add will create a new remote target, remove will de-register that target so the CLI no longer attempts to list it or deploy to it, and get simply prints out the details of the host.

<host> is the hostname or ip address of the remote device to add. You can use 'localhost' for things like virtual machines.

Options:
-k / --keypath : path to the SSH key to use (defaults to HOME/.ssh/id_rsa)
-p / --port : port to connect to (defaults to 22)
-w / --password: password to connect with (defaults to none)
--passphrase: the passphrase necessary to use the SSH key (defaults to none)
```

To modify a remote, just add it again with the same hostname and different options.

kevinmehall:
Have you tested whether mDNS works if you use a different network mode in VirtualBox? It's clear why it doesn't work in the default NAT mode, but I'd expect it to work in bridged mode.

Otherwise, would it work just to allow an IP address (or possibly also FQDN if a local network has a proper DNS setup) anywhere a name is allowed?

johnnyman727:
So after a little exploration, we've found that the CLI can detect VMs using mDNS under certain circumstances:

[...]

PR #102, opened 2015-05-06
Removes -all flag for tessel list

johnnyman727:


johnnyman727:
@Frijol r+?

Frijol:
lgtm, should this be linked to an issue?

johnnyman727:
Good point! This is part of the solution for #86 and also necessary to get VMs functional with the CLI (#70).

PR #106, opened 2015-05-07
Explicitly require promise in key.js

johnnyman727:


johnnyman727:
@Frijol quick r?

Frijol:
lgtm

Issue #64, opened 2015-04-30
t2 provision


Frijol:
command `t2 provision` [T2 CLI spec](https://forums.tessel.io/t/t2-cli-spec-comments-requested/1743): A silent command (not exposed in CLI): [ONLY WORKS OVER USB] creates a .tessel folder with ssh keys in your home directory and uses those ssh keys to authorize you to push code to the USB-connected Tessel

Current status:

* [x] `t2 setup` is a callable command
* [x] creates a .tessel folder in computer's home directory containing an ssh keypair
* [x] copies public key from ~/.tessel into tessel's /etc/dropbear/authorized_keys, thus authorizing the given computer to program that tessel over SSH
* [x] Change name to 'provision' https://github.com/tessel/t2-cli/pull/101

Remaining issues:
* [x] Needs to work on Windows as well (see #147)
* [x] Needs error message if you try to call it over SSH: "Connect via USB to use tessel provision" https://github.com/tessel/t2-cli/pull/111

Frijol:
There is some contention about this in the spec. Summarizing the conversation:

1. Silent vs. callable
I suggest that this should remain a callable command. I'm worried that if the command is silent, it's not obvious that a Tessel needs to be programmed over USB the first time it is used.
@johnnyman727 agrees that it should be explicit.
@nplus11 thinks it should be a silent command because it's an extra step for the user
What is the best interaction for authorizing a computer to program a given Tessel for the first time?

2. Name
The original name for the command was `setup`. @johnnyman727 suggests `provision`. I suggest `auth`. What is the most intuitive name?

johnnyman727:
@tcr @kevinmehall @jiahuang @ekolker @HarleyKwyn @nplus11 do any of you have opinions on this?

kevinmehall:
Explicitly called. What does deploying code have to do with manipulating wifi authentication credentials? You don't necessarily want to set up credentials just because you push once over USB, and you don't necessarily push over USB before wanting to push over WiFi.

johnnyman727:
+1 @kevinmehall any naming preferences?

rwaldron:
What is wrong with making it both callable _and_ automatic?

- if I call it before a push, good: it's done
- if try to push and the requisite .tessel + keys, etc are not created, then call it before the push 
- it's idempotent, so subsequent calls have same result. 
- I can purge and start anew.


Naming...

- "setup" is too ambiguous
- "auth" could work
- "provision" is definitive

[...]

PR #111, opened 2015-05-07
errors if you try to provision over LAN

Frijol:
per https://github.com/tessel/t2-cli/issues/64
r? @johnnyman727


johnnyman727:
Looks good!

PR #114, opened 2015-05-07
Stops renaming if newName and oldName are the same

nplus11:
This was simply to reduce the number of times mdns has to be reset. It's not like this is going to happen all the time, but is something worth avoiding.

### DIFF
--- a/lib/tessel/name.js
+++ b/lib/tessel/name.js
@@ -104,7 +104,7 @@ Tessel.prototype.rename = function(opts, callback) {
       if (err) return callback && callback(err);
       // Don't set the new name if it's the same as the old name
       if (oldName == opts.newName) {
-        tessel.logs.info("Name of device is already", oldName);
+        tessel.logs.warn("Name of device is already", oldName);
         callback && callback();
       }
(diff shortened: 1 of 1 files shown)


lib/tessel/name.js
```diff
@@ -102,6 +102,11 @@ Tessel.prototype.rename = function(opts, callback) {
   else {
     self.getName(function(err, oldName) {
       if (err) return callback && callback(err);
+      // Don't set the new name if it's the same as the old name
+      if (oldName == opts.newName) {
+        tessel.logs.info("Name of device is already", oldName);
```
johnnyman727: I recommend that you just return this string as the first argument of the callback so it gets printed like an error:
```
callback && callback("Name of device is already " + oldName);
```

nplus11: Yes I considered that. I don't think it's an error though. Happy to change it if you feel strongly.

johnnyman727: How about you print it as `tessel.logs.warn`? I think there should be some effort to bring it to the user's attention that they're doing something wrong.

Issue #79, opened 2015-05-04
t2 run needs flags


Frijol:
Current status: t2 run <index.js> will push index.js to RAM and run it immediately.

Spec: [T2 CLI spec](https://forums.tessel.io/t/t2-cli-spec-comments-requested/1743)
**tessel run <index.js> [--lan | --usb]**

Run the file in RAM immediately

```sh
$ tessel run --help
> Usage: tessel run <index.js> [--lan | --usb]
>  index.js:		the Javascript file to run in RAM
>  --lan | --usb:	connect over the specified connection type
$ 

$ tessel run index.js
> Writing index.js to RAM on Mewcat (16kB)...
> Deployed.
> Running index.js...
> Hello world!
$ 
```

To do:

* [x] Create --help flag
* [x] Create --lan and --usb flags and make code push over the specified connection

PR #115, opened 2015-05-08
Add lan and usb flags for run and push 

Frijol:
r? @johnnyman727 or @HarleyKwyn

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,11 @@ Specify which Tessel to use with the `--name <name>` option appended to any comm
 
 ### Code Deploy
 * `t2 run <file>` deploy the file and its dependencies
+  * `[--lan]` deploy over LAN connection
+  * `[--usb]` deploy over USB connection
 * `t2 push <file>` copy the file and its dependencies into Tessel's Flash & run immediately
+  * `[--lan]` deploy over LAN connection
+  * `[--usb]` deploy over USB connection
 * `t2 erase` erase any code pushed using the `t2 push` command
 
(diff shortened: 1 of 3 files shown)


HarleyKwyn:
This looks kosher to me, but I don't have a tessel 2 to test it with.

johnnyman727:
When you try to use an option that isn't available (like trying to deploy over LAN when the Tessel is not connected to the internet) it just prints out that no Tessels were found:
```
➜  t2-cli git:(kb-run) t2 run test/test-deploy-script.js --lan
INFO Connecting to Tessel...
ERR! No Tessels Found
```

Do we want to have a more descriptive error message for this case? The [issue](https://github.com/tessel/t2-cli/issues/79) doesn't actually document what should happen.

Frijol:
That's what I thought it should do– reads to me as "no tessels found on LAN". What would you prefer?

HarleyKwyn:
FWIW I think that this is a correct response. The suggestion Kelsey just
specified would also be good.

johnnyman727:
Fair enough - looks good!

HarleyKwyn:
Could quickly be added with
```
        var connectionType = '';
        if(opts.lan) connectionType = 'LAN';
        if(opts.usb) connectionType = 'USB';
        reject('No ' + connectionType + ' Tessels Found');
```
on line 98 I beleive.

Frijol:
I personally like it as is

HarleyKwyn:
:+1:

PR #121, opened 2015-05-12
Make a link for users to join Slack

johnnyman727:


tcr:
Should we mention what Slack is?

"Join us on Slack, a team communication platform."

Or something less corporate.

Frijol:
I kinda like it as is– "join the conversation" feels appropriate

johnnyman727:
I added a little note about what it is. Does this work for you @tcr and @Frijol?

Frijol:
:/ I like this less because it conveys no new information. If we need to specify, why not "the project's chat" or similar

johnnyman727:
@Frijol  How about `Join us on Slack, the project's chat client!`

Frijol:
sure

Issue #118, opened 2015-05-12
Task runner and common dev tasks


rwaldron:
- Pick one, it doesn't matter: Grunt or Gulp? I vote Grunt, but will use either
- Tasks: 
  - JSHint
  - JSCS
  - Test Runner

johnnyman727:
+1 to Grunt

rwaldron:
Cool. I can get started on this

johnnyman727:
That would be fantastic. I don't have much experience building scalable test benches so it'd be great to learn more about Grunt.

rwaldron:
Working through this and I'm a little overwhelmed. These changes necessary to make these new quality control tools happy are massive.

rwaldron:
There are a lot of bugs that will be fixed in this process, so care should be taken to test the patch thoroughly.

PR #124, opened 2015-05-13
Task runner and common dev tasks. Closes gh-118

rwaldron:
This is the first round. There will be a follow up issue to write some base line unit tests for the existing features. 

Additionally, automated ci runs with travis needs to be set up for this repo.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -50,7 +50,6 @@ module.exports = function(grunt) {
         // 'resources/**/*.js',
       ],
       options: {
-        mode: 'VERIFY_ONLY',
         braceStyle: 'collapse',
         breakChainedMethods: false,
         e4x: false,
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -89,9 +89,10 @@ module.exports = function(grunt) {
 
   // These plugins provide necessary tasks.
   grunt.loadNpmTasks('grunt-contrib-jshint');
+  grunt.loadNpmTasks('grunt-contrib-nodeunit');
   grunt.loadNpmTasks('grunt-jscs');
   grunt.loadNpmTasks('grunt-jsbeautifier');
-  grunt.loadNpmTasks('grunt-nodeunit');
+
 
   // 'npm test' runs these tasks
   grunt.registerTask('test', ['jshint']);
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -50,23 +50,25 @@ module.exports = function(grunt) {
         // 'resources/**/*.js',
       ],
       options: {
-        braceStyle: 'collapse',
-        breakChainedMethods: false,
-        e4x: false,
-        evalCode: false,
-        indentChar: ' ',
-        indentLevel: 0,
-        indentSize: 2,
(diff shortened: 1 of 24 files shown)


Gruntfile.js
```diff
@@ -0,0 +1,103 @@
...
+      options: {
+        jshintrc: '.jshintrc'
+      },
+      all: [
+        'bin/*',
+        'lib/**/*.js',
+        'test/**/*.js',
+        'Gruntfile.js',
+
+        // This is commented out because there are
```
johnnyman727: @rwaldron would it help if I went through these errors?

test/unit/readme.md
```diff
@@ -0,0 +1,5 @@
+TODO: 
+
+- write auto-mocks
+- install sinon
+- write units for existing functionality
```
johnnyman727: I can also help with this!

rwaldron: These "todo" items should be filed as new issues when this lands.

johnnyman727:
@rwaldron I'm new to Grunt so let me know if I've configured something wrong, but when I run `grunt` in the root of the project, I get the following error: ```Warning: Task "grunt-jsbeautifier" not found. Use --force to continue.```. Is this expected?

lib/usb/usb_process.js
```diff
@@ -13,47 +11,46 @@ function USBProcess(id, daemon) {
...
 
-  // This puts the stream into 'flowing mode' 
+  // This puts the stream into 'flowing mode'
   // so that the stream emits ('end') events without 'data' listeners
   this.stdout.resume();
   this.stderr.resume();
 
   this.stdout.signal = function(signalToSend) {
-    switch(signalToSend) {
+    // RWQ: Does this need to a `switch`?
```
johnnyman727: It eventually will; I only have one case at the moment but I suspect there will be more signals that we'd like to send. In fact, #126 adds another case.

johnnyman727:
Quick update: I started working through this PR but was having issues getting code deploys to work properly. Part of the issue is almost definitely that I'm using a new version of the USBDaemon firmware which handles USB comms but it may also be an issue with the bugs you fixed. Will dive in deeper tomorrow or Saturday.

rwaldron:
Sorry for the delayed follow up, I had family things going on during the end of last week

rwaldron:
There's been too much churn since this patch was created. I don't have the time resources right now to properly rebase and resolve all the conflicts. I'm going to take another approach, which is to do only one lint rule pass at a time.

johnnyman727:
Yesterday I had the chance to dive into this PR on a plane ride and I found only two issues with the PR:

- [Line 84](https://github.com/tessel/t2-cli/pull/124/files#diff-f68567477d803b49930337bf7fe1556bL86) was incorrectly replaced with `return Promise.settle([usb.findConnections(), lan.findConnections(timeout)])` when it needs to be `return Promise.settle(connectionTypes)`.

- The `t2 init` command no longer works but it appears to be an issue with the Promzard library. I haven't investigated very far.

I'm happy to move my attention to the new PR but I'd thought I'd place a record of my review here.

PR #125, opened 2015-05-13
Uses Tessel v1 CLI code only in packaging / deploy steps.

tcr:
"tessel classic cli" aborts when there's no USB controller. Current t2-cli behavior is similar, but this prevents simple scripts like `t2 key generate` from working on a VM, which is annoying.

This PR isolates the "tessel classic" code to where it's actually used. Ideally we'd split out the logging, packaging, and deploy code into some common core, or at least one that doesn't try to forcibly load a USB controller.


johnnyman727:
@tcr looks good if you want to merge

PR #126, opened 2015-05-14
Fixes #117 [and rewrites a large portion of USB-based comms]

### LINKED ISSUE #117: t2 wifi -l can hang
The root cause of this hang is [a firmware bug](https://github.com/tessel/t2-firmware/issues/30) and will be resolved with that issue. 

It seems to be caused when there are a ton of networks in the vicinity.

johnnyman727:
Ready for merge. Note that it may not work properly if you don't have [the firmware patch](https://github.com/tessel/t2-firmware/pull/37) loaded up.

### DIFF
--- a/lib/usb/usb_process.js
+++ b/lib/usb/usb_process.js
@@ -15,22 +15,30 @@ function USBProcess(id, daemon) {
   var self = this;
   // The numerical identifer for this process
   this.id = id;
-  // Status of whether this process is killed or active
+
+  /*
+  active && !closed = Process is running as usual
+  !active && !closed = Process has been sent a signal but not completely closed (resources freed)
+  !active && closed = Process has stopped running and has already been destroyed
+  */
+  // Status of whether this process is running
   this.active = true;
+  // Status of whether this process was cleaned up
+  this.closed = false;
+
   // 4 remote process streams
-  this.control = new RemoteWritableStream(id, daemon, protocol.controlWrite, protocol.controlClose);
-  this.stdin = new RemoteWritableStream(id, daemon, protocol.stdinWrite, protocol.stdinClose);
-  this.stdout = new RemoteReadableStream(id, daemon, protocol.stdoutAck);
-  this.stderr = new RemoteReadableStream(id, daemon, protocol.stderrAck);
+  this.control = new RemoteWritableStream(self, daemon, protocol.controlWrite, protocol.controlClose);
+  this.stdin = new RemoteWritableStream(self, daemon, protocol.stdinWrite, protocol.stdinClose);
+  this.stdout = new RemoteReadableStream(self, daemon, protocol.stdoutAck);
+  this.stderr = new RemoteReadableStream(self, daemon, protocol.stderrAck);
 
   // Once the process is killed
   self.once('death', function(data) {
+    // This process is no longer running
+    self.active = false;
     // Close the input streams because the remote proc can't handle data
     self.control.end();
     self.stdin.end();
-
-    // Mark this process as inactive
-    self.active = false;
 
     // Close the process
     self.closeProcessIfReady();
@@ -46,6 +54,9 @@ function USBProcess(id, daemon) {
   self.close = function() {
     // Write the close command
     daemon.write(protocol.closeProcess(self.id));
+
+    // Mark this process as closed
+    self.closed = true;
   };
 
   // Send a signal to the remote process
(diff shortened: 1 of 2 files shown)


lib/usb/usb_process.js
```diff
@@ -1,236 +1,271 @@
...
-    }});
-
-    if (self.control.credit) {
-      self._drainBackPressure(self.control, protocol.controlWrite);
+  // Checks if this process is ready to be closed
+  // and closes it if it can
+  self.closeProcessIfReady = function() {
+    // If this process hasn't been closed before
+    // and all of the streams are closed
+    if (self.active && self.stdout.closed
```
kevinmehall: Should this be `!self.active`? (i.e. the process has exited).

Also, I'm not sure about including stdin here, since closing stdin is controlled by the CLI side. It's fine if the user of this library is careful to close stdin when done writing. Programs that accept input on stdin like `tar` will themselves not exit until their stdin is closed, so those are fine. But when you run a program that doesn't accept input, you may not touch stdin at all and forget to close it, leaking the remote process.

lib/usb/usb_process.js
```diff
@@ -1,236 +1,271 @@
...
-      if (entry.buffer.length == 0) {
-        // Remove it from back pressure
-        stream.backPressure.shift();
-        // Call the callback if applicable
-        entry.callback && entry.callback();
-      }
+  // This function is primarily intended for compatibility with the ssh2 stream API
+  self.stdout.signal = function(signalToSend) {
+    switch(signalToSend) {
+      case 'KILL':
```
kevinmehall: I'd expect to see 'INT' here, as that's the signal you should be sending first. (I'm thinking of `tessel run` ctrl-c in particular)

lib/usb/usb_process.js
```diff
@@ -1,236 +1,271 @@
...
-    self.exitCode = data.arg;
-    // Close the input streams
+    // Close the input streams because the remote proc can't handle data
     self.control.end();
     self.stdin.end();
-    // Clean up remote resources
-    self.close();
+
+    // Mark this process as inactive
+    self.active = false;
```
kevinmehall: You should also `closeProcessIfReady` here, as it's possible for stdout and stderr to EOF before the process exits.

lib/usb/usb_process.js
```diff
@@ -1,236 +1,271 @@
...
-  // Intercepting write to the stdin stream
-  this.stdin._write = function(chunk, enc, callback) {
-    // Add this chunk to the backpressure array
-    self.stdin.backPressure.push({buffer: chunk, callback: callback});
-    if (self.stdin.credit) {
-      self._drainBackPressure(self.stdin, protocol.stdinWrite);
-    }
-  };
+  // When any of the streams complete, check if the process is ready to close
+  self.stdin.once('finish', self.closeProcessIfReady);
```
kevinmehall: If you remove stdin from the condition as described above, you can also remove this.

lib/usb/usb_process.js
```diff
@@ -1,236 +1,271 @@
...
+  this.control = new RemoteWritableStream(id, daemon, protocol.controlWrite, protocol.controlClose);
+  this.stdin = new RemoteWritableStream(id, daemon, protocol.stdinWrite, protocol.stdinClose);
+  this.stdout = new RemoteReadableStream(id, daemon, protocol.stdoutAck);
+  this.stderr = new RemoteReadableStream(id, daemon, protocol.stderrAck);
 
   // Once the process is killed
   self.once('death', function(data) {
-    self.active = false;
-    // Set the exit code
-    self.exitCode = data.arg;
```
kevinmehall: Are you no longer storing the process exit code?

lib/usb/usb_process.js
johnnyman727: Yeah, I can see how that would clean up the logic. I'll make that change.

lib/usb/usb_process.js
johnnyman727: :+1:

lib/usb/usb_process.js
johnnyman727: :+1:

lib/usb/usb_process.js
johnnyman727: :+1:

lib/usb/usb_process.js
johnnyman727: No - it wasn't being used so I took it out for now.

kevinmehall: You should check the exit code and propagate a nonzero exit code as an error.

johnnyman727: Good point - added.

johnnyman727:
@kevinmehall I raised an error on a non-zero return arg on process exit and fixed two minor race conditions.

johnnyman727:
@kevinmehall I believe all of your comments have been addressed and the code has been linted.

kevinmehall:
Looks good to me, but I haven't tested it.

johnnyman727:
:+1: I ran it through all of the available CLI commands multiple times with no issue.

PR #129, opened 2015-05-18
Removing ip address functionality which isn't used by anything

johnnyman727:
This has been bothering me for a while. The `ip` flag to `t2 wifi` (originally to print out the IP Address) is no longer available but the underlying functionality was still sitting in the library. This PR removes it.

It also removes duplicated functionality for the `wifi list` and `print IP` options.


Frijol:
lgtm

PR #130, opened 2015-05-18
[WIP] getting build versions over cli

jiahuang:
goes with https://github.com/tessel/t2-firmware/pull/38

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -228,4 +228,35 @@ parser.command('rename')
   })
   .help("Change the name of a Tessel to something new.");
 
+parser.command('version')
+  .option('mcu', {
+    abbr: 'm',
+    flag: true,
+    help: 'The version number from the MCU (Mediatek)'
+  })
+  .option('cp', {
+    abbr: 'c',
+    flag: true,
+    help: 'The version number from the CoProcessor (Atmel)'
+  })
+  .option('build', {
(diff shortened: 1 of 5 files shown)


bin/tessel-2.js
```diff
@@ -228,4 +228,35 @@ parser.command('rename')
   })
   .help("Change the name of a Tessel to something new.");
 
+parser.command('version')
+  .option('mcu', {
+    abbr: 'm',
+    flag: true,
+    help: 'The version number from the MCU (Mediatek)'
```
kevinmehall: MT7620n is not a microcontroller

kevinmehall:
The version number of the SoC OS should effectively be a version number for the openwrt-tessel repo, not the t2-firmware repo. The tessel-tools package in that OS image covers only usbexecd, spid, and tessel.js, but that version number does not cover all the other software on the system.

johnnyman727:
@jiahuang are you still actively pursuing this feature or should I close the PR?

jiahuang:
We do need this PR in. Path forward:
1. Add in a file during the openwrt build process. This will specify the openwrt build + the git hash of the t2-firmware repo for spid, usbexecd, and tessel js
2. the command for getting the mediatek build needs to read that file 

The code for getting the firmware from the samd doesn't need modification.

johnnyman727:
Fixed by #231.

PR #134, opened 2015-05-19
Remove unused vars

rwaldron:


jiahuang:
:+1:

lib/tessel/wifi.js
```diff
@@ -98,7 +97,7 @@ Tessel.prototype.connectToNetwork = function(opts) {
   });
 };
 
-Tessel.prototype.resetMDNS = function(callback) {
```
johnnyman727: This callback is still used below and needs to be defined as a parameter.

johnnyman727:
@rwaldron once that one callback is added back in, it's good to merge. Tested on the Hardware™.

PR #136, opened 2015-05-19
Changing console.logs to match run and push spec

Frijol:


lib/tessel/deploy.js
```diff
@@ -84,28 +84,38 @@ function untarStdin(t, filepath, opts, callback) {
...
-      if (err) throw err;
+    // TODO: This section unnecessarily synchronous
+    t.getName(function (err, tesselname) {
+      // Gather details about the file structure of the script being sent
+      var ret = tesselClassic.analyzeScript(process.cwd() + "/" + opts.entryPoint, {verbose: opts.verbose});
+      // Tar up the code to improve transfer rates
+      tesselClassic.tarCode(ret.pushdir, {node: true}, function(err, bundle) {
+        // RAM or Flash for log
+        var memtype;
+        if (opts._[0] == 'push') {
```
johnnyman727: This should be a '===' instead of '=='. (Grunt throws a warning about this)

lib/tessel/deploy.js
```diff
@@ -84,28 +84,38 @@ function untarStdin(t, filepath, opts, callback) {
...
+          memtype = 'Flash';
+        } else {
+          memtype = 'RAM';
+        }
+        // Log write
+        logs.info('Writing ' + opts.entryPoint + ' to ' + memtype + ' on ' + tesselname + ' (' + bundle.length/1000 + ' kB)....');
```
johnnyman727: It would be a little cleaner to separate substrings with commas instead of manually adding all of the spaces:
```.js
logs.info('Writing', opts.entryPoint, 'to' ...);
```

lib/tessel/deploy.js
```diff
@@ -84,28 +84,38 @@ function untarStdin(t, filepath, opts, callback) {
...
-      // Write the code bundle to the hardware
-      remoteProcess.stdin.end(bundle);
     });
   });
 }
 
-function runScript(t, filepath, ret) {
-  logs.info('Running script...');
+function runScript(t, filepath, ret, entryPoint) {
+  logs.info('Running ' + entryPoint + '...');
```
johnnyman727: Same comment here about using commas.

lib/tessel/deploy.js
```diff
@@ -156,14 +166,14 @@ function writeToFile(t, callback) {
   });
 }
 
-function startPushedScript(t) {
+function startPushedScript(t, entryPoint) {
   // Once it has been written, run the script with Node
   t.connection.exec(commands.startPushedScript(), function(err, remoteProcess) {
     remoteProcess.once('close', function() {
-      logs.info('Running script...');
+      logs.info('Running ' + entryPoint + '...');
```
johnnyman727: Same comment here about using commas.

lib/tessel/deploy.js
rwaldron: Even better still: 

```js
logs.info('Writing %s to %s on %s (%d kB)', opts.entryPoint, memtype, tesselname, bundle.length/1000);
```

This has the benefit of completely separating the "ui" from the "logic" (in a manner of speaking).

rwaldron:
Just a heads up: this is going to need rebasing before it can land

Frijol:
Ready to merge @rwaldron or @johnnyman727 ?

johnnyman727:
looks good to me

PR #137, opened 2015-05-19
Fixes Grunt warnings about undefined variables.

johnnyman727:
#134 should be landed first and then I'll rebase.


rwaldron:
:clap:

PR #144, opened 2015-05-27
Connects to LAN via IP, not hostname.local

tcr:
OSes lacking mDNS will also lack the ability to look up Tessels by hostname.local, which is how the SSH connection currently works. Accidentally, we were relying on the host mDNS implementation rather than the JS one.

### DIFF
--- a/lib/lan_connection.js
+++ b/lib/lan_connection.js
@@ -12,7 +12,8 @@ mdns.excludeInterface('0.0.0.0');
 function LANConnection(opts) {
   this.auth = {};
   this.ip = Array.isArray(opts.addresses) ? opts.addresses[0] : opts.host;
-  this.auth.host = opts.host || '';
+  this.host = opts.host || '';
+  this.auth.host = this.ip;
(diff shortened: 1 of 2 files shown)


lib/lan_connection.js
```diff
@@ -12,7 +12,8 @@ mdns.excludeInterface('0.0.0.0');
 function LANConnection(opts) {
   this.auth = {};
   this.ip = Array.isArray(opts.addresses) ? opts.addresses[0] : opts.host;
-  this.auth.host = opts.host || '';
+  this.host = opts.host || '';
+  this.auth.host = this.ip;
```
jiahuang: whats the auth obj used for now that controller.js doesn't use connection.auth anymore?

tcr: We should comment this more thoroughly, but it's the literal SSH connection object.

https://github.com/tessel/t2-cli/blob/master/lib/lan_connection.js#L75

jiahuang:
:+1:

PR #145, opened 2015-05-28
Changes stopRunningScript to do the correct thing.

tcr:
`stopRunningScript` will not actually kill the node script that is running, since it's not launched through `/etc/init.d/tessel-app`. So killing it will let the node script continue to run, eating memory, hogging up ports, causing havoc.

This is a weak solution but will do the correct thing always.

### DIFF
--- a/lib/tessel/commands.js
+++ b/lib/tessel/commands.js
@@ -5,7 +5,7 @@ module.exports.scanWiFi = function() {
   return ['ubus', 'call', 'iwinfo', 'scan', '{"device":"wlan0"}'];
 };
 module.exports.stopRunningScript = function() {
-  return ['/etc/init.d/tessel-app', 'stop'];
+  return ['killall', 'node'];
 };
 module.exports.deleteFolder = function(filepath) {
   return ['rm', '-rf', filepath];


lib/tessel/commands.js
```diff
@@ -5,7 +5,7 @@ module.exports.scanWiFi = function() {
   return ['ubus', 'call', 'iwinfo', 'scan', '{"device":"wlan0"}'];
 };
 module.exports.stopRunningScript = function() {
-  return ['/etc/init.d/tessel-app', 'stop'];
+  return ['killall', 'node'];
```
jiahuang: Seems a little extreme, especially if we ever add something on the system that uses node to run. Is there a way to have push/run return the pid and kill off of that?

tcr: Yeah, but I think that can be deferred; right now it's incorrect behavior and we have no other node scripts running afaik.

jiahuang: Before merging, add this as an issue so that we know it'll be a problem for later.

tcr: Done and done.

johnnyman727:
Did anybody actually test this? I can no longer deploy code to T2 hardware with this patch merged in. I get the following error:
```
➜  t2-cli git:(df128ff) t2 run test/test-deploy-script.js
INFO Connecting to Tessel...
INFO Connected over USB.
events.js:85
            ^
Error: Error upon remote process exit: 1
    at USBProcess.<anonymous> (/Users/Jon/Work/technical/t2-cli/lib/usb/usb_process.js:47:26)
    at USBProcess.g (events.js:199:16)
    at USBProcess.emit (events.js:107:17)
    at Parser.<anonymous> (/Users/Jon/Work/technical/t2-cli/lib/usb/usb_daemon.js:117:14)
    at Parser.emit (events.js:107:17)
    at Parser._write (/Users/Jon/Work/technical/usb-daemon-parser/index.js:152:16)
    at doWrite (_stream_writable.js:301:12)
    at writeOrBuffer (_stream_writable.js:288:5)
    at Parser.Writable.write (_stream_writable.js:217:11)
    at USBConnection.ondata (_stream_readable.js:540:20)
    at USBConnection.emit (events.js:107:17)
    at readableAddChunk (_stream_readable.js:163:16)
    at USBConnection.Readable.push (_stream_readable.js:126:10)
    at InEndpoint.<anonymous> (/Users/Jon/Work/technical/t2-cli/lib/usb_connection.js:74:12)
    at Transfer.callback (/Users/Jon/Work/technical/t2-cli/node_modules/usb/usb.js:274:6)
```

@tcr can you roll this back until we figure out what the issue is?

tcr:
@johnnyman727 If I roll back the patch, it becomes impossible for me to deploy code again due to port conflicts. Is it worth rolling back or just further debugging this?

tcr:
What is the error being thrown by USBProcess at `/Users/Jon/Work/technical/t2-cli/lib/usb/usb_process.js:47`?

johnnyman727:
Yeah, I can spend some more time debugging this weekend.

tcr:
Nevermind, it's probably a waste of time until it's actually an error. This should be the proper fix and undoes #45: https://github.com/tessel/t2-cli/pull/148

PR #148, opened 2015-05-30
Properly stops scripts on death.

### LINKED ISSUE #146: start/stopRunningScript commands should have an acutal lifetime
Temporarily in #145 this is changed to `killall node` to emulate the correct behavior, since running a script from RAM does not involve `/etc/init.d/tessel-app`.

`startRunningScript` and `stopRunningScript` should use a daemon, or at least link the lifetime of the PID to the lifetime of the running script.

tcr:
Restores #145, closes #146.

### DIFF
--- a/lib/tessel/commands.js
+++ b/lib/tessel/commands.js
@@ -5,7 +5,7 @@ module.exports.scanWiFi = function() {
   return ['ubus', 'call', 'iwinfo', 'scan', '{"device":"wlan0"}'];
 };
 module.exports.stopRunningScript = function() {
-  return ['killall', 'node'];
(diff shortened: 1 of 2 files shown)


johnnyman727:
@tcr This works for me and I'll merge it but I don't completely understand how it fixed the problem. Can you elaborate?

tcr:
@johnnyman727 There's two SIGINT handlers attached in two different locations. Normally both would be called. One of them calls `process.exit(1)`, however, so killing the process before the other can run.

This construction:
```
process.on('exit', function () {
  process.exit(1);
})
```

Changes the eventual status code of the exit, without actually causing an exit to occur. Read as: "*When* this program exists, terminate with a status code of 1."

PR #151, opened 2015-06-07
Improves tessel list command

johnnyman727:
This was really not high priority but it was bugging me. This change makes the `tessel list` functionality asynchronous so Tessels are reported as soon as they are found. It also cleans up code and makes the heuristic of which Tessel gets selected out of multiple more clear.


johnnyman727:
I'd love some feedback on the code written here. Specifically, the way I implemented the Scanner classes inside of `usb_connection.js` and lan_connection.js` are redundant and use a global variable. What would be a better way to do this?

Frijol:
Getting some weird behavior

```sh
➜  t2-cli git:(jon-list-rewrite) t2 list
INFO 	Kelsey	USB
INFO 	Kelsey	LAN	(USB connect and run `tessel provision` to authorize)
INFO 	Hidalgo	LAN	(USB connect and run `tessel provision` to authorize)
➜  t2-cli git:(jon-list-rewrite) t2 provision
INFO Connecting to Tessel...
INFO Connected to Kelsey over USB
INFO Authenticating Tessel with public key...
INFO Tessel authenticated with public key.
➜  t2-cli git:(jon-list-rewrite) t2 list
INFO 	Hidalgo	LAN	(USB connect and run `tessel provision` to authorize)
INFO 	Kelsey	LAN
➜  t2-cli git:(jon-list-rewrite) t2 list
INFO 	Hidalgo	LAN	(USB connect and run `tessel provision` to authorize)
➜  t2-cli git:(jon-list-rewrite) t2 list
INFO 	Hidalgo	LAN	(USB connect and run `tessel provision` to authorize)
INFO 	Kelsey	LAN
➜  t2-cli git:(jon-list-rewrite) t2 list
INFO 	Kelsey	USB
INFO 	Hidalgo	LAN	(USB connect and run `tessel provision` to authorize)
INFO 	Kelsey	LAN
```

tcr:
Works with the VMs. Haven't tested more than that.

Can we change the T2 listings to not be INFO-prefixed? It's actual stdout data that I would like to be able to pipe / redirect to a file, etc. `console.log` would replace it.

lib/spinner.js
```diff
@@ -0,0 +1,20 @@
+var charSpin = require('char-spinner');
+
+// Gratuitously taken from npm
+module.exports.spinner = {
+  interval: null,
+  started: false,
```
rwaldron: This isn't used anywhere in `t2-cli`

johnnyman727:
@tcr what kind of logs do you think we should prefix with INFO vs just `console.log`? I'd like to keep it consistent.

johnnyman727:
@Frijol I wasn't able to reproduce but I did see similar behavior before this patch so there is no regression.

tcr:
@johnnyman727 Anything that can be piped to another command. Anything output by INFO should be just debug information and able to be discarded.

If I were  to run `tessel list > logs.txt` I would expect logs.txt to have all the IP addresses from stdout, and not all the debug/INFO logs from stderr. Pretty handy unixism.

johnnyman727:
>  Anything that can be piped to another command

Isn't this the output from any tessel command?  How do you distinguish between which ones are useful and which ones aren't? I'm not trying to be difficult - I actually just don't understand where to draw the line.

tcr:
Just machine vs human readable, mostly. For example:

```
LAN Hidalgo (USB connect and run `tessel provision` to authorize)
LAN Kelsey
```

The first line is for presentation, so that I know what the command is doing. The other lines though are the structural, "data" part of the representation. It can be human readable, but also machine readable. An `awk` command would just split all the whitespace and get a table of (comms, name, comment)

Another idea would be, imagine if you had a `--format` switch (which we should add). Instead of `--format=plain` as usual, we could do `--format=json`. All JSON (or machine-readable) output would be on stdout:

```
{"name":"Hidalgo","type":"LAN"}
{"name":"Kelsey","type":"LAN"}
```

johnnyman727:
@tcr ping

johnnyman727:
We've had three people test it so I'm just merging.

PR #152, opened 2015-06-07
Fixes the reset test after we updated the fake mac addr

johnnyman727:


johnnyman727:
@Frijol can you check this?

Frijol:
lgtm

PR #153, opened 2015-06-07
Unit Tests for wifi.js

johnnyman727:


johnnyman727:
@rwaldron getting closer a bit at a time. How do these tests look?

johnnyman727:
@tcr or @rwaldron, do either of you know how to make Travis CI happy? It's a real shame that the Tessel CLI is being such a hindrance since the only part of it we use is the bundling logic.

test/unit/wifi.js
```diff
@@ -0,0 +1,236 @@
...
+}
+
+function RemoteProcessSimulator() {
+  this.stdin = new stream.Writable();
+  this.stdout = new stream.Readable();
+  this.stderr = new stream.Readable();
+
+  this.stdout._read = function() {};
+  this.stderr._read = function() {};
+}
```
rwaldron: If you think this will be useful for other test files, then move it out to a separate file that can be require()'ed as needed :)

rwaldron:
- The travis errors are unrelated to the actual tests. 
- Take a look at https://github.com/tessel/t2-cli/pull/153#discussion_r32056615

johnnyman727:
All right, I've broken TesselSimulator out into another comment. It would be super helpful if @tcr can check out TravisCI and give me some guidance on how to fix it.

johnnyman727:
Travis is fixed and we're good to go with Grunt!

PR #154, opened 2015-06-13
Proper ssh key file modes and callback usage

johnnyman727:
This tiny PR does three things:
1. Creates both SSH keys with a 0600 file mode
2. Adds a log that the `key generate` function worked (if it did)
3. Properly executes callbacks in the key generation function (which was preventing 2 from working)
4. Changes the `setupLocal` function to call callbacks with `function(err)` where 'Keys already exist` can be an error instead of `function(err, createdBool)`. This method is more consistent with Node.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -194,6 +194,7 @@ parser.command('key')
   .callback(function(opts) {
     key(opts)
       .then(function() {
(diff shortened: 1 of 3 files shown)


johnnyman727:
@Frijol can you give this a quick review?

Frijol:
lgtm

PR #161, opened 2015-06-20
enables wifi when ssid & password is set

jiahuang:
Goes with https://github.com/tessel/openwrt-tessel/pull/14


johnnyman727:
Looks good to me. Feel free to merge.

PR #160, opened 2015-06-17
Fix this-binding issue.

rwaldron:
When either `tessel.close()` is called or `SIGINT` from console, the following error occurs:

```
 t2 run index.js
INFO Connecting to Tessel...
INFO Connected over USB.
INFO Writing index.js to RAM on arnold (14286.336 kB)...
ERR! Unable to deploy code: tar: index.js: time stamp 2015-06-17 20:25:35 is 100537.87948 s in the future

^C/Users/rwaldron/clonez/t2-cli/lib/tessel/tessel.js:22
    this.connection.end(function() {
                   ^
TypeError: Cannot read property 'end' of undefined
    at process.endConnection (/Users/rwaldron/clonez/t2-cli/lib/tessel/tessel.js:22:20)
    at process.g (events.js:199:16)
    at process.emit (events.js:104:17)
    at Signal.wrap.onsignal (node.js:655:46)
```

The issue is caused by `this` being `undefined` inside of the `endConnection`

PR #163, opened 2015-06-22
Adds a unit test for the erase command

johnnyman727:
@rwaldron can you look this over?

### DIFF
--- a/test/common/tessel-simulator.js
+++ b/test/common/tessel-simulator.js
@@ -7,6 +7,12 @@ function TesselSimulator() {
 
   tessel.connection = {
     exec: function(command, callback) {
+
+      if (!Array.isArray(command)) {
+        throw new Error('Invalid command passed to exec.');
+      }
+
+      tessel._rps.control.write(command.join(' '));
       if (typeof callback === 'function') {
         callback(null, tessel._rps);
       }
--- a/test/unit/erase.js
+++ b/test/unit/erase.js
@@ -1,0 +1,85 @@
+var sinon = require('sinon');
+var Tessel = require('../../lib/tessel/tessel');
+var commands = require('../../lib/tessel/commands');
+var logs = require('../../lib/logs');
+var TesselSimulator = require('../common/tessel-simulator');
+
+exports['Tessel.prototype.erase'] = {
+  setUp: function(done) {
+
+    this.erase = sinon.spy(Tessel.prototype, 'eraseScript');
+    this.logsWarn = sinon.stub(logs, 'warn', function() {});
+    this.logsInfo = sinon.stub(logs, 'info', function() {});
+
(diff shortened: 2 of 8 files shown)


test/common/tessel-simulator.js
```diff
@@ -7,6 +7,7 @@ function TesselSimulator() {
 
   tessel.connection = {
     exec: function(command, callback) {
+      tessel._rps.control.write(command.join(' '));
```
rwaldron: This will blow up if command is not an array. I suggest either: 

- Check if it's an array and if not, make it an empty array, or...
- Throw if it's likely to put the program in an unknown state

test/unit/tessel.js
```diff
@@ -35,17 +35,17 @@ exports['Tessel (endConnection)'] = {
...
+  //   tessel.close();
 
-    test.equal(process._events.SIGINT.length, length);
-    test.equal(this.end.callCount, 0);
-    test.done();
-  },
+  //   test.equal(process._events.SIGINT.length, length);
+  //   test.equal(this.end.callCount, 0);
+  //   test.done();
+  // },
```
rwaldron: Was this intentional? What's the issue? Keep in mind: checking the events is not ideal, but was the quickest way to confirm that an explicit `tessel.close();` had removed the event. Now that I'm revisiting it, I think this should be changed to:

```js
  closeConnection: function(test) {
    test.expect(2);

    var processremoveListener = sinon.stub(process, 'removeListener');
    var tessel = new Tessel(this.mockConnection);

    tessel.close();

    test.equal(processremoveListener.callCount, 1);
    test.equal(this.end.callCount, 0);

    processremoveListener.restore();
    test.done();
  },
```


I just tried that locally and it passes and doesn't make me want to vomit ;)

test/unit/erase.js
```diff
@@ -0,0 +1,82 @@
...
+        test.done();
+      })
+      .catch(function() {
+        test.fail('Error thrown on proper flash erase');
+      });
+
+    self.tessel._rps.emit('close');
+  },
+
+  noCodeInFlash: function(test) {
```
rwaldron: Add: 

```
test.expect(1);
```

Otherwise, if `test.done()` is reached but no assertions have been executed nodeunit will consider that a "pass".

test/unit/tessel.js
johnnyman727: This was in response to some issues I ran into after merging your latest PR (see comments [here](https://github.com/tessel/t2-cli/pull/160)). Do you want me to modify this test to what you suggest above or do you want to do that in another PR?

test/common/tessel-simulator.js
johnnyman727: :+1: I will throw an error because exec is expecting an Array.

PR #166, opened 2015-06-23
Use pid.codes VID/PID

kevinmehall:


johnnyman727:
Will merge once https://github.com/tessel/t2-cli/pull/184 is merged so folks know how to update on their own.

PR #167, opened 2015-06-23
Adding all unit tests for provision

johnnyman727:
@rwaldron can you look this over?

### DIFF
--- a/lib/tessel/provision.js
+++ b/lib/tessel/provision.js
@@ -8,17 +8,18 @@ var async = require('async');
 var Tessel = require('./tessel');
 var commands = require('./commands');
 var logs = require('../logs');
-var dir = path.join(osenv.home(), '.tessel');
+var directory = path.join(osenv.home(), '.tessel');
 var filename = 'id_rsa';
-var filepath = path.join(dir, filename);
+var filepath = path.join(directory, filename);
+var remoteAuthFile = '/etc/dropbear/authorized_keys';
 
 Object.defineProperty(Tessel, 'TESSEL_AUTH_PATH', {
   get: function() {
-    return dir;
+    return directory;
   },
   set: function(value) {
-    dir = value;
-    filepath = path.join(dir, filename);
+    directory = value;
+    filepath = path.join(directory, filename);
   }
 });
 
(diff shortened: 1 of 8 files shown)


johnnyman727:
@rwaldron This was probably the most painful rebase ever. Could you look it over for me?

rwaldron:
```
rwaldron at nova in ~
$ t2 provision
INFO Connecting to Tessel...
INFO Connected to arnold over USB
INFO Creating public and private keys for Tessel authentication...
INFO SSH Keys written.
INFO Authenticating Tessel with public key...
INFO Tessel authenticated with public key.
/Users/rwaldron/clonez/t2-cli/node_modules/usb/usb.js:42
	this.__close()
	     ^
Error: Can't close device with a pending request
    at Error (native)
    at Device.usb.Device.close (/Users/rwaldron/clonez/t2-cli/node_modules/usb/usb.js:42:7)
    at USBConnection.<anonymous> (/Users/rwaldron/clonez/t2-cli/lib/usb_connection.js:151:17)
    at Device.<anonymous> (/Users/rwaldron/clonez/t2-cli/node_modules/usb/usb.js:185:7)
```

lib/tessel/provision.js
```diff
@@ -8,17 +8,18 @@ var async = require('async');
 var Tessel = require('./tessel');
 var commands = require('./commands');
 var logs = require('../logs');
-var dir = path.join(osenv.home(), '.tessel');
+var directory = osenv.home() + '/.tessel/';
```
rwaldron: `path.join(osenv.home(), '.tessel');`

johnnyman727:
@rwaldron thanks for pointing that out - this is a new bug that cropped up because I've started being vigilant about closing the connection once the task is done.

[Partially talking out loud for my own understanding]
The issue here is that the `USBConnection` inherits from `DuplexStream` so `_read` gets called over and over again automatically by the native Node Stream implementation. When `_read` gets called, it waits for the default number of bytes to get sent over from T2 before calling the callback with data.
```.js
USBConnection.prototype._read = function(n) {
  var self = this;

  if (self.closed) {
    return self.push(null);
  }

  self.epIn.transfer(4096, function(err, data) {
    if (err) {
      self.emit('error', err);
    } else {
      self.push(data);
    }
  });
};
```
Once the data is pushed, `_read` is called again and it starts waiting for more inbound USB data. That's why `usb` throws the error that a request is pending when the we try to close the connection.

We need a way to only have `_read` called when data is actually being waited on, or a way to cancel the request (is that possible @kevinmehall?)

johnnyman727:
Perhaps I should be using the [USB Streams](https://github.com/kevinmehall/node-usb#startstreamntransfers3-transfersizemaxpacketsize) interface.

johnnyman727:
Okay, I believe this was fixed. @kevinmehall can you check out https://github.com/tessel/t2-cli/commit/87b04cac813bf5c5795998b3c1b75cf3c676bf87 and make sure I didn't do anything stupid with USB?

test/unit/provision.js
```diff
@@ -82,12 +114,11 @@ exports['controller.provisionTessel'] = {
     this.provisionTessel().catch(function(error) {
       test.equal(error, 'Keys already exist. Refusing to overwrite them.');
       test.done();
-    });
+    }.bind(this));
```
rwaldron: ?

rwaldron:
:clap: 

```sh
rwaldron at nova in ~
$ rm -r .tessel/
rwaldron at nova in ~
$ t2 provision
INFO Connecting to Tessel...
INFO Connected to arnold over USB
INFO Creating public and private keys for Tessel authentication...
INFO SSH Keys written.
INFO Authenticating Tessel with public key...
INFO Tessel authenticated with public key.
rwaldron at nova in ~
$ t2 provision
WARN Keys already exist. Refusing to overwrite them.
rwaldron at nova in ~
$ t2 provision -f
INFO Connecting to Tessel...
INFO Connected to arnold over USB
INFO Creating public and private keys for Tessel authentication...
INFO SSH Keys written.
INFO Authenticating Tessel with public key...
INFO Tessel authenticated with public key.
```

PR #170, opened 2015-06-24
Unit tests for deploying code

johnnyman727:


test/unit/name.js
```diff
@@ -31,6 +31,7 @@ exports['Tessel.prototype.rename'] = {
   },
 
   tearDown: function(done) {
+    this.tessel.close();
```
johnnyman727: This fixed a memory leak where the process listener for SIGINT was never removed.

lib/controller.js
```diff
@@ -283,7 +283,10 @@ function deployScript(opts, push) {
   return Tessel.get(opts)
     .then(function(tessel) {
       // Run the script on Tessel
-      tessel.deployScript(opts, push);
+      return tessel.deployScript(opts, push)
+        .then(function() {
+          tessel.connection.end();
```
johnnyman727: We want to make sure we always close connections after we're done with them.

johnnyman727:
This is dependent on the USB fixes in #167. Otherwise, it's going to have the same errors when the connection is attempted to be closed.

test/unit/name.js
rwaldron: Nice!

test/common/remote-process-simulator.js
```diff
@@ -13,10 +14,27 @@ function RemoteProcessSimulator() {
 
   this.control._write = function(command, enc, cb) {
     // Emit commands for validation in tests
-    this.emit('command', command);
+    self.emit('command', command);
     // Call the callback so we can receive more
     cb();
-  }.bind(this);
+  };
+
```
rwaldron: This changeset creates a very minor conflict—easily resolved.

test/common/remote-process-simulator.js
```diff
@@ -13,10 +14,27 @@ function RemoteProcessSimulator() {
 
   this.control._write = function(command, enc, cb) {
     // Emit commands for validation in tests
-    this.emit('command', command);
+    self.emit('command', command);
```
rwaldron: When correcting the conflict mentioned below, don't forget to update this to `'control'`

johnnyman727: Ah good catch! Will fix.

johnnyman727:
@rwaldron thanks! I also made one more fix to delete the `funkytown` folder that existed after tests complete.

PR #171, opened 2015-06-24
Ensure TesselSeeker.prototype.stop always returns the same value.

rwaldron:
And a bunch of other tests


johnnyman727:
Looks good! Will merge.

PR #178, opened 2015-06-25
Update VID and PID.

rwaldron:
Ref: https://github.com/tessel/t2-firmware/commit/4b2f05e2bf97c6672fcc33b77846cb709a1655e3


johnnyman727:
@rwaldron this is a dup of #166, no?

rwaldron:
Weird, I don't know how/why I completely missed that 0_o

PR #182, opened 2015-06-27
[WIP] Prefer LAN connections

### LINKED ISSUE #162: Intelligent selection of Tessel connection to use
USB connections are fast to start up but slow to transfer data. WiFi connections can transfer large files much faster. We should have a way to choose between a given Tessel's connections based on the command being executed.

johnnyman727:
LAN connections are generally faster than USB connections. This PR does a couple of things:

- [x] After collecting available Tessels, combines USB and LAN connections into the same Tessel rather than treating them as distinct.
- [x] When `tessel.connection` is called on a Tessel with both USB and LAN connections, it returns the LAN connection. 
- [x] Fixes process completion and connection closing once deployment completes.
- [x] Adds tests for all of these functionalities.

Fixes #162.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -15,7 +15,7 @@ var timeoutOption = {
   abbr: 't',
   metavar: 'TIMEOUT',
   help: 'Set timeout in seconds for scanning for networked tessels',
-  default: 3
+  default: 5
 };
 
 parser.command('provision')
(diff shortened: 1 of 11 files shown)


rwaldron:
This works great so far.

johnnyman727:
:+1: Fantastic. I'll add some tests tomorrow and hopefully get it merged.

rwaldron:
@johnnyman727 if you have time to write one of the tests (just so I can see your general testing goals), I can fill in the rest

johnnyman727:
@rwaldron thanks for the offer! I honestly hadn't thought through the tests yet but I'll try to get one or more in tonight.

johnnyman727:
@rwaldron better late than never! two tests added.

rwaldron:
Awesome :) I will dig in and see what I can fill in for you

PR #185, opened 2015-06-27
Updates the init script to one that will work

johnnyman727:


johnnyman727:
Dependent on https://github.com/tessel/t2-firmware/pull/52.

PR #188, opened 2015-07-07
Sets default path if not provided to key setup

johnnyman727:


rwaldron:
Tests added!

johnnyman727:
Awesome, thanks a ton @rwaldron!

PR #192, opened 2015-07-07
lib: work around bug in mdns library

LinusU:
This works around kmpm/node-mdns-js#40 which otherwise swallows our errors making debugging a pain. :bug:

### DIFF
--- a/lib/lan_connection.js
+++ b/lib/lan_connection.js
@@ -130,21 +130,29 @@ Scanner.prototype.start = function() {
       try {
         // Start discovering Tessels
         self.browser.discover();
-      } catch (error) {
-        self.emit('error', error);
+      } catch (err) {
+        process.nextTick(function() {
+          self.emit('error', err);
+        });
       }
     });
 
     // When the browser finds a new device
     self.browser.on('update', function(data) {
-      // Check if it's a Tessel
-      if (_.findIndex(self.discovered, data) > -1) {
-        return;
+      try {
+        // Check if it's a Tessel
+        if (_.findIndex(self.discovered, data) > -1) {
+          return;
+        }
+
+        var connection = new LANConnection(data);
+
+        self.emit('connection', connection);
+      } catch (err) {
+        process.nextTick(function() {
+          self.emit('error', err);
+        });
       }
-
-      var connection = new LANConnection(data);
-
-      self.emit('connection', connection);
     });
   });
 };


LinusU:
I can't imagine this error being caused by my changes

> Error: Cannot find module 'node-pre-gyp'

LinusU:
We could wait with merging this and see if kmpm/node-mdns-js#41 gets merged, that would be the correct fix.

lib/lan_connection.js
```diff
@@ -130,21 +130,29 @@ Scanner.prototype.start = function() {
...
+          return;
+        }
+
+        var connection = new LANConnection(data);
+
+        self.emit('connection', connection);
+      } catch (err) {
+        process.nextTick(function() {
+          self.emit('error', err);
+        });
```
rwaldron: Neither call to `self.emit('error'...)` should be wrapped in a nextTick, because they already occur within `setImmediate`, so any calling code has already had a chance to create an `error` event handler in the last event turn.

LinusU: The `emit` function will issue a throw right away which will get caught by the faulty library.

[emit function, lib/events.js L70-L92](https://github.com/joyent/node/blob/master/lib/events.js#L70-L92)

rwaldron: Right, I'm familiar with the synchronous nature of emit and the EventEmitter class as a whole. Looking at the bug filed [here](https://github.com/kmpm/node-mdns-js/issues/40), and a larger portion of `lib/lan_connection.js` source [here](https://github.com/LinusU/t2-cli/blob/mdns-swallow-workaround/lib/lan_connection.js#L125-L157) the issue becomes clear. Thanks :)

johnnyman727:
It doesn't look like the mdns lib is going to fix this anytime soon.

@LinusU would you mind rebasing on master so Travis will pass?

LinusU:
@johnnyman727 Done, travis is running now...

rwaldron:
Awesome! I will review this ASAP

rwaldron:
Landed, plus tests: https://github.com/tessel/t2-cli/commit/2230b8eb483e66a0a0068fd748d1625fce9032c3

LinusU:
:+1: Thank you!

johnnyman727:
:clap: This will be a huge help for debugging.

LinusU:
This is now fixed upstream! We should revert this patch and bump the dependency

LinusU:
Well, as soon as a new release is published. I've asked for it thought

johnnyman727:
:+1: Cool! Thanks for following up on that @LinusU!

rwaldron:
Just to be clear, it's not as simple as reverting the patch. We'll need to remove the fixes and update the tests in a fresh changeset.

LinusU:
Well this pull request only contains those exact fixes, it should be reverted in it's entirely. But yes, in a fresh changeset and with the dependency bumped as soon as @kmpm publishes the next release to npm.

LinusU:
It's now released as `0.4.0`, I'll submit a pull request in a bit

Issue #196, opened 2015-07-11
error Windows_NT 6.3.9600 using npm link after cloning repository


dalestone:
Hi,

I have just cloned the repository and then run the following:

    npm link

but then I encounter this error:

```
19798 error Windows_NT 6.3.9600
19799 error argv "C:\\Program Files\\nodejs\\\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "link"
19800 error node v0.12.4
19801 error npm  v2.10.1
19802 error code ELIFECYCLE
19803 error tessel@0.3.23 postinstall: `tessel install-drivers || true; tessel trademark || true`
19803 error Exit status 1
19804 error Failed at the tessel@0.3.23 postinstall script 'tessel install-drivers || true; tessel trademark || true'.
19804 error This is most likely a problem with the tessel package,
19804 error not with npm itself.
19804 error Tell the author that this fails on your system:
19804 error     tessel install-drivers || true; tessel trademark || true
19804 error You can get their info via:
19804 error     npm owner ls tessel

```

Is there something I am doing incorrectly? Is it maybe an issue with the version of node I am running?

rwaldron:
Hi @dalestone! According to @othiym23, you'll need Administrator privileges (reference: https://github.com/npm/npm/issues/3861#issuecomment-67905700 ). Please confirm. Thanks!

dalestone:
@rwaldron  I can confirm setting permissions to "Full control"  on the node_modules folder worked. Thanks very much!

rwaldron:
If you'd like to contribute, you could add info to the readme that describes what you did :)

dalestone:
@rwaldron Yeah ok, I will write something up and do a pull request.

rwaldron:
@dalestone awesome—thanks!

PR #203, opened 2015-07-14
Added windows installation instructions

dalestone:
@rwaldron  Added installation instructions for windows users. I didn't know whether to leave the error message out? Let me know if you think it would be better without.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # t2-cli
-The starting point for the command line interface to the next version of Tessel
+The starting point for the Tessel 2 command line interface.
 
 Join the [conversation on Slack](https://tessel-slack.herokuapp.com/), our project's chat client!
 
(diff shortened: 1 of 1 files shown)


Frijol:
Thanks for doing this! I'll let someone with a windows computer try this out before merging but I'm really excited that you wrote this.

rwaldron:
@Frijol I asked @dalestone to contribute these changes since they solved his installation issue [here](https://github.com/tessel/t2-cli/issues/196) . The issue itself isn't Tessel specific and the fix is recommended by the npm team; should we wait for another confirmation or is that adequate? Either is fine with me

dalestone:
@Frijol @rwaldron  No problem happy to help! I am going to take a look at the t2-vm next as I've hit a few issues with trying to run that on Windows.

Frijol:
Okay, that seems pretty well tested then! Merging.

rwaldron:
@dalestone @Frijol thanks!

PR #204, opened 2015-07-16
CI: Adds stubs for tessel-classic, which avoids need to install usb driver

rwaldron:
This PR is just to trigger travis


rwaldron:
Green!!

johnnyman727:
I am so so so thankful that you figured this out!

PR #206, opened 2015-07-17
lib: provisioning error handling

LinusU:
Error handling in provisioning was missing some errors which caused problems in https://github.com/tessel/t2-vm/issues/24#issuecomment-122308893.

[feross/standard](https://github.com/feross/standard) has a rule that detects when you are not handling the `err` parameter in a callback. It's plenty useful, we should have something similar (or maybe switch to standard :+1:).

### DIFF
--- a/lib/tessel/provision.js
+++ b/lib/tessel/provision.js
@@ -56,14 +56,15 @@ function setupLocal(keyPath) {
     logs.info('Creating public and private keys for Tessel authentication...');
     // Generate SSH key
     keygen({}, function(err, out) {
-
+      if (err) {
+        return reject(err);
+      }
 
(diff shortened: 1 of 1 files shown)


rwaldron:
I tried to write tests for this but keygen is basically impossible to stub with sinon. So, instead of this I confirmed that the existing behavior is not broken, by walkng through the following smoke tests: 

- `t2 provision` with no existing ~/.tessel/
- `t2 provision` with existing ~/.tessel/
  - refused, as expected. 
- `t2 provision -f` with no existing ~/.tessel/ (intentional)
- `t2 provision -f` with existing ~/.tessel/

Everything worked as expected

PR #209, opened 2015-07-17
[WIP] lib: better ssh keypair library

LinusU:
This uses a library for generating ssh keypairs that links directly against openssl instead of invoking a different binary. Windows support is still kind of blocked on joyent/node#4051, unless the user want's to install openssl or we somehow bundle it.

I just wrote the other library, [LinusU/node-generate-rsa-keypair](https://github.com/LinusU/node-generate-rsa-keypair), which I have test for that makes sure it generates correct RSA keys. I haven't tested them with openssh yet.

I haven't tested this code at all, my tessel is a few kilometers away. Hopefully I'll have time for this tomorrow :smile:


LinusU:
So why I wanted to do this is to always stay in sync with the version of openssl that Node.js are providing. Since this is crypto I think that is the best way, certinely better than relying on a third party to provide built binaries of the `ssh-keygen` tool (the windows solution that I proposed earlier).

What I wrote in the other thread:

> Sorry for the late reply, I got stuck inside the spaghetti code and realized that this is the wrong way to solve this. We shouldn't depend on a third party library to ship up to date binaries when it comes to cryptography.

> Instead we should the version of openssl that Node.js provides for us, and that is exactly what I have done in tessel/t2-cli#209. Unfortunately this doesn't work on windows until joyent/node#4051 gets resolved.

> If you have any experience with compiling it would be really cool if you could take a look at [LinusU/node-generate-rsa-keypair](https://github.com/LinusU/node-generate-rsa-keypair) to see if you can get it to build on windows. It could work by installing openssl locally and pointing out the path to it, but ideally we really would like to link against the openssl that node provides for us. That way we will always stay in sync with them.

rwaldron:
Can you take a look at the failing tests?

johnnyman727:
@LinusU could you rebase on master? I've been fixing tests.

LinusU:
Rebased and the tests are now working :+1: 

I would like to provision my t2 just to see that it's working before merging this thought, although it should work.

johnnyman727:
@linusu if this successfully provisions will it be ready to merge? I have a new T2 coming in today that I can try this out on.

LinusU:
Absolutely, I just got back from traveling which is why I've been a bit more offline than usually. I'll still try to get it done in the nearest days but if it works for you then we are good to go! :+1:

johnnyman727:
@LinusU I gave this a try and unfortunately it doesn't work. I printed out the contents of my `authorized_keys` file on the T2 and the difference between the working and not working keys was pretty obvious:

Working:
```
ssh-rsa 
XXXXXXXXXXXXXXXX
```

Not Working (transferred by this library):
```
-----BEGIN PUBLIC KEY-----
XXXXXXXXXXXXXXXXXXX
-----END PUBLIC KEY-----
````

Where the Xs represent an encrypted key. I tried just removing the header and footer and adding `ssh-rsa` to the public key but still no dice.

LinusU:
Oh, I completely forgot about testing this...

Looks like I'll have to do some more work on it than :)

johnnyman727:
@LinusU mind if I close this for now to keep our PR tab clean? We currently have 12 open PRs that are in some state of completion.

LinusU:
Go ahead, I might open a new one if I get the time :)

Issue #216, opened 2015-07-20
t2 provision will throw an unhandled error event for all new users/devices


rwaldron:
A combination of these commits: 

- https://github.com/tessel/t2-cli/commit/42d8c293c2b4f122e6e12c7a8032e19114832277
- https://github.com/tessel/t2-cli/commit/8a9b67cf7ad27f05b4f287db8f1bd0e4196d01c8 

Has resulted in a bug that allows failures that were previously silent to crash the `t2 provision` process. Before, errors would be suppressed, but now they are emitted and if there is no handler, then the whole process fails. This error should be viewable by users, but only if a `verbose` flag has been explicitly provided. The reason for this suppression is because the `/.tessel/id_rsa` will never exist until the first use and we certainly don't want users seeing "File Doesn't Exist" errors the first time they plug in their new Tessel 2.

johnnyman727:
@rwaldron I can't repro this :/

rwaldron:
Did you delete the `~/.tessel` directory?

johnnyman727:
Yes - it may have been fixed by #221?

rwaldron:
Confirmed

PR #215, opened 2015-07-20
General Cleanup (Order: 1)

johnnyman727:
Goal: fix #104.

This PR will also add some better debugging capabilities for `t2 list`. You'll just need to type `DEBUG=discovery:* t2 list` or `DEBUG=discovery:lan t2 list` or `DEBUG=discovery:usb t2 list`.

PR #217, opened 2015-07-20
Implement verbose logging flag for all commands. Wrap `t2 provision` error in verbose mode. Fixes gh-216

rwaldron:


rwaldron:
This may be partly addressed in https://github.com/tessel/t2-cli/pull/215, so let's wait until that lands and I will see where this fits after that.

Gruntfile.js
```diff
@@ -87,4 +87,16 @@ module.exports = function(grunt) {
   // Default task.
   grunt.registerTask('default', ['test']);
 
+
+  // Support running a single test suite
+  grunt.registerTask('nodeunit:only', 'Run a single test specified by a target; usage: "grunt nodeunit:only:<module-name>[.js]"', function(file) {
```
johnnyman727: Oh thank god. I was just editing my grunt file to only run one file with `grunt nodeunit`. This is much better.

bin/tessel-2.js
```diff
@@ -172,13 +210,21 @@ parser.command('key')
...
   })
   .callback(function(opts) {
+    logs.verbose = opts.verbose;
     key(opts)
       .then(function() {
         logs.info('Key successfully generated.');
       })
       .then(closeSuccessfulCommand, closeFailedCommand);
+  })
+  .option('verbose', {
```
johnnyman727: It seems like many commands use these same settings. Could you refactor the 
`{
   default: false,
   flag: true,
   abbr: 'v',
    help: verbose
}`
object into it's own variable and assign it to each `'verbose'` option?

johnnyman727:
@rwaldron Looks good. I only have the one suggestion about the options var.

How do you feel like the verbose option usage differs from passing `DEBUG=` to the command? It seems like we might want to centralize on one debugging method. What do you think?

bin/tessel-2.js
rwaldron: Derp... why didn't I just do that to begin with??

johnnyman727: :)

rwaldron:
Updated. 


> How do you feel like the verbose option usage differs from passing DEBUG= to the command? It seems like we might want to centralize on one debugging method. What do you think?

I'm not sure actually... Let me think about that a bit more. In the meantime, I can move the grunt task to another PR

Frijol:
Any update on this?

johnnyman727:
@rwaldron should we close this PR? The one thing I liked about it was the ability to specify a nodeunit file to run.

johnnyman727:
@rwaldron closing this for now since we already have the Grunt single file tests in master. Feel free to re-open if you think this PR is relevant.

PR #219, opened 2015-07-22
[WIP] Properly closes Tessel connections after each use (Order: 3)

johnnyman727:
This is less of an issue for SSH connections, but the USB Daemon needs to be notified when connections are completed so it can remove that entry from its table. It can currently run 256 processes currently but if we aren't consistent about closing them, they can eventually crash the daemon with a fair amount of use.

- [x] Close discovered Tessels
- [x] Close Tessels used for each command
- [x] Write tests to confirm that they are closed


johnnyman727:
@rwaldron is killing it.

PR #221, opened 2015-07-23
[WIP] Fixing provisioning issues

### LINKED ISSUE #213: Improper provisioning logic
If you have two Tessels, and you provision the first, everything works as expected. However, if you then try to provision the second, the CLI will give a warning that the keys already exist and not provision it.

It should realize the keys already exist and continue to provision the second device.

### LINKED ISSUE #214: Can't provision Tessels already on LAN
If a Tessel is already connected to a LAN (perhaps someone else was using it), and you plug in USB to provision it, the provisioning will fail. This is because the discovery of Tessel prefers LAN devices and there is currently no good way to fetch a USB connection (if it's connected over both).

johnnyman727:
FIxes #213 and fixed #214. 

In the midst of fixing tests...

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -355,7 +355,8 @@ controller.provisionTessel = function(opts) {
             }
           });
         } else {
-          reject('Keys already exist. Refusing to overwrite them.');
+          // ~/.tessel exists with keys
+          resolve();
         }
       } else {
         // There is no ~/.tessel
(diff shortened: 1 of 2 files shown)


lib/controller.js
```diff
@@ -355,7 +355,8 @@ controller.provisionTessel = function(opts) {
             }
           });
         } else {
-          reject('Keys already exist. Refusing to overwrite them.');
+          // ~/.tessel exists with keys
+          resolve();
```
rwaldron: The tests needs to be updated for this.

johnnyman727: Yes, I'm in the midst of adding it and fixing the other tests but it's been more involved than I expected.

rwaldron: No worries, I can help get them updated.

Issue #180, opened 2015-06-27
Tessel List is inconsistent with which USB and LAN Tessels are available


johnnyman727:
I've been noticing that sometimes Tessels that are plugged in via USB or available with LAN (I can see the yellow connection light) aren't showing up with `t2 list`.

johnnyman727:
One way to reproduce the USB devices missing is to try to call `t2 run PATH_TO_CODE` and then cntl+c out of it before it finishes. Then run `t2 list` and the USB connection won't show up. 

Additionally, the `usb` node library takes about 11 seconds to load after I perform this test:
```.js
  console.log('starting');
  var a = new Date();
  var usb = require('usb');
  console.log('done', (new Date()-a) / 1000);
```
@kevinmehall any idea why this could be?

johnnyman727:
The LAN inconsistency is definitely due to a timing issue. Currently, the `list` command prints out devices that are found over the course of 3 seconds. The devices are usually found within the 3 seconds, but the SSH connection isn't always completely opened by the completion of the three seconds. We try to connect in order to test whether or not that Tessel was provisioned already or not. Perhaps the best way get fix this one is to keep track of these 'pending' connections and wait for them to complete.

johnnyman727:
A simple workaround for now that seems to work consistently is to set the `timeout` flag:
```t2 run {PATH} --timeout 5```

johnnyman727:
Fixed by #269.

PR #231, opened 2015-07-25
t2 update

johnnyman727:
- [x] Controller logic for parsing args and starting updates
- [x] Fetch current firmware/openWRT image version
- [x] Update openWRT image
- [x] Update SAMD21 image
- [x] Actually fetch from AWS instead of locally
- [x] Refactor to be able to switch between LAN and USB for a given Tessel
- [x] Polish logging, error message, tests, etc.


johnnyman727:
@kevinmehall Any suggestion for figuring out when the OpenWRT image is finished reflashing so we can configure the name and authorized_keys after updating? This will also be useful for https://github.com/tessel/t2-firmware/issues/29.

johnnyman727:
Depends on https://github.com/tessel/openwrt-tessel/pull/16.

johnnyman727:
Just as a note: I considered two ways of implementing image downloads from AWS. 

1. Tar up a folder that contains both the samd21 firmware and openwrt image and place that on AWS. The CLI downloads, creates a temp folder, untars, uploads into T2, then deletes the temp folder.

2. Create a folder on AWS with the samd21 firmware and openwrt binaries inside. The CLI downloads each buffer separately and uploads to T2.

I ended up going with #2 because I don't want to deal with managing a temporary folder (at the trade-off of a little more complexity on AWS)

LinusU:
@johnnyman727 [tar-stream](https://github.com/mafintosh/tar-stream) would be a perfect fit for skipping the temporary directory. I have tried it myself and it works superb :ok_hand:

johnnyman727:
Ooooh shiny! I'll give that a try.

johnnyman727:
@LinusU have ever used tar-stream with a stream coming in from the `request` library? I can't get a stream from a GET request to https://s3.amazonaws.com/builds.tessel.io/t2/firmware/9a85c84f5a03c715908921baaaa9e7397985bc7f.tar.gz ever uncompress without saying the file is corrupt (but I can download from AWS and unzip in Finder).

johnnyman727:
I also tried something to the effect of:
```.js
    request.get('https://s3.amazonaws.com/builds.tessel.io/t2/firmware/9a85c84f5a03c715908921baaaa9e7397985bc7f.tar.gz') , function(err, response, body) {
      // If there wasn't an issue with the request
          extract.on('entry', function(header, stream, callback) {
            // header is the tar header
            // stream is the content body (might be an empty stream)
            // call next when you are done with this entry

            stream.on('end', function() {
              callback() // ready for next entry
            })

            stream.resume() // just auto drain the stream
          })

          extract.on('finish', function() {
            // all entries read
          })

          var s = new stream.Readable();

          s.pipe(extract)
          s._read = function noop() {}; // redundant? see update below
          s.push(body);
          s.push(null);
    });
```

LinusU:
I think that the problem is that it's a gziped tar stream. You need to first pass it thru `Gunzip` (see `zlib` builtin module in Node.js).

Also, please don't buffer the entire body and then pass it thru to tar-stream. That would be an unnecessary speed and memory impact.

Something like this should work:

```javascript
var req = request.get('https://s3.amazonaws.com/builds.tessel.io/t2/firmware/9a85c84f5a03c715908921baaaa9e7397985bc7f.tar.gz')
var gunzip = zlib.createGunzip()
var extract = tar.extract()

extract.on('entry', function (header, stream, callback) {
  // Check header.name, decide what to do
})

req.pipe(gunzip).pipe(extract)
```

johnnyman727:
Oh gotcha! I had originally tried the elegant request streaming method before I posted that hackish example and had the same error. Now I understand that you have to unzip it *and* extract it.

bin/tessel-2.js
```diff
@@ -196,6 +196,29 @@ parser.command('rename')
   .option('timeout', timeoutOption)
   .help('Change the name of a Tessel to something new.');
 
+parser.command('update')
+  .option('build', {
```
Frijol: add a .help to each option

lib/controller.js
```diff
@@ -165,7 +166,9 @@ Tessel.get = function(opts) {
       } else {
         logs.info('Please specify a Tessel by name [--name <tessel name>]');
         controller.closeTesselConnections(connectionsToClose)
-          .then(reject);
+          .then(function() {
+            reject('Multiple possible Tessel connection found.');
```
Frijol: typo, should be "Multiple possible Tessel connections found"

lib/controller.js
```diff
@@ -452,6 +455,165 @@ controller.connectToNetwork = function(opts) {
...
+      // Otherwise, just resolve
+      else {
+        shaPromise = Promise.resolve();
+      }
+
+      // Once we have our Tessel and build version
+      return Promise.all([tesselPromise, shaPromise]).spread(function(tessel, currentVersion) {
+        // If it's not connected via USB, we can't update it
+        if (!tessel.usbConnection) {
+          return reject('Must have Tessel connected over USB to complete update. Aborting update.');
```
Frijol: Why is this?

lib/controller.js
```diff
@@ -452,6 +455,165 @@ controller.connectToNetwork = function(opts) {
...
+        }
+
+        // If this Tessel isn't connected to the LAN
+        if (!tessel.lanConnection || tessel.lanConnection.authorized) {
+          // Warn the user that it's going to take a while
+          logs.warn('Warning: Could not detect Tessel over LAN. Updates will be much slower over USB.');
```
Frijol: Better warning: "Warning: LAN connection not detected for this Tessel. For a faster update, cancel and connect Tessel to LAN with the `tessel wifi` command before update."

lib/controller.js
```diff
@@ -452,6 +455,165 @@ controller.connectToNetwork = function(opts) {
...
+        } else {
+          // If they have requested the latest firmware
+          var verifiedVersion;
+          if (requestedVersion === 'latest') {
+            // Figure out what the version is
+            verifiedVersion = updates.fetchBuildList()
+              .then(function(builds) {
+                // Make sure we actually have builds
+                if (Object.keys(builds).length === 0) {
+                  return reject('Could not detect any builds on server.');
```
Frijol: When might this happen? Does this mean they're not reaching the server? Seems like a good place to have a more helpful error message.

lib/controller.js
```diff
@@ -452,6 +455,165 @@ controller.connectToNetwork = function(opts) {
...
+              })
+              // The version does not exist
+              .catch(reject);
+          }
+          // Once we have our verified version
+          verifiedVersion.then(function(requestedVersion) {
+            // Check if the current build is the same or newer if this isn't a forced update
+            if (!opts.force && currentVersion >= requestedVersion) {
+              // If it's not, close the Tessel connection and print the error message
+              var errMessage = tessel.name + ' is already on the latest firmware build. You can force an update with “tessel update --force”.';
```
Frijol: "t2 update --force" not "tessel update --force". Would it be worthwhile to make "tessel/t2" a variable we include in strings of this type?

lib/controller.js
```diff
@@ -452,6 +455,165 @@ controller.connectToNetwork = function(opts) {
...
+        }
+      });
+    }
+  });
+};
+
+controller.updateTesselWithVersion = function(tessel, currentVersion, requestedVersion) {
+  // Fetch the requested build
+  return updates.fetchBuild(requestedVersion)
+    .then(function startUpdate(build) {
```
Frijol: How long does this process take? Do we need blinking lights (red + yellow flashing in sync was the T1 indicator) and/or logs during this process?

lib/tessel/update.js
```diff
@@ -0,0 +1,125 @@
...
+        });
+
+        // Once the process is complete
+        remoteProcess.once('close', function() {
+          // If we had an error
+          if (errContents.length) {
+            // This process failed
+            return reject(new Error(errContents));
+          }
+          // Otherwise we supposedly read the file
```
Frijol: "supposedly"?

lib/update-fetch.js
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/* 
+  Accepts a build name and attempts to fetch
+  the build images from the server. Returns build contents
+  in a Promise
+*/
+function fetchBuild(version) {
+  return new Promise(function(resolve, reject) {
+    logs.info('Beginning update download. This could take a couple minutes..');
```
Frijol: "a couple of minutes" - also, how long should we expect? Is there something we can watch to make sure it's not just frozen?

lib/usb_connection.js
```diff
@@ -180,6 +183,93 @@ USB.Connection.prototype._close = function(callback) {
...
+  var self = this;
+  return new Promise(function(resolve, reject) {
+    // Stop trying to read this endpoint
+    self.epIn.stopPoll(function() {
+      // Tell the Daemon to kill any remote processes associated with this connection
+      Daemon.deregister(self, function startBootloader(err) {
+        if (err) {
+          return reject(err);
+        }
+        // Tell the mcu to go into bootloade rmode
```
Frijol: typo bootloade r

lib/usb_connection.js
```diff
@@ -180,6 +183,93 @@ USB.Connection.prototype._close = function(callback) {
...
+        self.device.controlTransfer(VENDOR_REQ_OUT, REQ_BOOT, 0, 0, new Buffer(0), function(err) {
+          if (err) {
+            reject(err);
+          } else {
+            // Wait for it to tenter the mode
+            return self._waitUntilInBootloader()
+              .then(function createDFU(device) {
```
Frijol: Is this the right tabbing?

Frijol:
Hopefully helpful for your "polish" step. Other than that it looks great & I'm super excited to see this merged!

rwaldron:
I'm going to run the process here on the T2's I just received and will report back.

johnnyman727:
Thanks for the helpful comments, @Frijol!

@rwaldron testing is much appreciated. Make sure you run `t2 update -f`. If you don't force it, the CLI will try to check the version number of your T2s which is in a file that doesn't exist yet. 

Also, I plan on updating the build server with the most recent OpenWRT build (which includes the latest `tessel.js`.

rwaldron:
Ok, noted!

lib/controller.js
johnnyman727: The USB firmware update happens over dfu-util. We could eventually move this to the MediaTek but for now it happens on the host computer.

lib/controller.js
johnnyman727: Good suggestion!

lib/controller.js
johnnyman727: This *should* never happen. I guess it's possible if something gets mucked up on the build server by somebody. There isn't much the user can do except report it :/

lib/controller.js
johnnyman727: I think we should just type `tessel` for now since this will shortly be merged with the main `tessel` CLI. We have no other instances of mentioning `t2

lib/controller.js
johnnyman727: @Frijol the whole process takes ~3 minutes right now. Maybe a bit longer depending on connection speeds. We should definitely have blinky lights eventually. (Will make an issue for it).

lib/tessel/update.js
johnnyman727: I think I was just being snarky. We probably wrote the file :)

lib/update-fetch.js
johnnyman727: Yes, we can see how much data has been downloaded. This can be related to the LED blinking comment above.

johnnyman727:
@rwaldron if this works on your T2s, go ahead and merge it.

Frijol:
Please include a note on the Readme as well as one to https://github.com/tessel/t2-docs/blob/master/cli.md#updating-tessel-2-on-board-osfirmware

johnnyman727:
Note: this can't be merged yet - there is an existing issue with the tests that I need to work out.

rwaldron:
Now that the tesselignore stuff is finished, I will spend time with this tomorrow

rwaldron:
@johnnyman727 let me know when you're ready for a full review!

johnnyman727:
@rwaldron ready for a full review. I believe @Frijol already put some eyes on it but the more the merrier.

rwaldron:
I'm traveling today, so I'm on this tomorrow

rwaldron:
Using a T2 "out of the box/bag"...

1. Renamed to `bishop`
2. Set wifi credentials
3. Provisioned
4. Verify connection via ssh


### When attempting to update the T2 while it's in bootloader mode: 


```
rwaldron at nova in ~/clonez/t2-cli on jon-t2-update*
$ t2 list
/Users/rwaldron/clonez/t2-cli/node_modules/usb/usb.js:142
  this.interfaceNumber = this.descriptor.bInterfaceNumber
                                        ^
TypeError: Cannot read property 'bInterfaceNumber' of undefined
    at Interface.__refresh (/Users/rwaldron/clonez/t2-cli/node_modules/usb/usb.js:142:40)
    at Device.<anonymous> (/Users/rwaldron/clonez/t2-cli/node_modules/usb/usb.js:208:9)
```


Q: Should it _NOT_ be in bootloader mode? Either way, needs to be documented. 


Anyway, it's there...

```
$  dfu-util --list
dfu-util 0.8

Copyright 2005-2009 Weston Schmidt, Harald Welte and OpenMoko Inc.
Copyright 2010-2014 Tormod Volden and Stefan Schmidt
This program is Free Software and has ABSOLUTELY NO WARRANTY
Please report bugs to dfu-util@lists.gnumonks.org

Deducing device DFU version from functional descriptor length
Found Runtime: [05ac:8286] ver=0150, devnum=6, cfg=1, intf=3, alt=0, name="UNKNOWN", serial="UNKNOWN"
Found DFU: [1209:7551] ver=0002, devnum=5, cfg=1, intf=0, alt=1, name="SRAM", serial="XFJV95TJCWVS92NV53WF0QPXAW"
Found DFU: [1209:7551] ver=0002, devnum=5, cfg=1, intf=0, alt=0, name="Flash", serial="XFJV95TJCWVS92NV53WF0QPXAW"
```



### When attempting to update the T2, in "normal connection mode": 

```
$ t2 update -f
INFO Connecting to Tessel...
INFO Connected to bishop over USB
WARN Warning: Authorized LAN connection not detected for this Tessel. For a faster update, cancel and connect Tessel to LAN with the `tessel wifi` command before update.
INFO Updating bishop to latest build...
INFO Download complete!
INFO Updating OpenWRT (1/2)
INFO Transferring image of size 19.92 MB
```

Which is strange, because it's definitely connected via LAN _and_ USB. Also, this seems to imply a contradiction with: 

```
rwaldron at nova in ~/clonez/t2-cli on jon-t2-update*
$ t2 list
  bishop  LAN
rwaldron at nova in ~/clonez/t2-cli on jon-t2-update*
$ t2 update -f
INFO Connecting to Tessel...
INFO Connected to bishop over LAN
WARN Must have Tessel connected over USB to complete update. Aborting update.
```


Note: the T2 is definitely connected and the connection is sound.

lib/update-fetch.js
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  // If there was an issue with the server endpoint, reject
+  else if (response.statusCode !== 200) {
+    reject(new Error('Invalid status code on build server request: ' + response.statusCode));
+    return false;
+  }
+  // Otherwise
+  else {
+    return true;
```
rwaldron: This function is bad news :\ It's used as a [predicate](https://github.com/tessel/t2-cli/pull/231/files#diff-1ff7e3d75491fb02338063a4a45559fdR24), which is expected to return a boolean value (or some kind of coercible value), but may handle the rejection? 

Suggest: 

```js
function reviewResponse(response) {
  var outcome = {
    success: true, 
  };

  // If there was an issue with the server endpoint, reject
  if (response.statusCode !== 200) {
    outcome.success = false;
    outcome.reason = 'Invalid status code on build server request: ' + response.statusCode;
  }

  return outcome;
}
```

Then above: 

```js
function fetchBuildList() {
    // Fetch the list of available builds
    request.get(buildsListURL, function(err, response, body) {
      if (err) {
        return reject(err);
      }

      var outcome = reviewResponse(response);

      // If there wasn't an issue with the request
      if (outcome.success) {
        // Resolve with the parsed data
        try {
          resolve(JSON.parse(body));
        }
        // If the parse failed, reject
        catch (err) {
          reject(err);
        }
      } else {
        reject(outcome.reason);
      }
    });
  });
}
```

I'm sure that can even be further refined

rwaldron:
There appears to be several places where the build.json data is converted to an array before use. I think it would make much more sense if that data was already an array. So, given the current version of builds.json, which looks like this: 


```
{
  "78f2bd20af9eaf76796657186f3010f03a979dc8" : {
    "released" : "2015-08-20T16:20:08.704Z",
    "version" : "0.0.3"
  },
  "bf327359b4a13b4da07bc5776fe8a22ae88d54f9" : {
    "released" : "2015-08-18T15:12:13.070Z",
    "version" : "0.0.2"
  },
  "9a85c84f5a03c715908921baaaa9e7397985bc7f" : {
    "released" : "2015-08-12T03:01:57.856Z",
    "version" : "0.0.1"
  }
}
```

It should be refactored to look like this: 

```
[
  {
    "sha": "9a85c84f5a03c715908921baaaa9e7397985bc7f",
    "released" : "2015-08-12T03:01:57.856Z",
    "version" : "0.0.1"
  },
  {
    "sha": "bf327359b4a13b4da07bc5776fe8a22ae88d54f9",
    "released" : "2015-08-18T15:12:13.070Z",
    "version" : "0.0.2"
  },
  {
    "sha": "78f2bd20af9eaf76796657186f3010f03a979dc8",
    "released" : "2015-08-20T16:20:08.704Z",
    "version" : "0.0.3"
  }
]
```

Doing this is simple and would eliminate the following: 

1. any work currently required to convert the "dict" data into an array
2. any work currently required to locate the "latest" (because in the list form, the "latest" _must always_ the _last_ item)
3. the assignment of the sha as a value of property on the build data object—because it's already there. 

Finding the latest is as simple as: 

```js
builds[build.length - 1]; 
```

Find a build by version (or any data property, really): 

```js
// This assumes that t2-cli doesn't mind using es6-shim... and it shouldn't mind ;)
function findBuild(builds, property, value) {
  return builds.find(function(build) { return build[property] === value; });
}

// otherwise: 
function findBuild(builds, property, value) {
  return builds.filter(function(build) { return build[property] === value; })[0];
}
```

lib/controller.js
```diff
@@ -457,6 +460,167 @@ controller.connectToNetwork = function(opts) {
...
+  return new Promise(function(resolve, reject) {
+    // If the user requested a list of builds
+    if (opts.list) {
+      // Fetch those builds from the build server
+      logs.info('Fetching available builds...');
+      return updates.fetchBuildList()
+        .then(function(builds) {
+          logs.info('Latest builds:');
+          var maxBuildsListed = 10;
+          var SHAs = Object.keys(builds);
```
rwaldron: With my suggested change to the builds.json file, this would become: 

```js
// The last 10 are the 10 most recent builds
var SHAs = builds.slice(-10).map(function(build) { return build.sha; });
```

lib/controller.js
```diff
@@ -457,6 +460,167 @@ controller.connectToNetwork = function(opts) {
...
+            // Grab the specific build
+            var build = builds[SHAs[i]];
+            // Format the release time as a date
+            var released = new Date(build.released);
+            // Print when it was published
+            logs.basic('\t Version:', build.version, '\tPublished:', released.toLocaleString());
+          }
+          // Finish up
+          return resolve();
+        }, reject);
```
rwaldron: With my suggested change to the builds.json file, this whole operation (fetch and operation handler body) can be replaced with: 

```js
return updates.fetchBuildList()
  .then(function(builds) {
    logs.info('Latest builds:');
    
    builds.slice(-10).forEach(function(build) {
      logs.basic('\t Version:', build.version, '\tPublished:', released.toLocaleString());
    });
    // Finish up
    return resolve();
  }, reject);
```

lib/controller.js
```diff
@@ -457,6 +460,167 @@ controller.connectToNetwork = function(opts) {
...
+        if (opts.build) {
+          // Fetch the requested build
+          return controller.updateTesselWithVersion(opts.force, tessel, currentVersion, requestedVersion)
+            .then(resolve, reject);
+        } else {
+          // If they have requested the latest firmware
+          var verifiedVersion;
+          if (requestedVersion === 'latest') {
+            // Figure out what the version is
+            verifiedVersion = updates.fetchBuildList()
```
rwaldron: `updates.fetchBuildList()` Should only be called once per `tessel update ...`

lib/controller.js
```diff
@@ -457,6 +460,167 @@ controller.connectToNetwork = function(opts) {
...
+              .then(function(builds) {
+                // Make sure we actually have builds
+                if (Object.keys(builds).length === 0) {
```
rwaldron: With changes to builds.json, this becomes `builds.length === 0`

bin/tessel-2.js
rwaldron: `build` -> `version` because that's the name of the property being read for this operation and that's the word used in the `tessel update list` output

rwaldron:
Per our discussion, here is the refactoring with all the review notes: https://github.com/rwaldron/t2-cli/commit/c4b293a9cb8a43d2310a6431fa6257a0320c8882

README.md
```diff
@@ -43,14 +43,7 @@ There are two components to the Tessel on-board software: The OpenWRT Linux imag
 Unfortunately, [we have an open PR](https://github.com/tessel/t2-cli/pull/130) to detect the version of code running on T2 but it hasn't merged yet! If you aren't seeing your Tessel 2 show up with `t2 list`, then you probably have an old version of firmware (because the USB VID/PID is out of date). If you see functionality in the [Tessel API](https://github.com/tessel/t2-firmware#t2-hardware-api) that isn't defined on your board, then you probably need to update.
```
Frijol: Shouldn't this part change too?

johnnyman727:
@rwaldron Thanks for all these suggestions. I'm on board for all of them. 

To answer your first question - yes, T2 can't be in bootloader mode and use `t2 update` - otherwise it won't be found by the `usb` module. This can be changed in the future to look for both an "app" Tessel and a "bootloader" Tessel.

To your second point that you had T2 connected by USB and LAN but the CLI claimed it was only connected via USB - I've seen this too. I'm pretty sure it's a sympton of [this issue](https://github.com/tessel/t2-cli/issues/180). Essentially, when you tried `t2 update`, the CLI found your LAN connection but didn't finish opening it. We should fix this soon but I don't think this PR needs to wait for it.

The reason I made the builds.json use a list of build objects instead of a formal array was that I wanted to  make sure the CLI handled it properly even when it wasn't sorted (in case someone tries to upload a new build manually but isn't quite sure how it works), but we could handle that case anyway.

I'll review your PR now!

johnnyman727:
@rwaldron I've seen this Travis failure before but I can't seem to repro it on my own machine. I had the script running constantly for about an hour (it would have stopped if it failed) and it never hit this :/

rwaldron:
@johnnyman727 yeah, I'm not too worried about it. I just wanted to make sure I did a reality check locally before I pushed master

rwaldron:
Anyway, we're green!

PR #250, opened 2015-08-14
Move usage from here to the t2-docs repo

Frijol:
r? @johnnyman727


johnnyman727:
I think this should have a link to the actual documentation.

Frijol:
oh whoops thanks for catching that

johnnyman727:
Once that's fixed, it's good to go

PR #251, opened 2015-08-15
move update instructions back to here

Frijol:
@johnnyman727 I had moved this out when I moved instructions to the t2-docs repo, but I shouldn't have. These ones should stay here; we can update the other one to `t2 update` but people working on the CLI should know how to update their firmware manually.


johnnyman727:
looks good

PR #260, opened 2015-08-24
Close exited processes properly

### LINKED ISSUE #229: Processes not ending properly
CLI isn't telling Node to end the process at end of script or CTRL+C, so multiple processes stay open, resulting in ECONNREFUSED eventually

johnnyman727:
Fix #229.

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -257,20 +257,26 @@ actions.runScript = function(t, filepath, entryPoint) {
   return new Promise(function(resolve) {
     return t.connection.exec(commands.runScript(filepath, entryPoint))
       .then(function(remoteProcess) {
-        function scriptStopped() {
-          logs.info('Stopping script...');
-          resolve();
+
+        function echoSIGINT() {
+          // pass the SIGINT to the remote process
+          remoteProcess.stdout.signal('SIGINT');
         }
 
-        // Once we get a SIGINT from the console, stop the script
-        process.once('SIGINT', scriptStopped);
+        // Once we get a SIGINT from the console
+        process.once('SIGINT', echoSIGINT);
 
         // When the stream closes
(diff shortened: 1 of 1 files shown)


lib/tessel/deploy.js
```diff
@@ -225,8 +225,11 @@ actions.runScript = function(t, filepath, entryPoint) {
           resolve();
         }
 
-        // Once we get a SIGINT from the console, stop the script
-        process.once('SIGINT', scriptStopped);
+        // Once we get a SIGINT from the console
+        process.once('SIGINT', function() {
+          // pass the SIGINT to the remote process
+          remoteProcess.stdout.signal('SIGINT');
+        });
```
rwaldron: Should this actually be done in `scriptStopped`?

johnnyman727: When the `SIGINT` signal is sent to the remoteProcess, the remoteProcess then sends the CLI the signal that it has closed so the `remoteProcess.once('close')` event can fire (which executes the `scriptStopped` function). 

To your point, the log should probably be `Script stopped` rather than `stopping`.

johnnyman727: Also, we can just move the contents of `scriptStopped` to the `remoteProcess.once('close')` event because I don't think any other functions call it.

johnnyman727: Also also, I'll need to make the SIGINT handler non-anonymous so that [this line](https://github.com/tessel/t2-cli/pull/260/files#diff-191f4ebe176e9e816abc0445a664be98R237) can effectively remove the event listener.

johnnyman727:
@rwaldron when you get a chance could you give me the go-ahead to merge?

johnnyman727:
I've confirmed that this isn't actually needed by checking the logs from the USB Daemon. With this branch, the command to kill the process gets executed twice.

PR #261, opened 2015-08-24
Update refactor

rwaldron:
Note that the target is `jon-t2-update`

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -460,150 +460,114 @@ controller.connectToNetwork = function(opts) {
     });
 };
 
+
+
 controller.update = function(opts) {
   return new Promise(function(resolve, reject) {
-    // If the user requested a list of builds
-    if (opts.list) {
-      // Fetch those builds from the build server
-      logs.info('Fetching available builds...');
-      return updates.fetchBuildList()
-        .then(function(builds) {
-          logs.info('Latest builds:');
-          var maxBuildsListed = 10;
-          var SHAs = Object.keys(builds);
+    return updates.requestBuildList().then(function(builds) {
 
-          // Limited the number of builds we print out
-          SHAs = SHAs.splice(0, SHAs.length > maxBuildsListed ? maxBuildsListed : SHAs.length);
-          // Print out details for each build
-          for (var i = 0; i < SHAs.length; i++) {
-            // Grab the specific build
-            var build = builds[SHAs[i]];
-            // Format the release time as a date
-            var released = new Date(build.released);
-            // Print when it was published
-            logs.basic('\t Version:', build.version, '\tPublished:', released.toLocaleString());
-          }
-          // Finish up
-          return resolve();
-        }, reject);
-      // This is a standard update
-    } else {
-      // Fetch the latest version
-      var requestedVersion = 'latest';
-      // Unless a specific build was request
-      if (opts.build) {
-        // Then we'll fetch that build
-        requestedVersion = opts.build;
-      }
+      // tessel update -l
+      if (opts.list) {
+        logs.info('Latest builds:');
 
-      // Grab our Tessel
-      var tesselPromise = Tessel.get(opts);
-      var shaPromise;
-
-      // If we aren't forcing, we'll want to get the current SHA on Tessel 
-      if (!opts.force) {
-        // Once we have the Tessel
-        shaPromise = tesselPromise.then(function(tessel) {
-          // Figure out what commit SHA is running on it
-          return tessel.fetchCurrentBuildInfo()
-            // Once we have the current SHA
-            .then(function(currentSHA) {
-              // Figure out the human readable version for that code release
-              return updates.fetchBuildVersionforSHA(currentSHA);
-
-            });
+        builds.slice(-10).forEach(function(build) {
+          logs.basic('\t Version:', build.version, '\tPublished:', build.released.toLocaleString());
         });
-      }
-      // Otherwise, just resolve
-      else {
-        shaPromise = Promise.resolve();
-      }
+        // Finish up
+        return resolve();
+      } else {
 
-      // Once we have our Tessel and build version
-      return Promise.all([tesselPromise, shaPromise]).spread(function(tessel, currentVersion) {
-        // If it's not connected via USB, we can't update it
-        if (!tessel.usbConnection) {
-          return reject('Must have Tessel connected over USB to complete update. Aborting update.');
-        }
+        var version = opts.version || 'latest';
 
-        // If this Tessel isn't connected to the LAN
-        if (!tessel.lanConnection || !tessel.lanConnection.authorized) {
-          // Warn the user that it's going to take a while
-          logs.warn('Warning: Authorized LAN connection not detected for this Tessel. For a faster update, cancel and connect Tessel to LAN with the `tessel wifi` command before update.');
-        }
+        // Grab our Tessel
+        var tessel = Tessel.get(opts);
+        var sha = Promise.resolve();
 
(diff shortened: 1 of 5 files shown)


rwaldron:
"Test updates—fully green!"

Hahahahaha—except when running jshint. Whoops!!

lib/controller.js
```diff
@@ -460,150 +460,121 @@ controller.connectToNetwork = function(opts) {
...
+              var message = tessel.name + ' is already on the latest firmware version (' + currentVersion + '). You can force an update with "tessel update --force".';
+
+              logs.warn(message);
+
+              return controller.closeTesselConnections([tessel]).then(resolve);
             } else {
               if (!opts.force) {
                 // If it is a newer version, let's update...
-                logs.info('New firmware version found...' + requestedVersion);
+                logs.info('New firmware version found...' + version);
```
rwaldron: @johnnyman727 should this be `verifiedVersion`?

lib/controller.js
```diff
@@ -460,150 +460,121 @@ controller.connectToNetwork = function(opts) {
...
               }
 
-              logs.info('Updating ' + tessel.name + ' to latest build...');
-              // Fetch the requested build
-              return controller.updateTesselWithVersion(opts.force, tessel, currentVersion, requestedVersion)
+              logs.info('Updating ' + tessel.name + ' to latest version (' + version + ')...');
```
rwaldron: Same question, re: `verifiedVersion`

lib/controller.js
```diff
@@ -460,150 +460,121 @@ controller.connectToNetwork = function(opts) {
...
-                  return reject('Could not detect any builds on server.');
-                }
-                // Sort the builds by version
-                var buildArray = updates.buildDictToSortedArray(builds);
-
-                // Return the newest one
-                return buildArray[0].version;
-              });
+          // If it's not connected via USB, we can't update it
+          if (!tessel.usbConnection) {
```
johnnyman727: We should do this check before we download a build.

lib/update-fetch.js
```diff
@@ -1,27 +1,38 @@
...
 var extract = require('tar-stream').extract();
 var logs = require('./logs');
 var streamToBuffer = require('stream-to-buffer');
+
+
+var buildServerRoot = 'https://builds.tessel.io/t2';
+var firmwareBuildURL = urljoin(buildServerRoot, 'firmware');
+// THIS IS ONLY HERE FOR DEMONSTRATION, THE URL WILL CHANGE ONCE
+// THE BUILD DATA FORMAT IS UPDATED.
+var buildsListURL = 'https://dl.dropboxusercontent.com/u/3531958/builds.json';
```
rwaldron: @johnnyman727 we need to make sure that the actual `https://builds.tessel.io/t2/firmware/builds.json` is updated, then change this. I can make that change, just point me to it.

lib/controller.js
```diff
@@ -460,150 +460,121 @@ controller.connectToNetwork = function(opts) {
...
+          if (opts.version) {
+            // Fetch the requested version
+            return controller.updateTesselWithVersion(opts.force, tessel, currentVersion, build)
+              .then(resolve, reject);
+          } else {
+            // If they have requested the latest firmware
+            var verifiedVersion;
+            if (version === 'latest') {
+              build = builds[builds.length - 1];
+              version = verifiedVersion = build.version;
```
johnnyman727: Do we need both a `verifiedVersion` and `version` variable?

lib/controller.js
```diff
@@ -460,150 +460,121 @@ controller.connectToNetwork = function(opts) {
...
+            else {
+              build = updates.findBuild(builds, 'version', version);
+              if (build) {
+                version = build.version;
+                verifiedVersion = version;
+              }
+            }
+
+            if (!verifiedVersion) {
+              return reject();
```
johnnyman727: Should this `reject` have an error message?

rwaldron: That would be good, suggestion?

lib/controller.js
rwaldron: Probably not—let me find out

lib/controller.js
johnnyman727: "The requested build was not found. Please see the available builds with `tessel update -l`."

lib/update-fetch.js
johnnyman727: I've updated the `build.json` on the build server. It will take ~10 minutes to propogate but you can test with the non-Cloudfronted URL: https://s3.amazonaws.com/builds.tessel.io/t2/firmware/builds.json.

lib/controller.js
rwaldron: Yep. This will be reflected in the next commit.

lib/controller.js
rwaldron: Will be reflected in next commit

lib/controller.js
```diff
@@ -460,150 +460,121 @@ controller.connectToNetwork = function(opts) {
...
-        shaPromise = tesselPromise.then(function(tessel) {
-          // Figure out what commit SHA is running on it
-          return tessel.fetchCurrentBuildInfo()
-            // Once we have the current SHA
-            .then(function(currentSHA) {
-              // Figure out the human readable version for that code release
-              return updates.fetchBuildVersionforSHA(currentSHA);
-
-            });
+        builds.slice(-10).forEach(function(build) {
```
johnnyman727: These builds are in reverse order when I print them out. Perhaps we should sort them in `requestBuildList`?

lib/controller.js
rwaldron: Fixed this

lib/controller.js
rwaldron: Fixed this

lib/update-fetch.js
rwaldron: Awesome

lib/controller.js
rwaldron: Interesting, I will take a look. As long as the builds.json is in the right order, which it is, this shouldn't happen. I will get back to you asap.

rwaldron:
```
$ t2 update -l
https://builds.tessel.io/t2/firmware/builds.json
INFO Latest builds:
	 Version: 0.0.1 	Published: 2015-08-12T03:01:57.856Z
	 Version: 0.0.2 	Published: 2015-08-18T15:12:13.070Z
	 Version: 0.0.3 	Published: 2015-08-20T16:20:08.704Z
```

(I added the logging)

rwaldron: Would you prefer them to be printed out with latest first?

lib/controller.js
rwaldron: `findBuild` doesn't return a Promise, it's operating on data that we already have from `requestBuildsList` (which does return a Promise). Either way, I think you're right about _not_ doing any of this without that connection.

lib/controller.js
johnnyman727: I would personally prefer them to be printed with latest first. If you think that's non-standard, feel free to ignore!

lib/controller.js
rwaldron: We'll need to get the `tessel` first, which will take a bit of refactoring, but I think makes sense in the bigger picture

lib/controller.js
rwaldron: No, I think that's perfectly fine, just wanted to make sure I understood!

rwaldron:
Review items addressed, Travis is green, @johnnyman727 said " merge into my branch and merge my branch into master once you've handled those comments."... so here we go!

Issue #264, opened 2015-08-26
t2 restart <file> --type=flash|RAM, defaults to ?


rwaldron:
Often it's not necessary or desirable to do a complete rebundling of a Tessel device application and its dependencies. The `restart` command will skip bundling and simply restart the script, determined by the `--type` parameter (which defaults to `flash`)

```
# Restart a device application stored in RAM, default
t2 restart <file> 

# Restart a device application stored in RAM
t2 restart <file> --type=flash

# Restart a device application stored in Flash
t2 restart <file> --type=ram
```

Thoughts?

johnnyman727:
I think this is a great idea but it should default to `ram` (ie `tessel run`) because I think that's generally used more often (at least in my experience).

rwaldron:
I have those memory types backwards 0_o

rwaldron:
With this program: 

```js
var tessel = require("tessel");
var count = 0;
console.log("Device Running");

setInterval(function() {
  console.log(Date.now());

  tessel.led[0].toggle();

  if (++count === 4) {
    process.exit(0);
  }
}, 500);
```

```
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 run index.js
INFO Connecting to Tessel...
INFO Connected to bishop over LAN
INFO Writing index.js to RAM on bishop (4.096 kB)...
INFO Deployed.
INFO Running index.js...
Device Running
1440612170730
1440612171353
1440612171875
1440612172379
INFO Stopping script...
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 restart index.js
INFO Connecting to Tessel...
INFO Connected to bishop over USB
INFO Running index.js...
Device Running
1440612186344
1440612186968
1440612187490
1440612187993
INFO Stopping script...
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 push index.js
INFO Connecting to Tessel...
INFO Connected to bishop over LAN
INFO Writing index.js to Flash on bishop (4.096 kB)...
INFO Deployed.
INFO You may now disconnect from the Tessel. Your code will be run whenever Tessel boots up. To remove this code, use `tessel erase`.
INFO Running index.js...
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 restart index.js --type=flash
INFO Connecting to Tessel...
INFO Connected to bishop over USB
INFO Running index.js...
rwaldron at nova in ~/robotz/tessel-gps-monitor
```


Patch + tests to follow.

PR #265, opened 2015-08-26
t2 restart <file> --type=flash|ram, defaults to "ram". Fixes gh-264

rwaldron:


bin/tessel-2.js
```diff
@@ -49,6 +49,26 @@ parser.command('provision')
...
+  .option('entryPoint', {
+    position: 1,
+    required: true,
+    help: 'The entry point file to deploy to Tessel'
+  })
+  .option('type', {
+    default: 'ram',
+    help: 'Specify where in memory the script is located: `--type=flash` (push) or `--type=ram` (run)'
+  })
+  .help('Restart a script in RAM or Flash memory. (Does not rebundle)');
```
johnnyman727: I might add 'Restart a previously deploy script in RAM...'

lib/controller.js
```diff
@@ -380,12 +380,24 @@ controller.provisionTessel = function(opts) {
     });
 };
 
-controller.deployScript = function(opts, push) {
```
johnnyman727: Wasn't this changed in your previous PR that improved deploy? Or was this just overlooked on that one? It doesn't actually change the functionality for `restart` right?

lib/tessel/deploy.js
```diff
@@ -54,6 +54,26 @@ Tessel.prototype.deployScript = function(opts) {
   });
 };
 
+Tessel.prototype.restartScript = function(opts) {
+  var self = this;
+  var isPush = opts.type === 'flash';
+  var filepath = isPush ? PUSH_PATH : RUN_PATH;
+
+  return new Promise(function(resolve, reject) {
+    actions.findProject(opts.entryPoint).then(function(project) {
```
johnnyman727: What happens if there isn't a script in that directory?

test/unit/deploy.js
```diff
@@ -113,6 +113,68 @@ exports['Tessel.prototype.deployScript'] = {
...
+
+  tearDown: function(done) {
+    this.tessel.mockClose();
+
+    sandbox.restore();
+
+    done();
+  },
+
+  restartFromRam: function(test) {
```
johnnyman727: I suggest one more test that shows that an error is displayed if there is no project in the run or push directory.

lib/controller.js
rwaldron: I thought it was too, but it appears that I missed these two spots.

lib/tessel/deploy.js
rwaldron: Those errors should just be forwarded through as rejections—right? Right now it looks like USBProcess is just emitting an error event and nothing is handling that. I'm not sure this feature is the right place to fix that.

rwaldron: Also...

```
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 restart index.js
INFO Connecting to Tessel...
WARN No Tessels Found.
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 restart index.js
INFO Connecting to Tessel...
WARN No Tessels Found.
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 restart index.js
INFO Connecting to Tessel...
WARN No Tessels Found.
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 restart index.js
INFO Connecting to Tessel...
WARN No Tessels Found.
rwaldron at nova in ~/robotz/tessel-gps-monitor
$ t2 restart index.js
INFO Connecting to Tessel...
WARN No Tessels Found.
```

(same problem with any command)

bin/tessel-2.js
rwaldron: Fixed in next commit

lib/tessel/deploy.js
johnnyman727: Was this just after an update or a reboot? Or did it randomly disappear?

johnnyman727:
@rwaldron It seems like you shouldn't be specifying an entry point. To me, it's most intuitive to just run
```
t2 restart
```
or
```
t2 restart --push
```

When would you want to restart a script that wasn't the one that you just deployed?

lib/tessel/deploy.js
johnnyman727: I think we can remove the entrypoint argument and not check for a local folder before restarting on the remote device.

johnnyman727: I just tested what happens when the remote file doesn't exist and it appears to fail silently:
```
➜  t2-demo  t2 restart index.js --type=flash
INFO Connecting to Tessel...
INFO Connected to Tessel-02A337692559 over LAN
INFO Running index.js...
```

I think we should explicitly check that code exists in that remote directory prior to deployment and throw an error if it doesn't.

rwaldron: That's not what I'm seeing. I get the "module doesn't exist" failure. 

> I think we should explicitly check that code exists in that remote directory prior to deployment and throw an error if it doesn't.

We can get this _for free_ already, we just need to catch the error and reject with the message accordingly.

rwaldron:
@johnnyman727 

```
rwaldron at new-host-2 in ~/robotz/tessel-gps-monitor
$ t2 restart  index.js
INFO Connecting to Tessel...
INFO Connected to bishop over LAN
WARN "index.js" not found on bishop
```

rwaldron:
> When would you want to restart a script that wasn't the one that you just deployed?

There is no memory from one cli call to the next, so the name of the file that the user wants to restart is not concretely knowable. There could be several files in the root of a project directory that get deployed on either a `run` or `push`— it's impossible to know which one the user wants to restart, I can imagine deploying different versions in one `t2 run ...` and then restarting from one or the other. It would be super messy to look at all files in the remote directory and try to figure out which is the one to restart. What we _can_ do, is allow `t2 restart` and let it read the local package.json for a `main` entry; if the user needs to be explicit, they can be.

johnnyman727:
Love it! Tried it out and it works great. Feel free to merge!

PR #269, opened 2015-08-31
More reliable Tessel discovery

johnnyman727:
This PR makes Tessel discovery much more robust. Especially during `t2 update`, I was noticing that my LAN connection would come up on `t2 list` but when I ran `t2 update`, only my USB connection was found and that's what was used. USB connection takes forever for update so I'd basically need to run the command again.

The problem is that the CLI was just using a timeout for when Tessel discovery is complete. That leads to situations where a LAN connection is found but not opened before the timeout is hit.

This PR makes the discovery process wait for all pending connections to either open or fail before ending the search.

Additionally, I've pulled the default timeout back to 2 seconds because generally, you find the LAN connection within ~1s so 2s gives a nice margin of error. The fetch usually completes around 3s now.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -16,7 +16,7 @@ var timeoutOption = {
   abbr: 't',
   metavar: 'TIMEOUT',
   help: 'Set timeout in seconds for scanning for networked tessels',
-  default: 5
+  default: 2
 };
 
 function closeSuccessfulCommand() {
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -16,7 +16,7 @@ var timeoutOption = {
   abbr: 't',
   metavar: 'TIMEOUT',
   help: 'Set timeout in seconds for scanning for networked tessels',
-  default: 2
+  default: 5
 };
 
 function closeSuccessfulCommand() {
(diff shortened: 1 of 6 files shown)


johnnyman727:
Ready for review! @LinusU would you do me the honors?

LinusU:
Absolutely!

Code looks great, all tests passes on my local machine; but there is one problem.

For some reason it can't find my tessel over the LAN anymore, could this be because it isn't updated? I thought that I should run it before and after I updated to see that it works smooth.

I'll run the update now and see if it fixes it...

LinusU:
Just to better describe what happens; on `master` it lists both USB and LAN. When switching to `jon-discover-pending` it only shows USB. This is consistent.

LinusU:
So actually, if I increase the timeout to 10 it works. Maybe the default should be higher than 2?

@johnnyman727 Why did you change it down to `2` from `5`?

LinusU:
Yeah the timeout definitely the issue. When I change it back to 5 it works great.

*(sorry for the spam)*

johnnyman727:
@LinusU thanks for the review! I mostly moved it down to 2 to make the command seem faster (and it was consistently picking up my LAN device). I'll move it back to 5 seconds and merge.

PR #282, opened 2015-09-07
Add neutral wifi command to list Tessel Wifi Information

MrNice:
It does this:

```bash
~/s/t/bin ❯❯❯ /.nvm/versions/io.js/v2.2.1/bin/iojs tessel-2.js wifi
INFO Connecting to Tessel...
INFO Connected to pixel over USB
INFO Connected to "NETWORK"
INFO IP Address: 10.0.0.40
INFO Signal Strength: (61/70)
INFO Bitrate: 65mbps
```

Adds two commands, and a fairly painful method to wifi.js.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -197,7 +197,6 @@ parser.command('wifi')
         closeFailedCommand(new Error(msg));
       }
     } else {
-      console.log(opts);
       controller.getWifiInfo(opts)
         .then(closeSuccessfulCommand, closeFailedCommand);
     }
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -23,18 +23,21 @@ function closeSuccessfulCommand() {
   process.exit(0);
 }
 
-function closeFailedCommand(err) {
-  // If the returned value is an error
+// Allow options to be partially applied
+function closeFailedCommand(opts, err) {
+  if (!err) {
+    err = opts;
+    opts = {};
+  }
+
   if (err instanceof Error) {
-    // Throw it
     throw err;
+  } else {
+    // Print a stern warning by default
+    opts.type = opts.type || 'warn';
+    logs[opts.type](err);
   }
-  // Otherwise
-  else {
-    // Print a stern warning
-    logs.warn(err);
-  }
-  // Exit with non-zero code
+  // NOTE: Exit code is non-zero
   process.exit(1);
 }
 
@@ -186,12 +189,24 @@ parser.command('init')
 
 parser.command('wifi')
   .callback(function(opts) {
-    //TODO: Refactor switch case into controller.wifi
+    // TODO: Refactor switch case into controller.wifi
     if (opts.list) {
       controller.printAvailableNetworks(opts)
         .then(closeSuccessfulCommand, closeFailedCommand);
-    } else if (opts.ssid && opts.password) {
-      controller.connectToNetwork(opts)
+    } else if (opts.ssid || opts.password) {
+      if (opts.ssid && opts.password) {
+        controller.connectToNetwork(opts)
+          .then(closeSuccessfulCommand, closeFailedCommand);
+      } else {
+        var msg = opts.ssid ?
+          'Please provide a password with -p <password>' :
+          'Please provide a network name (SSID) with -n <name>';
+        closeFailedCommand({
+          type: 'err'
+        }, msg);
+      }
+    } else {
+      controller.getWifiInfo(opts)
         .then(closeSuccessfulCommand, closeFailedCommand);
     }
   })
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -23,19 +23,12 @@ function closeSuccessfulCommand() {
   process.exit(0);
 }
 
-// Allow options to be partially applied
-function closeFailedCommand(opts, err) {
-  if (!err) {
-    err = opts;
-    opts = {};
-  }
-
+function closeFailedCommand(err) {
   if (err instanceof Error) {
     throw err;
   } else {
     // Print a stern warning by default
-    opts.type = opts.type || 'warn';
-    logs[opts.type](err);
+    logs.warn(err);
   }
   // NOTE: Exit code is non-zero
   process.exit(1);
@@ -201,11 +194,10 @@ parser.command('wifi')
         var msg = opts.ssid ?
           'Please provide a password with -p <password>' :
           'Please provide a network name (SSID) with -n <name>';
-        closeFailedCommand({
-          type: 'err'
-        }, msg);
+        closeFailedCommand(new Error(msg));
       }
     } else {
+      console.log(opts);
       controller.getWifiInfo(opts)
         .then(closeSuccessfulCommand, closeFailedCommand);
     }
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -457,7 +457,6 @@ controller.getWifiInfo = function(opts) {
       return selectedTessel.getWifiInfo()
         .then(function(network) {
           // Grab inet lines, flatmap them, remove empty
-          // Wanted to do this with awk and cut inside commands.js
           var ips = network.ips.filter(function(item) {
               return /inet/.exec(item);
             })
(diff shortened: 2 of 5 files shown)


bin/tessel-2.js
```diff
@@ -23,18 +23,21 @@ function closeSuccessfulCommand() {
   process.exit(0);
 }
 
-function closeFailedCommand(err) {
-  // If the returned value is an error
+// Allow options to be partially applied
+function closeFailedCommand(opts, err) {
```
johnnyman727: @MrNice what's the rationale modifying this function? You could previouslyt report an error if you pass it an error or you can have it warn the user if it's just a string

bin/tessel-2.js
```diff
@@ -186,12 +189,24 @@ parser.command('init')
...
   .callback(function(opts) {
-    //TODO: Refactor switch case into controller.wifi
+    // TODO: Refactor switch case into controller.wifi
     if (opts.list) {
       controller.printAvailableNetworks(opts)
         .then(closeSuccessfulCommand, closeFailedCommand);
-    } else if (opts.ssid && opts.password) {
-      controller.connectToNetwork(opts)
+    } else if (opts.ssid || opts.password) {
+      if (opts.ssid && opts.password) {
```
johnnyman727: If we move the SSID and Password check here (which makes sense so you don't have to wait for it to connect to a Tessel) we might as well remove the check that already exists here: https://github.com/tessel/t2-cli/pull/282/files#diff-ad1c3af214118b233a97e4dd4f67d729R127

lib/tessel/wifi.js
```diff
@@ -12,9 +16,7 @@ Tessel.prototype.findAvailableNetworks = function() {
 
         var resultsJSON = '';
 
-        remoteProcess.stderr.on('data', function(d) {
-          logs.err(d.toString());
-        });
+        remoteProcess.stderr.on('data', logErr);
```
johnnyman727: What we should do here (and below in `getWiFiInfo` is concatenate the stderr data as it comes in. Once the process is closed, we should check if the stderr data has any length, and if it does, reject the function with that string. See an example [here](https://github.com/tessel/t2-cli/blob/47219b6d28a6db9871ef12ff65203055a152edd4/lib/tessel/update.js#L24-L39). There are a couple of other points where we *don't* do this and we should start being more disciplined about it.

lib/controller.js
```diff
@@ -451,6 +451,45 @@ controller.printAvailableNetworks = function(opts) {
...
 
+controller.getWifiInfo = function(opts) {
+  return Tessel.get(opts)
+    .then(function(selectedTessel) {
+      return selectedTessel.getWifiInfo()
+        .then(function(network) {
+          // Grab inet lines, flatmap them, remove empty
+          // Wanted to do this with awk and cut inside commands.js
+          var ips = network.ips.filter(function(item) {
+              return /inet/.exec(item);
```
johnnyman727: What happens if the Tessel isn't connected to WiFi?

johnnyman727:
@MrNice thanks for the contribution! It works really well when Tessel is connected. I added a few implementation notes on the relevant lines.

In addition, the `t2 wifi` command should report that it's not connected to WiFI according to the spec. Currently, it throws an angry error:
```
➜  t2-cli git:(add-wifi-command) t2 wifi
INFO Connecting to Tessel...
INFO Connected to Harold over USB
Unhandled rejection TypeError: Reduce of empty array with no initial value
    at Array.reduce (native)
    at /Users/Jon/Work/technical/t2-cli/lib/controller.js:467:14
    at tryCatcher (/Users/Jon/Work/technical/t2-cli/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/Jon/Work/technical/t2-cli/node_modules/bluebird/js/main/promise.js:503:31)
    at Promise._settlePromiseAt (/Users/Jon/Work/technical/t2-cli/node_modules/bluebird/js/main/promise.js:577:18)
    at Promise._settlePromises (/Users/Jon/Work/technical/t2-cli/node_modules/bluebird/js/main/promise.js:693:14)
    at Async._drainQueue (/Users/Jon/Work/technical/t2-cli/node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (/Users/Jon/Work/technical/t2-cli/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/Jon/Work/technical/t2-cli/node_modules/bluebird/js/main/async.js:15:14)
    at processImmediate [as _immediateCallback] (timers.js:367:17)
```

Lastly, you'll have to run `grunt` when you finish writing your code. That will run our JSHinter, check for any styling issues and run our unit tests. It looks like this commit caused one of the unit tests to fail.

Let me know if you have any questions!

johnnyman727:
Actually, I can't reproduce the failure Travis detected... I'm restarting the test to check if it's an intermittent type of issue.

MrNice:
Re: Travis: last night it bounced this PR for jshint reasons, so I added 15
semicolons, but then it failed on a unit test that I couldn't reproduce
locally.

That specific error with the reduce will be fixed if its just passed an
array literal.

I'll work on this stuff when I get home.

johnnyman727:
Ah the Travis issue is directly related to it being a build server with no USB peripherals. Somehow a path of code is running that attempts to get a USB Device List (not sure why this particular PR caused it though). I can look further into it later today.

bin/tessel-2.js
MrNice: Undoing, sorry.

lib/controller.js
```diff
@@ -451,6 +451,45 @@ controller.printAvailableNetworks = function(opts) {
...
+            .filter(function(item) {
+              return /addr/.exec(item);
+            })
+            .map(function(chunk) {
+              return chunk.split(':')[1];
+            })
+            .filter(function(addr) {
+              return addr.length;
+            });
+
```
rwaldron: I'd like to see this consolidated. The last three could be a single `reduce`

rwaldron:
In addition to @johnnyman727's review points and fixing the existing failure, this needs new tests to cover the new behavior.

lib/controller.js
MrNice: If the intent of the code is clear, why consolidate

I can replace the last three with the following
```js
.reduce(function(list, ip) {
  if (/addr/.exec(item)) {
    var ip = chunk.split(':')[1];
    if (ip.length) list.push[ip];
  }
  return list;
});
```

But it doesn't make it more clear (to me) what is happening.

lib/tessel/wifi.js
MrNice: Why isn't there a callback / promise API for remoteProcess? I'm only waiting on a single value.

lib/controller.js
rwaldron: :/

This is what `reduce` is for: consolidating map/filter operations.

MrNice: I'll change it if that's wanted, but I don't think it's clearer.

MrNice:
@johnnyman727 

Is this a reasonable error message?
```bash
~/s/t/bin ❯❯❯ iojs tessel-2.js wifi
INFO Connecting to Tessel...
INFO Connected to pixel over USB
ERR! This Tessel is not connected to Wi-Fi
```

johnnyman727:
@MrNice the issue with the spec suggests:
```
Not Connected (run 'tessel wifi -l' to see available networks)
```

I'd suggest one modification:
```
{TESSEL_NAME} is not connected to WiFi (run 'tessel wifi -l' to see available networks)
```

johnnyman727:
@MrNice did you need any more help with this or are you all good?

MrNice:
Oh I'm rushing around busy. Got those other 2 pr's to wrap up first :(

johnnyman727:
No worries - just wanted to make sure you weren't hung up on anything from me.

johnnyman727:
@MrNice can I help merge this?

MrNice:
@johnnyman727 Wrapping this up and adding tests. Did we ever figure out travis?

bin/tessel-2.js
johnnyman727: @MrNice On second thought, I take this back. Once we make this CLI more usable from other Node programs, it would be great if we could export functions so having this extra check doesn't hurt. In fact, because this check is removed they are causing [this test](https://github.com/tessel/t2-cli/pull/282/files#diff-7275035954a27dab20f24918ba5bea6aR146) and [this test](https://github.com/tessel/t2-cli/pull/282/files#diff-7275035954a27dab20f24918ba5bea6aR160) to fail right now.

johnnyman727:
Closing in favor of rebased and fixed https://github.com/tessel/t2-cli/pull/362.

PR #285, opened 2015-09-07
lib: remove workaround for mdns library bug

LinusU:
This workaround is no longer required since this have been fixed upstream in `0.4.0`

@johnnyman727 care to review?

### DIFF
--- a/lib/discover.js
+++ b/lib/discover.js
@@ -52,6 +52,7 @@ TesselSeeker.prototype.start = function(timeout) {
     // Push this Promise into the pending array
     pendingOpen.push(opening);
   }
+
   // If a timeout was provided
   if (timeout && typeof timeout === 'number') {
     // Set a timeout function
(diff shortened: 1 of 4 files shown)


rwaldron:
@LinusU can you take a look at the travis-ci failures? You can run these all locally:

```
npm install # one time set up
```

Then run: 

```
grunt
```

rwaldron:
Probably just need to swap these around: 

- https://github.com/tessel/t2-cli/blob/master/test/unit/lan_connection.js#L108-L115
- https://github.com/tessel/t2-cli/blob/master/test/unit/lan_connection.js#L155-L164

LinusU:
Yes, that's probably it, thanks! I should have run the tests before I committed, don't know why I didn't... I'll fix this :)

johnnyman727:
@LinusU I made it very easy: https://github.com/LinusU/t2-cli/pull/1 :)

johnnyman727:
Thanks @LinusU!

PR #286, opened 2015-09-07
Automatically resolves wifi info so no Tessel is fetched

johnnyman727:
Grunt is still not passing this and I'm not sure why. But it will take care of the previous issue we were seeing with Travis because the `getWifiInfo` was not stubbed so it was attempting to fetch a real Tessel.

@rwaldron or @MrNice could you look at this Grunt test? I can't figure out what the issue might be. It looks like if you comment out the `getWifiInfo` stub, it passes (except you'd get the same `getDeviceList` error on Travis if you pushed it.


johnnyman727:
Will close this and open on master.

Issue #235, opened 2015-07-25
Run `tessel list` to let users specify a tessel if nothing else selected


Frijol:
If there is no tessel specified by:
* --name
* an environment variable in the terminal, set as `export TESSEL=Bulbasaur`
* a single tessel connected over USB
* only one tessel visible

run `tessel list` and prompt the user to select.

Moved from https://github.com/tessel/t2-cli/issues/93 to de-prioritize

johnnyman727:
Will be fixed with #348.

PR #290, opened 2015-09-09
T2 root

Student007:
### Purpose
**This pull request will add the `t2 root` command to the t2-cli.**

### Side effects
- test/unit/deploy.js: console.log(\'testing deploy\') // due to frequently linter bleating
- same in  test/common/tessel-classic.js
- changes of test/unit/provision.js are caused by automatic linter activities ...
- test/unit/usb_connection.js is changed due to changes in lib/usb_connection.js (err -> warn)

### New Files

-  lib/root_controller.js 

Everything else is located in bin/controller.js

### New Test files
- test/unit/menu.js -> unit tests for Menu creation
- test/unit/seekTessels.js -> unit tests for seeking Tessels with root command
- test/unit/root.js -> root ssh without menu (single Tessel found)

**Note:** Tests run successfully on all reference machines (Mac OS X Yosemite, Windows 7 Professional, Ubuntu 14.04)

### Fixme
There is no handler for managing a ssh-connected Tessel is powering off ! Needs to be fixed within the firmware if necessary... (Terminal freezes imediatelly after poweroff the (virtual) Tessel)

### Improvements
The style of the menu needs a little bit love. It is reduced to functional working mode due to expectation it maybe would be replaced by a more complex and useful terminal-menu what is also providing realtime data of Tessels are found.

### Usabiliy
Due to some mistakes I did during my journey/while learning curve .... there are some feedbacks/tips and a debug mode:
- netmask is the limit 
- network topology could cause a single Tessel is found twice (this should be added to the list cmd also)

### TODO
Because I do not own at least one Tessel 2 ... I am not able to check how it works with multiple physical Tessels :crying_cat_face:


bin/tessel-2.js
```diff
@@ -197,6 +197,36 @@ parser.command('rename')
...
+$ 
+
+$ t2 root
+> Accessing root...
+root@192.168.128.124 # 
+*/
+var default_id_rsa = '~/.tessel/id_rsa';
+var functional_msg = '\nGain SSH root access to one of your authorized tessels (menu listing if multiple targets)';
+parser.command('root')
+  .usage(functional_msg + '\n\nUsage: t2 root [-i PATH] [--help]\n\n-i PATH:   Optional targeting a different Private Key \n\n(Note: default target created by "t2 key generate" is ' + default_id_rsa + ')\n')
```
rwaldron: @johnnyman727 eventually we're going to need to put these icky strings into template files and load them in at runtime

lib/controller.js
```diff
@@ -170,6 +177,134 @@ Tessel.get = function(opts) {
     }
   });
 };
+/*
+Because of sync problem with master trunk and parallel development on Tessel.list 
+it was necessary to go my own way...
```
rwaldron: Why not just pull & rebase as development proceeded?

lib/controller.js
```diff
@@ -170,6 +177,134 @@ Tessel.get = function(opts) {
...
+    var seeker = new discover.TesselSeeker().start();
+    var tessels = [];
+
+    // When we find Tessels
+    seeker.on('tessel', function(tessel) {
+      controller.closeTesselConnections(tessel)
+        .then(function() {
+          var known = false;
+          if (tessels.length >= 1) {
+            for (var i in tessels) {
```
rwaldron: Don't use `for-in` to iterate arrays, use either regular `for` loop or `tessels.forEach`

lib/controller.js
```diff
@@ -170,6 +177,134 @@ Tessel.get = function(opts) {
...
+  The structure of code and maybe unusual parts are paid due being testable.
+  The lightweight terminal-menu isn't written testable what needs a little bit hacking.
+
+  Finally some parts are causing from my personnal learning curve about writing unit tests using sinon!
+*/
+controller.root = function(opts) {
+  // ~ conversion to home because spawn isn't able to handle this right
+  if (opts.path && opts.path.substring(0, 1) === '~') {
+    var home = process.env.HOME;
+    opts.path = opts.path.replace('~', home);
```
rwaldron: This will break if there is a `~` anywhere else in the path. 

```
rwaldron at new-host-3 in ~
$ mkdir -p foo/ba~r/baz
rwaldron at new-host-3 in ~
$ cd foo/
rwaldron at new-host-3 in ~/foo
$ ls
.
└── [drwxr-xr-x rwaldron staff     102 Sep  9 12:52]  ba~r/

1 directory, 0 files
rwaldron at new-host-3 in ~/foo
$ cd ba~r/
rwaldron at new-host-3 in ~/foo/ba~r
$ ls
.
└── [drwxr-xr-x rwaldron staff      68 Sep  9 12:52]  baz/

1 directory, 0 files
rwaldron at new-host-3 in ~/foo/ba~r
$ cd baz/
rwaldron at new-host-3 in ~/foo/ba~r/baz
$ ls
.

0 directories, 0 files
```

Please use the method shown here: https://github.com/tessel/t2-cli/blob/master/lib/tessel/deploy.js#L105-L123

lib/root_controller.js
```diff
@@ -0,0 +1,174 @@
+var Tessel = require('./tessel/tessel');
+var logs = require('./logs');
+var debug = require('debug')('controller');
+
+// create an object for t2 root steps for beeing more handy about unit testing of the T2 root command itself.
+var controller = {};
+controller.ssh = {
```
rwaldron: This should be merged into `controller.js`

lib/root_controller.js
```diff
@@ -0,0 +1,174 @@
...
+    debug('runSSH...\n root@' + tessels[id].connections[0].host);
+    if (opts.menu) {
+      // because the test needs to resolve the promise ...
+      resolve({
+        opts: opts,
+        tessels: tessels
+      });
+    } else {
+      // if this is no test, starting the menu as child process
+      process.stdout.write('\u001b[2J\u001b[0;0H');
```
rwaldron: `'\u001b[2J\u001b[0;0H'` occurs 6 times in this file—please store in a variable and include a comment for explanation

lib/root_controller.js
```diff
@@ -0,0 +1,174 @@
...
+    tessels.push({
+      connection: {
+        connectionType: 'exit'
+      }
+    });
+
+    rtm.reset();
+    rtm.write('                                                \n');
+    // create menu entries
+    for (var i in tessels) {
```
rwaldron: Same as above, don't use `for-in` for array iteration

lib/usb_connection.js
```diff
@@ -105,7 +105,7 @@ USB.Connection.prototype.open = function() {
       self.device.open();
     } catch (e) {
       if (e.message === 'LIBUSB_ERROR_ACCESS' && process.platform === 'linux') {
-        console.error('Please run `sudo tessel install-drivers` to fix device permissions.\n(Error: could not open USB device.)');
+        logs.warn('Please run `sudo tessel install-drivers` to fix device permissions.\n(Error: could not open USB device.)');
```
rwaldron: Nice catch, but these should be addressed in a different issue & PR

test/common/tessel-classic.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     if (dir === 'fail') {
       callback(new Error());
     } else {
-      callback(null, new Buffer('console.log("testing deploy");'));
+      callback(null, new Buffer('console.log(\'testing deploy\');'));
```
rwaldron: Instead of `\'`, just use `"`

test/unit/deploy.js
```diff
@@ -9,7 +9,7 @@ var mkdirp = require('mkdirp');
 var path = require('path');
 var deployFolder = path.join(__dirname, 'tmp');
 var deployFile = path.join(deployFolder, 'app.js');
-var codeContents = 'console.log("testing deploy");';
+var codeContents = 'console.log(\'testing deploy\');';
```
rwaldron: Undo this change

test/unit/menu.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+  tessel.serialNumber = serialNumber(options);
+  tessel.name = options.name || 'a';
+
+  options.sandbox.stub(tessel, 'close', function() {
+    return Promise.resolve();
+  });
+  // return the fake tessel
+  return tessel;
+}
```
rwaldron: This should be rolled into `TesselSimulator` https://github.com/tessel/t2-cli/blob/master/test/common/tessel-simulator.js#L4-L29

test/unit/menu.js
```diff
@@ -0,0 +1,272 @@
...
+        this.y++;
+      };
+    });
+    this.sandbox.stub(this.rtm, 'write', function write() {
+      this.write = function() {
+        return this;
+      };
+    });
+    this.sandbox.stub(this.rtm, '_draw', function _draw() {
+      this._draw = function() {
```
rwaldron: What is `this` for `write` (line 67) and `_draw`?

test/common/tessel-classic.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
...
     }
   }
 };
```
rwaldron: Incidentally, this file no longer exists in the project.

test/unit/usb_connection.js
```diff
@@ -155,7 +155,7 @@ exports['USB.Connection.prototype._receiveMessages'] = {
 
     // The immediate return prevents these from being
     // called from in the error handler
-    test.equal(this.err.callCount, 0);
+    test.equal(this.warn.callCount, 0);
     test.equal(this.processExit.callCount, 0);
     test.done();
   },
```
rwaldron: These changes should be made in a different issue & PR

test/unit/provision.js
```diff
@@ -121,18 +121,18 @@ exports['controller.provisionTessel'] = {
...
+      .then(function() {
+        test.equal(this.exec.callCount, 1);
+        test.equal(this.exec.lastCall.args[0], 'rm -r ' + Tessel.TESSEL_AUTH_PATH);
+        test.equal(this.provisionSpy.callCount, 1);
+        rimraf(path.join(process.cwd(), Tessel.TESSEL_AUTH_PATH), function(err) {
+          test.ifError(err);
+          Tessel.TESSEL_AUTH_PATH = tesselAuthPath;
+          test.done();
+        });
+      }.bind(this));
```
rwaldron: I just ran `grunt jsbeautifier` and it did not make this change, please undo

bin/tessel-2.js
johnnyman727: @Student007 there is no need to add the `usage` function here. `nomnom` takes care of it for you as long as you add `help` tags to all the parameters. Check out the other commands for an example.

@rwaldron ugh agreed!

bin/tessel-2.js
```diff
@@ -197,6 +197,36 @@ parser.command('rename')
...
+  .option('path', {
+    abbr: 'i',
+    full: 'path',
+    metavar: 'PATH',
+    default: default_id_rsa,
```
johnnyman727: @Student007 this path is already used in another part of the codebase (probably `lan_connection.js`). We should have this defined in only one place that's accessible to both files.

lib/controller.js
```diff
@@ -22,21 +26,22 @@ Tessel.list = function(opts) {
...
       }
-
       // Print out details...
       logs.basic(sprintf('\t%s\t%s\t%s', tessel.name, tessel.connection.connectionType, note));
     });
 
     // Called after CTRL+C or timeout
     function stopSearch() {
+      if (foundTessels.length > 1) {
+        debug('heuristics ? => found: ' + foundTessels.length + ' === 1 || (' + foundTessels.length + ' === 2 && ' + foundTessels[0].name + ' === ' + foundTessels[1].name + ' && ' + foundTessels[0].connections[0].connectionType + ' !== ' + foundTessels[1].connections[0].connectionType + ')');
```
johnnyman727: @Student007 what is this debug line for?

lib/controller.js
```diff
@@ -22,21 +26,22 @@ Tessel.list = function(opts) {
...
       var note = '';
-
       // Add it to our array
       foundTessels.push(tessel);
-
       // Add a note if the user isn't authorized to use it yet
       if (tessel.connection.connectionType === 'LAN' && !tessel.connection.authorized) {
         note = '(USB connect and run `tessel provision` to authorize)';
       }
-
```
johnnyman727: Can you remove these unrelated styling changes?

lib/controller.js
```diff
@@ -54,8 +59,10 @@ Tessel.list = function(opts) {
       }
       // If we have only one Tessel or two Tessels with the same name (just USB and LAN)
       else if (foundTessels.length === 1 ||
-        (foundTessels.length === 2 && foundTessels[0].name === foundTessels[1].name)) {
+        (foundTessels.length === 2 && foundTessels[0].name === foundTessels[1].name && foundTessels[0].connections[0].connectionType !== foundTessels[1].connections[0].connectionType)) {
         // Close all opened connections and resolve
+        console.log(foundTessels[0].connections.connectionType);
```
johnnyman727: @Student007 can you remove these `console.log`s?

lib/controller.js
johnnyman727: @Student007 do you need some assistance rebasing this commit on master? How can we take advantage of the work that was done on master while you were working on this? What of this new code can we remove?

johnnyman727: Ideally, we should remove the entire `seek` method because this functionality is in master.

lib/controller.js
```diff
@@ -170,6 +177,134 @@ Tessel.get = function(opts) {
...
+controller.root = function(opts) {
+  // ~ conversion to home because spawn isn't able to handle this right
+  if (opts.path && opts.path.substring(0, 1) === '~') {
+    var home = process.env.HOME;
+    opts.path = opts.path.replace('~', home);
+  }
+  var rtm;
+  if (!opts.menu) {
+    // TODO: Tessel cooperate identity conform menu required (color, ascii logo)
+    rtm = Menu({
```
johnnyman727: @Student007 we should pull the menu out of this PR, and place it into another PR where it can pop up a menu when *any* command is run and there is ambiguity about which T2 is the "right T2. In fact, it would fix another long standing issue, https://github.com/tessel/t2-cli/issues/235.

lib/root_controller.js
johnnyman727: Yes, I agree with @rwaldron. @Student007 any reason why this function wouldn't be added to the controller within `controller.js`?

lib/root_controller.js
```diff
@@ -0,0 +1,174 @@
...
+    if (tessel.connection.connectionType === 'LAN') {
+      if (!tessel.connection.authorized) {
+        rtm.add(index + ') ' + tessel.name + ': ' + tessel.connection.ip + ' (not authorized) \n');
+      } else {
+        rtm.add(index + ') ' + tessel.name + ': ' + tessel.connection.ip + ' (authorized) \n');
+      }
+
+    } else if (tessel.connection.connectionType === 'USB') {
+      if (!tessel.connection.authorized) {
+        rtm.add(index + ') ' + tessel.name + ': [USB] (not authorized) \n');
```
johnnyman727: All USB connections are authorized.

lib/root_controller.js
johnnyman727: +1

lib/root_controller.js
```diff
@@ -0,0 +1,174 @@
...
+      logs.info('Connect to ' + tessels[id].connections[0].host + '...');
+
+      // FIXME: There is no handler for poweroff Tessel ! Needs to be fixed within the firmware...
+      // (Terminal freezes imediatelly after poweroff the Tessel)
+
+      ch.once('error', function(e) {
+        if (e === 255) {
+          process.stdout.write('\u001b[2J\u001b[0;0H');
+          logs.warn('Sorry, you are not authorized!');
+          logs.info('"t2 key generate" might help :-)');
```
johnnyman727: We should encourage `t2 provision`. `t2 key generate` was admittedly a bit of a hack to get the VM usable.

lib/root_controller.js
johnnyman727: @rwaldron out of curiosity, what's the rationale behind that?

lib/usb_connection.js
johnnyman727: Agreed with Rick and they should be `logs.err` (because they are serious).

lib/usb_connection.js
```diff
@@ -228,10 +228,13 @@ util.inherits(USB.Scanner, EventEmitter);
 
 USB.Scanner.prototype.start = function() {
   var self = this;
+  if (haveusb) {
```
johnnyman727: We shouldn't have to need this check because any tests that might use T2 fetching should mock the discovery process.

lib/controller.js
Student007: @johnnyman727 yes I did many times (also at other code parts) but the linter automatically edited this many times. So I gave up to do this.

lib/controller.js
Student007: @johnnyman727 it was related to debug network topology behaviors. I used it for my own but decided to keep it in because some people do not know they will their Tessel twice due to network topology. A simple case I would expect lots of time is a iMac is connected by LAN and WLAN to the same router for different development reasons like multi IP services. Of cause people who do this, know it, but ... they will forgot it and need to be pointed on while contributing to Tessel... 
But of cause I can remove it - it is not well formed anyway.
Shall I remove it or change it ?

lib/controller.js
Student007: oh -- of cause ... don't know why I didn't see this

lib/controller.js
Student007: @rwaldron ... you are right. I didn't expect someone would use ~ in a different way ...

lib/root_controller.js
Student007: I learned to use module pattern where ever possible. Maybe it would be better to locate the file on a better place? I think the controller.js would grow much more over time - while developing and writing tests I noticed I am losing overview due to much additional things I put into code (what is away now - but it pointed me on to create a module). Of cause it isn't a reusable module but I also saw there are different parts in the t2-cli are done that way - so I thought I should do it the same way.
If no more comments on it, I will simple merge it back into controller.js

lib/root_controller.js
Student007: :bulb: ... ? If I steal the Professors Tessel 2 for installing a root kit .... I simple need to connect it to USB ? COOL :smiling_imp:

lib/root_controller.js
Student007: oh -- yes ... of cause !

lib/root_controller.js
Student007:
```
logs.warn('Sorry, you are not authorized!');
logs.info('Maybe this Tessel didn\'t got '"t2 provision" until now?');
```
Any better solution/wording ? - Also this is multiple times in the code - so I will put this into a const or method.

lib/root_controller.js
Student007: @rwaldron I need to add the reference to rtm so I replaced the former foreach by for in

lib/usb_connection.js
Student007: @johnnyman727 It causes from my work on reference machines. Is this behavior fixed by a newer part of code ? I belief there are many professional companies provide VMs for new employees. And if they did things wrong, they need that warning.

test/common/tessel-classic.js
Student007: ah .. ok ...

test/unit/menu.js
Student007: @rwaldron you are right, it could be an empty function.

test/unit/provision.js
Student007: @rwaldron ... it happens if I use `grunt test --force`

Student007: on Mac OS X Yosemite

test/unit/usb_connection.js
Student007: @rwaldron @johnnyman727 ... if they are necessary ... I think I didn't got the point yesterday ... for some reason I thought `logs.err` isn't legal ...

bin/tessel-2.js
Student007: The problem is, it would write `Usage: node t2 root` instead of `t2 root` - I like the idea of templates. 
@Frijol it looks like you are figuring out some things at https://github.com/tessel/t2-cli/pull/274 - is there also an general way to fix the `node t2 root` problem ?

test/unit/menu.js
Student007: as a different pull request what will fix all test files what are using this ?

lib/usb_connection.js
Student007: https://github.com/tessel/t2-cli/pull/293 fix this

test/unit/menu.js
Student007: @rwaldron is  https://github.com/tessel/t2-cli/pull/294 what you suggested?

lib/controller.js
Student007: @johnnyman727 I am currently working on this #235 and created a branch `menu` and `list` ... 
Looks like I have to make controller.ssh reusable (and give it a better name or divide it int o .ssh and .menu) :wink: but I will not get it done today.
If I understand it right, `Tessel.list` the Menu select event listener should set `process.env.TESSEL` to the selected tessel.name ? That way the next run would use it. Or maybe the provision needs to read the `process.env.TESSEL` somewhere additional ... 
(I am out of time for today)

bin/tessel-2.js
Frijol: If I have:

```js
parser.command('root')
  .option('path', {
    abbr: 'i',
    full: 'path',
    metavar: 'PATH',
    default: '~/.tessel/id_rsa',
    help: 'Private Key (Note: created by "t2 key generate")'
  })
  .callback(function(opts) {
    controller.root(opts)
  })
  .help('Gain SSH root access to one of your authorized tessels (menu listing if multiple targets)');
```

I get:

```sh
➜  t2-cli git:(t2-root) ✗ t2 root -h

Usage: /usr/local/bin/iojs t2 root [options]

Options:
   -i PATH, --path PATH   Private Key (Note: created by "t2 key generate")  [~/.tessel/id_rsa]

Gain SSH root access to one of your authorized tessels (menu listing if multiple targets)
```

What's the issue? You want to get rid of `/usr/local/bin/iojs` part?

Student007: @Frijol yes :smile:

Student007: if you will not find a solution, I think the best solution is to override parts of nomnom

Student007: what would be a new issue / contribution task

Frijol: Ah, I see. Yeah, that should be a separate issue. Good idea. A brief look through nomnom.js doesn't reveal a solution.

lib/controller.js
Student007: @johnnyman727 setting `process.env.TESSEL` isn't possible to be set for next call of `t2 list`. 
(maybe it would be by fs open and change profile on different operating systems ... but this would be an other task/issue).

So it will simple keep the menu is opened for setting the default Tessel to work with.

An other idea would be a `.tessel/defaultTessel.tmp` with the `tessel.name` in it ?!

The controller.js code doesn't show something about where and how you are using the selected Tessel. Heuristics only return one Tessel and logs.info is talking about... 
I need to know the plan about this :smile:

bin/tessel-2.js
Student007: (I couldn't find a provided solution in nomnom.js ... also reading code ... did this before some weeks)

lib/root_controller.js
Frijol: Isn't this already taken care of by controller.js?

```js
      // Add a note if the user isn't authorized to use it yet
      if (tessel.connection.connectionType === 'LAN' && !tessel.connection.authorized) {
        note = '(USB connect and run `tessel provision` to authorize)';
      }
```
Either way, the wording should match.

Student007: @Frijol The error 255 is responded by ssh when login failed. At this point this INFO will face up. Yours will came up when `t2 list` finding Tessels are not authorized yet. Finally it is only usability improvements.

lib/root_controller.js
johnnyman727: @Student007 You have a good point about making it modular, but I think we shouldn't have just one function be modular. We should either modularize the whole file or none of the file. `controller.js` is still pretty small so I think we're okay keeping it in that file for now :)

lib/controller.js
johnnyman727: @Student007 I think we should remove it and add infrastructure to `controller.list` to weed out duplicates.

lib/controller.js
johnnyman727: :+1:

johnnyman727: @Student007 did it only do this when you ran with `-force`?

lib/controller.js
johnnyman727: > If I understand it right, Tessel.list the Menu select event listener should set process.env.TESSEL to the selected tessel.name ?

A menu for selecting a Tessel should only be used with `Tessel.get`, not `Tessel.list`. Listing shows all available Tessels so there is no need to select one at that time. Let me know if that doesn't make sense. 

> An other idea would be a .tessel/defaultTessel.tmp with the tessel.name in it ?!

@Student007 I don't recommend storing state in a file across CLI calls because things can get hairy if you run two commands at once or another program is using the CLI API. In the latter case, you could have two processes writing to the file at the same time. 

When [logAndFinish](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L150-L169) get called after the search for a Tessel, if no `tessel` is passed in as the arg, it means multiple possible Tessels were found and this is where users should be able to select which they'd prefer. We should also print out a message that indicates that they can use the `--name` flag to avoid the manual selection process.

lib/root_controller.js
johnnyman727: ;)

lib/root_controller.js
johnnyman727: @Student007 if the code gets to this point, it should have already run through `Tessel.get` which weeds out Tessels that aren't authorized. This case should never be hit.

lib/usb_connection.js
johnnyman727: That's fair!

lib/root_controller.js
Student007: @Frijol hmm... the `t2 root` implementation currently seek for Tessels before attempting to login. While doing this it doesn't do heuristics and filtering of non authorized Tessels. The reason is, there are many cases the list of Tessels will variegate. Also there is not global List of Tessels I could attempt. One more reason is the use of option `t2 root -i ~\.ssh\id_rsa` what will fail on a not authorized Tessel. In that case the child_process will error with code 255. Another case would be, a user is accidental switched to superuser mode and uses `t2 root` what would use `/root/.tessel/id_rsa` or no key.

hmm ... maybe the wording should address this used case also (any suggestions)?

In general, the new branch for pull requesting (not published now) provides a method because this phrase is used often.
The call looks something like this:

```
controller.ssh.notAuthorized();
```
I have done this in the current working branch `menu` also. Here https://github.com/Student007/t2-cli/blob/menu/lib/controller.js#L405-L408 you can see the suggested divided objects `controller.ssh` and `controller.menu`

lib/root_controller.js
Student007: https://github.com/Student007/t2-cli/blob/menu/lib/controller.js#L321-L324
```
controller.menu.clear()
```
replaces this ...

lib/controller.js
Student007: OK like current idea of [logSelected](https://github.com/Student007/t2-cli/blob/menu/lib/controller.js#L82-L114) but moved to [logAndFinish else {...}](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L162-L168) (btw: the [new Menu](https://github.com/Student007/t2-cli/blob/menu/lib/controller.js#L92-L98) is moved to [prepareMenu](https://github.com/Student007/t2-cli/blob/menu/lib/controller.js#L247))

> @Student007 I don't recommend storing state in a file across CLI calls because things can get hairy if you run two commands at once or another program is using the CLI API. In the latter case, you could have two processes writing to the file at the same time. 

You are right. But in this special case it overrides the file when the user is doing interaction. Another process what overrides the file would fail the same way two terminals try to pipe data into one file or do other stuff with the same target.

I think it is the only useful way to fix a special Tessel to work with. I think there would be cases a developer will have e.g. 8 Tessels with different apps what should work together and he likes to select the current one without typing every time. He would use `t2 list` and select one. 

Another idea would be a new command `t2 select` what will show the menu to fix the Tessel to work with. Because it isn't possible to `export process.env.TESSEL` outside the running process, the only idea is to use a `~/.tesse./selectedTessel.log` (with the risk of other developers will hack around: My current non published version is checking whether the written name is matching an existing Tessel).

lib/controller.js
Student007: Yes I am currently working on that feature in [Tessel.list](https://github.com/Student007/t2-cli/blob/menu/lib/controller.js#L29-L53). At the moment I got crazy about sometimes there are objects with `tessel.connection` and sometimes `tessel.connections` my current workaround (not shown here and public) is causing secondary errors. I am still figuring out this and that line helped me. But of cause it will be away later - it has no usage for developing apps ...

lib/controller.js
Student007: yes, if the [logAndFinish](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L162-L168)  got a menu and **returns the selected Tessel (or set the process.env.TESSEL)** the `controller.ssh.seek(opts)` could be replaced by `Tessel.get(opts)` in [controller.root](https://github.com/Student007/t2-cli/blob/t2-root/lib/controller.js#L285-L305) - I have to take a look at `Tessel.get` I am currently not familiar how it works like (comming soon ;) ).

bin/tessel-2.js
Student007: @johnnyman727 it is:
- [Tessel.TESSEL_AUTH_PATH + '/id_rsa'](https://github.com/Student007/t2-cli/blob/master/lib/lan_connection.js#L32) -> used in lan_connection.js
- [path.join(os.tmpDir(),'id_rsa')](https://github.com/ericvicenti/ssh-keygen/blob/master/src/ssh-keygen.js#L98) -> extern ssh-keygen module
- [filepath](https://github.com/Student007/t2-cli/blob/master/lib/tessel/provision.js#L3-L13) -> provision.js

`Tessel` is included in `provision.js` and also `lan_connection.js` but useless in `bin/tessel-2.js`!
The only solution would be to use `controller.js` to provide that *const* because `controller.js` is included within `bin/tessel-2.js`. A possible solution could be:
```
controller.TESSEL_AUTH_KEY = Tessel.TESSEL_AUTH_PATH+'/id_rsa';
```
within [controller.js](https://github.com/Student007/t2-cli/blob/master/lib/controller.js#L10)

That way I would replace `default: default_id_rsa` by `controller.TESSEL_AUTH_KEY`

bin/tessel-2.js
johnnyman727: @Student007 what do you think about leaving the usage string our for this PR and then we'll make a separate PR that grabs all strings from a template file. I'd rather keep them all consistent at any one time so that it's more straightforward for new contributors.

bin/tessel-2.js
johnnyman727: @Student007 I think it makes the most sense for the path to live in `provision.js` since that file is responsible for creating the keys. Could you export the path from `provision.js` and import the path in `lan_connection.js` and `tessel-2.js`?

lib/controller.js
johnnyman727: @Student007 I don't think you need to worry about the number of connections for a Tessel. You can just check the `lanConnection` object: https://github.com/tessel/t2-cli/blob/master/lib/tessel/tessel.js#L13

That should give you the name and even ip address of the Tessel.

lib/controller.js
johnnyman727: `Tessel.get` essentially scans for all Tessels (just like `Tessel.list`) and then runs the found Tessels through a bunch of heuristics to determine which one the user likely wants. It can be considered a black box where you always get the Tessel the user would like to operate on.

lib/controller.js
johnnyman727: In the case where a user has several Tessels they are trying to work with, they could more easily set a different `TESSEL` env variable in a shell for each one. That way they don't have to manually interact - it would automatically select the proper Tessel. Once you open a separate PR for Tessel selection, we can continue the discussion there.

lib/root_controller.js
johnnyman727: @Student007 I'm sorry, I still don't understand. Why would a user not be authenticated at the point where they are trying to open a root shell on their T2? If you got this Tessel from `Tessel.get`, the Tessel is guaranteed to be authorized.

bin/tessel-2.js
Student007: > [since 6 days Nomnom is deprecated !](https://github.com/harthur/nomnom#deprecation-notice)
I've figured out ... and created a pull request:
https://github.com/tessel/t2-cli/pull/310
The reason I though it doesn't work was to use the "script" part on on the wrong object level.

bin/tessel-2.js
Student007: @johnnyman727 you mean changing `bin/tessel-2.js` by adding
```
default: require(../lib/tessel/provision.js).TESSEL_AUTH_KEY,
```
after changing provision.js like
```
module.exports.TESSEL_AUTH_KEY = filepath;
```

lib/controller.js
Student007: yes ... that was new ... ;)

lib/controller.js
Student007: also `grunt test` if I remember right. But last pull requests and after forking branches from tessel/master the problem doesn't come up anymore :+1:

lib/controller.js
Student007: @johnnyman727 I still didn't think about this from that point of view. You are right, if someone opens several terminals (one for each Tessel 2) the `~/.tesse./selectedTessel.log`idea isn't the right solution.
On the other hand a possible `t2 select` command will not work, because I don't know a way to set `process.env.TESSEL` out of Node.js/t2-cli to stay selected after running the command.
Maybe a possible solution could be to add the PID to the log-name and remove all log-names by t2-cli when the PID isn't alive anymore when running the t2-cli next time.
That way it would work to create a `t2 select` command with the goal to prevent the non-familiar way (to many non hackers).
If you would agree I will create that pull request.

lib/root_controller.js
Student007: @johnnyman727 if I add 't2 root -i my-secret-new-key-from-an-old-tessel-somewhere-out-of-the-cloud' the `Tessel.get` would still use the `~/.tessel/id_rsa` :key:

johnnyman727: Ah okay - so `t2 provision` really wouldn't help here. I'm not sure there is much advice we can give them besides "are you sure this public key has already been provided to Tessel?".

lib/controller.js
Student007: @johnnyman727 based on this ideas I am preparing an additional pull request, but I am not done at the moment.

What will be in:
- [x] makeCommand('select')
- [x] Tessel.seekTessels() -> returns list of found Tessels **asynchronously!**
- [x] controller.ssh.checkAuthorized() **asynchronously!**
- [x] testSSHConnection() **asynchronously!**
- [x] getAuthorizedTessels() with async.forEachOf(tessels, testSSHConnection)  **asynchronously!**
- [x] getTerminalsPID() **asynchronously!**
- [x] checkPIDexists(pid) process.platform based **asynchronously!**
- [x] checkAllPIDfiles() process.platform based **asynchronously!**
- [ ] createSelectedTesselPIDfile(pid) **asynchronously!**
- [x] checkAllPIDfiles() fs.readdirAsync & fs.readFileAsync **asynchronously!**
- [ ] set process.env.Tessel
- [ ] get global id_rsa
- [ ] let it work with menu pull request
```
var authorizedTessels = [];
    if (opts.name) {
      // if name selected test if authorized
      testSSHConnection()
      .getTerminalsPID()
      .checkAllPIDfiles()
      .createSelectedTesselPIDfile();
      //.catch()
    } else {
      getTerminalsPID()
      .checkAllPIDfiles()
      .getAuthorizedTessels(function(tessels) {
        if(tessels.length === 0) {
          // no authorized Tessels
          reject('No authorized Tessels found!');
        } else if (tessels.length === 1) {
          // select direct
          createSelectedTesselPIDfile();
        } else {
          // menu
          controller.menu.prepare(tessels)
          .createSelectedTesselPIDfile();
        }
      })
      .controller.menu.prepare()
      .createSelectedTesselPIDfile();
    }
```

johnnyman727:
@Student007 should we close this PR in favor of #307?

johnnyman727:
Closing in favor of reviewing #307 and various other PRs this was broken into.

PR #291, opened 2015-09-10
Displays a warning message that T2 might be booting if there are no USB comms

johnnyman727:


johnnyman727:
@kevinmehall after the timeout code paths runs the first time, the `usb` node module no longer loads. As soon as I unplug T2 in this stuck mode, the code starts running again.

Do I need to close the USB connection somehow? I attempted releasing the file descriptor but then the `usb` library yelled at me because the device has a pending request.

I'll need to follow up this PR with a `t2-firmware` PR to not store messages while OpenWRT is booting up because then the CLI doesn't know how to handle them after the command that executing that message finishes. (the process closes and there is no memory saved between them)

kevinmehall:
Not sure what you mean by "`usb` node module no longer loads". After the node process exits, a new process can't access the device? With what error? That would likely be if it's crashing or hanging the firmware. Or does it just break something within the process?

Yes, you can't cleanly close a device when transfers are pending on it. `node-usb` doesn't have a convenient API to cancel a transfer at the moment, but you could add one -- it already has C++ bindings for it, and uses it internally in `stopPoll`. Or set the [timeout](https://github.com/nonolith/node-usb#timeout-1) property on the endpoint before submitting the transfer and the transfer will fail when the timeout is hit.

johnnyman727:
@kevinmehall I was able to dig into this a little more. Here is how to recreate the issue consistently.

1. Plug in T2
2. Before OpenWRT boots up, run `t2 list`. It should print out a warning that T2 may not be booted within a few seconds.
3. Wait for OpenWRT to be finished booting by looking for the blinking green status LED on the ethernet port.
4. Run `t2 list` again and notice how it "hangs" on [`require('usb')`](https://github.com/tessel/t2-cli/blob/bd8285e087c715ae4838793922557b8869d53a83/lib/usb_connection.js#L11)
5. On my machine, it "hangs" for about 10 seconds before the `usb` module finishes loading, the CLI starts searching for Tessels, but does not discover any on the USB bus (it does not appear on `usb.getDeviceList()`.

johnnyman727:
@kevinmehall is there a way to know if the device timeout was hit? I couldn't find any evidence of this in the documentation. Also, the timeout is 1s by default and I was attempting to close the device after 2s so I'm not sure why there would still be a pending packet.

kevinmehall:
When the timeout is hit, the transfer will fail and call the callback with error code usb.LIBUSB_TRANSFER_TIMED_OUT. The timeout is unlimited by default.

I haven't seen `require('usb')` hang, and that means something "interesting" is going on with libusb or the OSX USB stack.

johnnyman727:
I'm finding that the timeout is set to `1000` by default. Or, at least, that's what I get when I print out the variable just [after the device is opened](https://github.com/tessel/t2-cli/blob/master/lib/usb_connection.js#L116).

It would be useful to know if this is just an OSX issue or a Linux issue. Let me know if you get a chance to repro it on your machine. 

Do you have any guesses on any other ways I can avoid that behavior?

johnnyman727:
@kevinmehall were you able to repro or do you have any suggestions for other techniques I can try?

Frijol:
related: https://github.com/tessel/t2-cli/issues/346

tcr:
Closing in favor of #386.

PR #293, opened 2015-09-10
Usb

Student007:
@rwaldron I also got the automatic changes in  test/unit/provision.js while running `grunt test` but I removed this changes to this pull request.

This tiny pull request will fix simple the usage of console.error to logs.err


rwaldron:
Thanks!

Issue #279, opened 2015-09-06
Provision attempts to write keys before creating them


johnnyman727:
```
~/self ❯❯❯ tessel provision                                                                                                        ⏎
INFO Connecting to Tessel...
INFO Connected to pixel over USB
INFO Creating public and private keys for Tessel authentication...
INFO Authenticating Tessel with public key...
INFO SSH Keys written.
```

Also, the keys were already written. This command should not have written over them.

johnnyman727:
As seen by @MrNice.

johnnyman727:
Here is how it should have looked (and it did after deleting `~/.tessel`)

```
~/self ❯❯❯ t2 provision
INFO Connecting to Tessel...
INFO Connected to pixel over USB
INFO Creating public and private keys for Tessel authentication...
INFO SSH Keys written.
INFO Authenticating Tessel with public key...
INFO Tessel authenticated with public key.
Add Comment Collapse
```

wprater:
fixed by #295

PR #295, opened 2015-09-10
fix issues with promise resolve in provisioning

### LINKED ISSUE #189: `t2 key generate` will overwrite existing SSH keys
The logic for checking if SSH keys already exists was moved out of the provisioning logic and into `controller.js` for the sake of speed.

Since `key generate` calls into the middle of the provisioning logic to just create a key (and not try to pass it to a Tessel 2), that check doesn't happen and new keys are made.

wprater:
fixes #189

### DIFF
--- a/lib/tessel/provision.js
+++ b/lib/tessel/provision.js
@@ -1,5 +1,6 @@
 //creates a .tessel folder with ssh keys in your home directory and uses those ssh keys to authorize you to push code to the USB-connected Tessel
 
+var util = require('util');
 var osenv = require('osenv');
 var fs = require('fs-extra');
 var path = require('path');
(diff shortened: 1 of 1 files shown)


wprater:
this should fix #279 as well?

johnnyman727:
@wprater this is great! I made a PR which should make our CI happy: https://github.com/wprater/t2-cli/pull/1

johnnyman727:
Bonus points if you can squash these commits into one for cleanliness.

wprater:
done!

johnnyman727:
Perfect! Feel free to merge when you're ready. Thanks for the contribution!

PR #296, opened 2015-09-10
print out the tessel version

### LINKED ISSUE #258: t2 version: print the version of openwrt/firmware running on T2

wprater:
fixes #258

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -277,6 +277,23 @@ parser.command('update')
   })
   .help('Update the Tessel firmware and openWRT image');
 
+parser.command('version')
+  .option('timeout', timeoutOption)
+  .option('name', nameOption)
+  .option('lan', {
+    flag: true,
+    help: 'Use LAN connection'
+  })
+  .option('usb', {
+    flag: true,
+    help: 'Use USB connection'
+  })
+  .callback(function(opts) {
+    controller.tesselFirmwareVerion(opts)
+      .then(closeSuccessfulCommand, closeFailedCommand);
+  })
+  .help('Display Tessel\'s current firmware version');
+
 
 module.exports = function(args) {
   parser.parse(args);
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -599,6 +599,27 @@ controller.updateTesselWithVersion = function(force, tessel, currentVersion, bui
     });
 };
 
+controller.tesselFirmwareVerion = function(opts) {
+  return Tessel.get(opts)
+    .then(function(tessel) {
+      return tessel.fetchCurrentBuildInfo()
+        .then(function(versionSha) {
+          return updates.requestBuildList().then(function(builds) {
+            // Figure out what commit SHA is running on it
+            var version = updates.findBuild(builds, 'sha', versionSha).version;
+            logs.info('Tessel [' + tessel.name + '] version: ' + version);
+
+            return controller.closeTesselConnections(tessel);
+          });
+        })
+        .catch(function() {
+          logs.info('Tessel [' + tessel.name + '] version: unknown');
+
+          return controller.closeTesselConnections(tessel);
+        });
+    });
+};
+
 module.exports = controller;
 
 // Shared exports


wprater:
might be nice to have `fetchCurrentBuildInfo` return the same object that builds.json stores.  this way we don't need an extra lookup when fetching version info?  this would require changing what's saved in `/etc/tessel-version`. thoughts?

bin/tessel-2.js
```diff
@@ -277,6 +277,15 @@ parser.command('update')
...
   .help('Update the Tessel firmware and openWRT image');
 
+parser.command('version')
+  .option('timeout', timeoutOption)
+  .option('name', nameOption)
+  .callback(function(opts) {
+    controller.tesselFirmwareVerion(opts)
+      .then(closeSuccessfulCommand, closeFailedCommand);
+  })
+  .help('Tessel firmware version');
```
johnnyman727: I recommend this text gets changed to "Display Tessel's current firmware version" to remain consistent with the other help commands which are generally more verbose.

lib/controller.js
```diff
@@ -599,6 +599,27 @@ controller.updateTesselWithVersion = function(force, tessel, currentVersion, bui
     });
 };
 
+controller.tesselFirmwareVerion = function(opts) {
+  return Tessel.get(opts)
+    .then(function(tessel) {
+      return tessel.fetchCurrentBuildInfo()
+        .then(function(versionSha) {
+          updates.requestBuildList().then(function(builds) {
```
johnnyman727: You'll need to `return` this value otherwise the process will close before this Promise is resolved (at least it does for me):

```.js
return updates.requestBuildList().then(function(builds) {
...
});
```

lib/controller.js
```diff
@@ -599,6 +599,27 @@ controller.updateTesselWithVersion = function(force, tessel, currentVersion, bui
...
+            // Figure out what commit SHA is running on it
+            var version = updates.findBuild(builds, 'sha', versionSha).version;
+            logs.info('Tessel version: ' + version);
```
johnnyman727: I suggest we display the name of the T2 whose version we're displaying. This is helpful for folks who might have several in a room and need to know if they accidentally fetched the version of the wrong one. You can access the name with `tessel.name`.

lib/controller.js
```diff
@@ -599,6 +599,27 @@ controller.updateTesselWithVersion = function(force, tessel, currentVersion, bui
...
+
+            return controller.closeTesselConnections(tessel);
+          });
+        })
+        .catch(function(err) {
+          logs.info('Tessel version: unknown');
```
johnnyman727: Same comment as above.

bin/tessel-2.js
```diff
@@ -277,6 +277,15 @@ parser.command('update')
   })
   .help('Update the Tessel firmware and openWRT image');
 
+parser.command('version')
+  .option('timeout', timeoutOption)
+  .option('name', nameOption)
```
johnnyman727: This command should also have the [`usb` and `lan` options](https://github.com/tessel/t2-cli/blob/master/bin/tessel-2.js#L74-L81). I wonder if there is a way to make the `timeout`, `name`, `lan`, and `usb` options common to all commands?

johnnyman727:
@wprater Thanks for this PR! I made a couple of suggestions but otherwise it works great.

Your suggestion about changing the contents of the version file on T2 seems reasonable to me. I ended up putting [just the commit SHA](https://github.com/tessel/openwrt-tessel/commit/bf327359b4a13b4da07bc5776fe8a22ae88d54f9#diff-b67911656ef5d18c4ae36cb6741b7965R44) because that's about the limits of my knowledge of build systems :) 

The change your suggesting would go great with the release script I've never had the chance to write which would make it one-step to publish a new build to the server.

johnnyman727:
Lastly, make sure you run `grunt` before pushing - it runs our linter and unit tests. Looks like you just need to add a semi-colon.

lib/controller.js
wprater: of course!

lib/controller.js
wprater: good idea!

bin/tessel-2.js
wprater: sure.  I made a little factory function, that we could use.

```
function tesselCommandFactory(commandName) {
  return parser.command(commandName)
    .option('timeout', timeoutOption)
    .option('name', nameOption)
    .option('lan', {
      flag: true,
      help: 'Use LAN connection'
    })
    .option('usb', {
      flag: true,
      help: 'Use USB connection'
    });
}

tesselCommandFactory('version')
  .callback(function(opts) {
    controller.tesselFirmwareVerion(opts)
  })
  .help('Display Tessel\'s current firmware version');
```

can make another PR for that if you like?

wprater:
thanks for the grunt tip.  I'll setup my linter to work with your rule files only now, but will also check the grunt output.

wprater:
want me to rebase these into 1 commit now?

johnnyman727:
Yep!

wprater:
cool, done.

bin/tessel-2.js
johnnyman727: That would be great! It would make this file so much more readable.

johnnyman727:
Looks perfect. Merge when you're ready.

PR #300, opened 2015-09-14
Forces an update if no version file found unless specifically requested not to do that

johnnyman727:


Frijol:
works great, merge away!

PR #301, opened 2015-09-14
Move off legacy Travis infra

johnnyman727:


johnnyman727:
Welp, it looks like we can't move off legacy infrastructure because we need to use `sudo` to install usb deps.

PR #303, opened 2015-09-14
friendlier blinky

Frijol:
Current blinky from `t2 init` flashes red and yellow in sync - not only overwriting error and wifi LEDs but also perfectly mimicking crash behavior from T1.

This blinks alternating green and yellow just like T1 blinky.


johnnyman727:
Looks good to me! Once Travis gives you the green, feel free to merge.

PR #304, opened 2015-09-14
[WIP] Adds a deploy progress bar

johnnyman727:
This progress bar is pointless at the moment because it hits 100% immediately. We need some way to get more information on what's going on with the internal backpressure...


johnnyman727:
Closing this because we'll need a different way to track backpressure relief through the stream.

PR #305, opened 2015-09-14
add factory method to make a tessel command

wprater:
- adds default options for name, usb, lan, and timeout

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -7,17 +7,6 @@ var key = require('../lib/key');
 var init = require('../lib/init');
 var logs = require('../lib/logs');
 
-var nameOption = {
-  metavar: 'NAME',
-  help: 'The name of the tessel on which the command will be executed'
-};
-
-var timeoutOption = {
-  abbr: 't',
-  metavar: 'TIMEOUT',
-  help: 'Set timeout in seconds for scanning for networked tessels',
-  default: 5
-};
 
 function closeSuccessfulCommand() {
   process.exit(0);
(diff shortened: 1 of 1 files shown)


wprater:
@johnnyman727 some methods did not have all the `name`, `usb`, `lan`, and `timeout` methods.  for example `erase`.  with the new `makeCommand` method, they will.  is this what you intended?

lib/tessel/deploy.js
```diff
@@ -29,18 +29,7 @@ Tessel.prototype.deployScript = function(opts) {
 
     // Stop running any existing scripts
     return actions.execRemoteCommand(self, 'stopRunningScript', filepath)
-      .then(function() {
```
johnnyman727: @wprater this seems unrelated to this PR. Does it need to be in here?

bin/tessel-2.js
```diff
@@ -38,6 +38,20 @@ function closeFailedCommand(err) {
   process.exit(1);
 }
 
+function makeCommand(commandName) {
+  return parser.command(commandName)
+    .option('timeout', timeoutOption)
+    .option('name', nameOption)
+    .option('lan', {
```
johnnyman727: I think we should either break the `lan` and `usb` option into separate variables just like `timeoutOption` and `nameOption` are or bring the `timeoutOption` and `nameOption` back into this function (for consistency). I don't think those two vars are used anywhere else.

johnnyman727:
@wprater just two comments but this is a great refactor.

lib/tessel/deploy.js
wprater: whoops.. it was from another PR.. will rebase out.

wprater:
good to merge when you are, @johnnyman727

johnnyman727:
Love it!

wprater:
need a quick hotfix!

wprater:
nevermind! I tricked myself while I was on another branch (:

PR #307, opened 2015-09-14
t2 root

Student007:
@johnnyman727 please take a look at both commit comments and additional requests on slack :smile:

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -7,6 +7,7 @@ var key = require('../lib/key');
 var init = require('../lib/init');
 var logs = require('../lib/logs');
 
+
 var nameOption = {
   metavar: 'NAME',
   help: 'The name of the tessel on which the command will be executed'
@@ -174,6 +175,34 @@ parser.command('list')
   })
   .option('timeout', timeoutOption)
   .help('Lists all connected Tessels and their authorization status.');
+// accessing the root shell of your tessels
+// Fixes issue https://github.com/tessel/t2-cli/issues/80
+/**
+$ t2 root --help
+> Usage: tessel root [-i <path>] [--help]
+>  -i <path>:   provide a path to the desired ssh key
+$ 
+
+$ t2 root
+> Accessing root...
+root@192.168.128.124 # 
+*/
+var functional_msg = '\nGain SSH root access to one of your authorized tessels (menu listing if multiple targets)';
+parser.command('root')
+  .usage(functional_msg + '\n\nUsage: t2 root [-i PATH] [--help]\n\n-i PATH:   Optional targeting a different Private Key \n\n(Note: default target created by "t2 key generate" is ' + controller.TESSEL_AUTH_KEY + ')\n')
+  .option('path', {
+    abbr: 'i',
+    full: 'path',
+    metavar: 'PATH',
+    default: controller.TESSEL_AUTH_KEY,
+    help: 'Private Key (Note: created by "t2 key generate")'
+  })
+  .callback(function(opts) {
+    controller.root(opts)
+      .then(closeSuccessfulCommand, closeFailedCommand);
+  })
+  .help(functional_msg);
+
 
 parser.command('init')
   .callback(init)
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -124,18 +124,18 @@ Tessel.get = function(opts) {
 
             // Run the heuristics to pick which Tessel to use
             return controller.runHeuristics(opts, tessels)
-              .then(function(tessel){
-                if(!tessel){
-                  controller.menu.prepareMenu(opts, tessels, resolve, reject,'Please select the default Tessel 2', function(name,index,resolve,reject){
-                    if(!name && name === 'EXIT'){
+              .then(function(tessel) {
+                if (!tessel && opts.root) {
+                  controller.menu.prepareMenu(opts, tessels, resolve, reject, 'Please select the default Tessel 2', function(name, index, resolve, reject) {
+                    if (!name && name === 'EXIT') {
                       reject('No Tessel selected!');
                     }
-                    logs.info('Selected Tessel: '+name);
-                     logAndFinish(tessels[index]);
+                    logs.info('Selected Tessel: ' + name);
+                    logAndFinish(tessels[index]);
                   });
                 } else {
-                    logAndFinish(tessel);
-                }  
+                  logAndFinish(tessel);
+                }
               });
           });
       }
@@ -167,8 +167,8 @@ Tessel.get = function(opts) {
       // The Tessels that we won't be using should have their connections closed
       var connectionsToClose = tessels;
       if (tessel) {
-        if(opts.root){
-          
+        if (opts.root) {
+
           controller.closeTesselConnections(connectionsToClose)
             .then(function() {
               controller.ssh.runSSH(0, opts, tessels, resolve, reject);
@@ -203,7 +203,6 @@ may or may not be open and closes them
 controller.closeTesselConnections = function(tessels) {
   return new Promise(function(resolve, reject) {
     async.each(tessels, function closeThem(tessel, done) {
-      console.log('error: ',tessel);
         // If not an unauthorized LAN Tessel, it's connected
         if (!(tessel.lanConnection &&
             !tessel.lanConnection.authorized)) {
@@ -668,51 +667,51 @@ controller.menu = {
 
   prepareMenu: function(opts, tessels, resolve, reject, title, callback) {
     //return new Promise(function(resolve, reject) {
-      
-      var rtm = new Menu({
-        //  width: process.stdout.columns - 4,
-        width: 55,
-        x: 1,
-        y: 2,
-        bg: 'red'
-      });
-      rtm.reset();
-      rtm.write(title + '\n');
-      rtm.write('                                                \n');
 
-      // create menu entries
-      for (var i in tessels) {
-        controller.menu.makeMenu(tessels[i], i, rtm);
-      }
-      
-      rtm.add('EXIT\n');
-      rtm._draw();
-      if (!opts.menu) {
-        // if this is no test, starting the menu as child process
-        controller.menu.showMenu(rtm, resolve, reject);
+    var rtm = new Menu({
+      //  width: process.stdout.columns - 4,
+      width: 55,
+      x: 1,
+      y: 2,
+      bg: 'red'
+    });
+    rtm.reset();
+    rtm.write(title + '\n');
+    rtm.write('                                                \n');
+
+    // create menu entries
+    for (var i in tessels) {
+      controller.menu.makeMenu(tessels[i], i, rtm);
+    }
+
+    rtm.add('EXIT\n');
+    rtm._draw();
+    if (!opts.menu) {
+      // if this is no test, starting the menu as child process
+      controller.menu.showMenu(rtm, resolve, reject);
+    } else {
+      // because the test needs to resolve the promise ...
+      resolve({
+        opts: opts,
+        tessels: tessels
+      });
+    }
+
+    rtm.once('select', function(label, index) {
+      rtm.close();
+      controller.menu.clear();
+
+      // Identify the Exit command by first letter (all other entries start with numbers/index)
+      if (label[0] !== 'E') {
+        callback(tessels[index].name, index, resolve, reject);
+        // controller.ssh.runSSH(index, opts, tessels, resolve, reject);
       } else {
-        // because the test needs to resolve the promise ...
-        resolve({
-          opts: opts,
-          tessels: tessels
-        });
-      }
-
-      rtm.once('select', function(label, index) {
-        rtm.close();
+        // going to clear screen and calling exit to resolve promise
         controller.menu.clear();
-
-        // Identify the Exit command by first letter (all other entries start with numbers/index)
-        if (label[0] !== 'E') {
-          callback(tessels[index].name, index, resolve, reject);
-          // controller.ssh.runSSH(index, opts, tessels, resolve, reject);
-        } else {
-          // going to clear screen and calling exit to resolve promise
-          controller.menu.clear();
-          controller.menu.exit(resolve);
-        }
-      });
-  //  });
+        controller.menu.exit(resolve);
+      }
+    });
+    //  });
   },
   makeMenu: function(tessel, index, rtm) {
     if (tessel.lanConnection) {
@@ -728,7 +727,7 @@ controller.menu = {
       } else {
         rtm.add(index + ') ' + tessel.name + ': [USB] (authorized) \n');
       }
-    } 
+    }
     // FIXME: Workaround for trouble with dynamic added menu elements
     rtm._draw();
   },
@@ -867,8 +866,8 @@ controller.root = function(opts) {
   opts.root = true;
   return new Promise(function(resolve, reject) {
     controller.menu.seek(opts)
-    .then(resolve)
-    .catch(reject);
+      .then(resolve)
+      .catch(reject);
   });
 };
 
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -7,7 +7,10 @@ var sprintf = require('sprintf-js').sprintf;
 var cp = require('child_process');
 var async = require('async');
 var updates = require('./update-fetch');
+var debug = require('debug')('controller');
+var Menu = require('terminal-menu');
 var controller = {};
+controller.TESSEL_AUTH_KEY = Tessel.TESSEL_AUTH_PATH + '/id_rsa';
 
 Tessel.list = function(opts) {
 
@@ -92,7 +95,7 @@ Tessel.list = function(opts) {
 
 Tessel.get = function(opts) {
   return new Promise(function(resolve, reject) {
-    logs.info('Looking for your Tessel...');
+    logs.info('Searching nearby LAN (netmask) and USB for Tessels...');
     // Store the amount of time to look for Tessel in seconds
     var timeout = (opts.timeout || 2) * 1000;
     // Collection variable as more Tessels are found
@@ -118,9 +121,22 @@ Tessel.get = function(opts) {
         controller.reconcileTessels(tessels)
           .then(function(reconciledTessels) {
             tessels = reconciledTessels;
+
             // Run the heuristics to pick which Tessel to use
             return controller.runHeuristics(opts, tessels)
-              .then(logAndFinish);
+              .then(function(tessel){
+                if(!tessel){
+                  controller.menu.prepareMenu(opts, tessels, resolve, reject,'Please select the default Tessel 2', function(name,index,resolve,reject){
+                    if(!name && name === 'EXIT'){
+                      reject('No Tessel selected!');
+                    }
+                    logs.info('Selected Tessel: '+name);
+                     logAndFinish(tessels[index]);
+                  });
+                } else {
+                    logAndFinish(tessel);
+                }  
+              });
           });
       }
     }
@@ -150,15 +166,22 @@ Tessel.get = function(opts) {
     function logAndFinish(tessel) {
       // The Tessels that we won't be using should have their connections closed
       var connectionsToClose = tessels;
-
       if (tessel) {
-        logs.info(sprintf('Connected to %s over %s', tessel.name, tessel.connection.connectionType));
-        connectionsToClose.splice(tessels.indexOf(tessel), 1);
+        if(opts.root){
+          
+          controller.closeTesselConnections(connectionsToClose)
+            .then(function() {
+              controller.ssh.runSSH(0, opts, tessels, resolve, reject);
+            });
+        } else {
+          logs.info(sprintf('Connected to %s over %s', tessel.name, tessel.connection.connectionType));
+          connectionsToClose.splice(tessels.indexOf(tessel), 1);
 
-        controller.closeTesselConnections(connectionsToClose)
-          .then(function() {
-            return resolve(tessel);
-          });
+          controller.closeTesselConnections(connectionsToClose)
+            .then(function() {
+              return resolve(tessel);
+            });
+        }
       } else {
         logs.info('Please specify a Tessel by name [--name <tessel name>]');
         controller.closeTesselConnections(connectionsToClose)
@@ -180,9 +203,10 @@ may or may not be open and closes them
 controller.closeTesselConnections = function(tessels) {
   return new Promise(function(resolve, reject) {
     async.each(tessels, function closeThem(tessel, done) {
+      console.log('error: ',tessel);
         // If not an unauthorized LAN Tessel, it's connected
-        if (!(tessel.connection.connectionType === 'LAN' &&
-            !tessel.connection.authorized)) {
+        if (!(tessel.lanConnection &&
+            !tessel.lanConnection.authorized)) {
           // Close the connection
           return tessel.close()
             .then(done, done);
@@ -333,7 +357,6 @@ controller.runHeuristics = function(opts, tessels) {
           }
         }
       }
-
       // At this point, we know that no name option or env variable was set
       // and we know that there is only one USB and/or on LAN Tessel
       // We'll return the highest priority available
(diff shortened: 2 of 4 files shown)


wprater:
thanks for your submission.  I have not followed all of this, but could you add a brief description in the PR so anyone can see quickly grok the context of this PR (help for historical reasons as well)?  might I also suggest that you remove all code that's [been](https://github.com/tessel/t2-cli/pull/307/files#diff-f68567477d803b49930337bf7fe1556bR707)  [commented](https://github.com/tessel/t2-cli/pull/307/files#diff-f68567477d803b49930337bf7fe1556bR672 ) [out](https://github.com/tessel/t2-cli/pull/307/files#diff-f68567477d803b49930337bf7fe1556bR669)?  it's confusing for a code reviewer and should not be in the deployed codebase.

bin/tessel-2.js
```diff
@@ -7,6 +7,7 @@ var key = require('../lib/key');
 var init = require('../lib/init');
 var logs = require('../lib/logs');
 
+
```
johnnyman727: This is super nit-picky but before requesting reviews on PRs, can you remove whitespace changes?

lib/controller.js
```diff
@@ -92,7 +95,7 @@ Tessel.list = function(opts) {
 
 Tessel.get = function(opts) {
   return new Promise(function(resolve, reject) {
-    logs.info('Looking for your Tessel...');
+    logs.info('Searching nearby LAN (netmask) and USB for Tessels...');
```
johnnyman727: This message is confusing - can you change it back to the original?

lib/controller.js
```diff
@@ -7,7 +7,10 @@ var sprintf = require('sprintf-js').sprintf;
 var cp = require('child_process');
 var async = require('async');
 var updates = require('./update-fetch');
+var debug = require('debug')('controller');
+var Menu = require('terminal-menu');
 var controller = {};
+controller.TESSEL_AUTH_KEY = Tessel.TESSEL_AUTH_PATH + '/id_rsa';
```
johnnyman727: I commented on the other PR, but I think this variable should actually be defined and exported in `provision.js` since that file is responsible for writing the keys. What do you think?

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+    if (tessel.lanConnection) {
+      if (!tessel.lanConnection.authorized) {
+        rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (not authorized) \n');
+      } else {
+        rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (authorized) \n');
+      }
+
+    } else if (tessel.usbConnection) {
+      if (!tessel.usbConnection.authorized) {
+        rtm.add(index + ') ' + tessel.name + ': [USB] (not authorized) \n');
```
johnnyman727: USB Connections are always authorized

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+  a ncurses like menu you can select the Tessel you like to gain root access...
+
+  The structure of code and maybe unusual parts are paid due being testable.
+  The lightweight terminal-menu isn't written testable what needs a little bit hacking.
+
+  Finally some parts are causing from my personnal learning curve about writing unit tests using sinon!
+*/
+controller.root = function(opts) {
+  // ~ conversion to home because spawn isn't able to handle this right
+  if (opts.path && opts.path.substring(0, 1) === '~') {
```
johnnyman727: Could you just use the [`osenv`](https://github.com/tessel/t2-cli/blob/master/lib/tessel/provision.js#L12) module here?

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+    var home = process.env[(process.platform === 'win32') ? 'USERPROFILE' : 'HOME'];
+    if (/^~/.test(opts.path)) {
+      opts.path = opts.path.replace(/~/, home);
+    }
+  }
+  opts.root = true;
+  return new Promise(function(resolve, reject) {
+    controller.menu.seek(opts)
```
johnnyman727: This should be something like:
```.js
// Grab the preferred Tessel
  return Tessel.get(opts)
    .then(function(tessel) {
      // Open up an SSH session
      return tessel.beginSSHSession(opts)
        .then(function() {
          return controller.closeTesselConnections(tessel);
        });
    });
```

As far as I can tell, `controller.menu.seek` is the same as `Tessel.get` so we might as well make it consistent with all the other commands.

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
         });
     });
 };
+controller.menu = {
```
johnnyman727: Let's remove all `menu`-related code and put it in it's own PR. It's not specifically related to SSH sessions.

lib/controller.js
```diff
@@ -181,8 +204,8 @@ controller.closeTesselConnections = function(tessels) {
   return new Promise(function(resolve, reject) {
     async.each(tessels, function closeThem(tessel, done) {
         // If not an unauthorized LAN Tessel, it's connected
-        if (!(tessel.connection.connectionType === 'LAN' &&
-            !tessel.connection.authorized)) {
+        if (!(tessel.lanConnection &&
```
johnnyman727: Good catch!

lib/controller.js
```diff
@@ -150,15 +166,22 @@ Tessel.get = function(opts) {
...
-        connectionsToClose.splice(tessels.indexOf(tessel), 1);
+        if (opts.root) {
 
-        controller.closeTesselConnections(connectionsToClose)
-          .then(function() {
-            return resolve(tessel);
-          });
+          controller.closeTesselConnections(connectionsToClose)
+            .then(function() {
+              controller.ssh.runSSH(0, opts, tessels, resolve, reject);
```
johnnyman727: @Student007, you shouldn't need to modify the `Tessel.get` function. It should be a black box that returns Tessels to functions that ask for it and shouldn't have any references to any other specific Tessel commands. You should call the `ssh` function from within `controller.root` after requesting the Tessel from `Tessel.get`.

lib/controller.js
johnnyman727: Additionally, if you need to make sure that you get a `lan` Tessel, you can set the `--lan` flag to true.

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+  },
+  notAuthorized: function() {
+    logs.warn('Sorry, you are not authorized!');
+    logs.info('Maybe this Tessel didn\'t got "t2 provision" until now?');
+  }
+};
+
+/*
+  The T2 root command is used to login into the firmware and gaining superuser access.
+  The security is provided by RSA - your Tessel get the keys while provisioning via USB.
```
johnnyman727: typo - 'your Tessel get the keys'. I suggest you make it 'your Tessel receives the keys over USB when provisioned'.

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+  
+  If you have only one Tessel, the T2 root command will login directly else there is
```
johnnyman727: I suggest you remove this. Being able to use the menu should be something you can do for any Tessel command.

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+};
+
+/*
+  The T2 root command is used to login into the firmware and gaining superuser access.
+  The security is provided by RSA - your Tessel get the keys while provisioning via USB.
+  
+  If you have only one Tessel, the T2 root command will login directly else there is 
+  a ncurses like menu you can select the Tessel you like to gain root access...
+
+  The structure of code and maybe unusual parts are paid due being testable.
```
johnnyman727: Remove this - we'll make it work but it shouldn't be on the code in production.

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+  The lightweight terminal-menu isn't written testable what needs a little bit hacking.
+
+  Finally some parts are causing from my personnal learning curve about writing unit tests using sinon!
```
johnnyman727: Remove this as well.

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+  },
+  exit: function(resolve) {
+    resolve();
+    //process.exit();
+  }
+};
+controller.ssh = {
+  runSSH: function(id, opts, tessels, resolve, reject) {
+    // clear console
+    debug('runSSH...\n root@' + tessels[id].lanConnection.host);
```
johnnyman727: This throws an error if a Tessel with only a USB connection is provided. You need to check for that and reject with an error if it's not possible.

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+      var ch = require('child_process')
+        .spawn('ssh', ['-i',
+          opts.path,
+          'root@' + tessels[id].lanConnection.ip
+        ], {
+          stdio: 'inherit'
+        });
+      logs.info('Connect to ' + tessels[id].lanConnection.host + '...');
+
+      // FIXME: There is no handler for poweroff Tessel ! Needs to be fixed within the firmware...
```
johnnyman727: What exactly are you looking for here? A notification that Tessel is turning off?

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+  exit: function(resolve) {
+    resolve();
+    //process.exit();
+  }
+};
+controller.ssh = {
+  runSSH: function(id, opts, tessels, resolve, reject) {
+    // clear console
+    debug('runSSH...\n root@' + tessels[id].lanConnection.host);
+    if (opts.menu) {
```
johnnyman727: What is the menu option for?

lib/controller.js
```diff
@@ -641,6 +663,213 @@ controller.tesselFirmwareVerion = function(opts) {
...
+        });
+      logs.info('Connect to ' + tessels[id].lanConnection.host + '...');
+
+      // FIXME: There is no handler for poweroff Tessel ! Needs to be fixed within the firmware...
+      // (Terminal freezes imediatelly after poweroff the Tessel)
+
+      ch.once('error', function(e) {
+        if (e === 255) {
+          controller.menu.clear();
+          controller.ssh.notAuthorized();
```
johnnyman727: How would it get to this point if it's not authorized?

johnnyman727:
In addition to @wprater's valid comments, I've added some suggestions to the code as well. Thanks for contributing @Student007!

bin/tessel-2.js
wprater: I think you can write a rule for this in jscs.  so it will fail travis builds.. or are you only using jshint

Student007:
@wprater sorry - I was in a harry and fretful of someone will change things while I am preparing my pull request. As you can see ... there are conflicts wasn't there yesterday.

wprater:
you can rebase from master and fix them.  they are probably from the
changes I made today.  should not be so tough to fix?

lib/controller.js
Student007: @johnnyman727 with the change of possible multiple `Looking for your Tessels ...` but being earnest: I think it is more useful to get an info about where and what is searched... 
The old version suggests the CLI is created for a single Tessel - and looking to the code I feel like this had bin the initial purpose ?!

lib/controller.js
Student007: - [Tessel.TESSEL_AUTH_PATH + '/id_rsa'](https://github.com/Student007/t2-cli/blob/master/lib/lan_connection.js#L32) -> used in lan_connection.js
- [path.join(os.tmpDir(),'id_rsa')](https://github.com/ericvicenti/ssh-keygen/blob/master/src/ssh-keygen.js#L98) -> extern ssh-keygen module
- [filepath](https://github.com/Student007/t2-cli/blob/master/lib/tessel/provision.js#L3-L13) -> provision.js

`Tessel` is included in `provision.js` and also `lan_connection.js` but useless in `bin/tessel-2.js`!
The only solution would be to use `controller.js` to provide that *const* because `controller.js` is included within `bin/tessel-2.js`. A possible solution could be:
```
controller.TESSEL_AUTH_KEY = Tessel.TESSEL_AUTH_PATH+'/id_rsa';
```
within [controller.js](https://github.com/Student007/t2-cli/blob/master/lib/controller.js#L10)

For this reason I decided to create `controller.TESSEL_AUTH_KEY`

Student007: it is used in `bin/tessel-2.js` to replace `default: default_id_rsa` due to your request for reusable central.

lib/controller.js
Student007: @johnnyman727 - no. If I use a Tessel is not provisioned at this computer, I may not login. If this isn't the case, it is a big security problem and requires to open a major issue bound to milestone

lib/controller.js
Student007: @johnnyman727 you could add this comment also [in answer to @rwaldron wishes](https://github.com/tessel/t2-cli/pull/290/files#r39067831).
How do you think the replacement should look like ?

lib/controller.js
Student007: @johnnyman727 as `controller.menu.seek` calls `Tessel.get` - the reason is `unit testing ...` 
(but of cause I can kick the tests and do understandable code - it causes from the internals of `terminal-menu` code desing)

About the `--lan` flag ? What do you mean ? The `t2 root` command doesn't know that.

lib/controller.js
Student007: I do not understand the security concept ? Could you explain please ?

lib/controller.js
Student007: https://github.com/tessel/t2-cli/pull/309 ok

lib/controller.js
Student007: ok - but why not `Tessels receive their keys over USB when provisioned` - I am in conflict with thinking about only one Tessel.

lib/controller.js
Student007: you mean `If you have only one Tessel, the T2 root command will login directly ` ?

lib/controller.js
Student007: see

lib/controller.js
Student007: yes .. nobody should know or remember it was hard to dive into Tessel development ;)

lib/controller.js
Student007: yes, yes ... good to know people own a Tessel 2 .... ... ...

lib/controller.js
Student007: I think it is not possible to fix this problem. The terminal freeze if you power down the Tessel 2 virtual machine. If you do `poweroff` logged in to a Tessel 2 by `t2 root` (ssh) this behavior is bad ... but there is no easy solution. I would expect it is fixing poweroff command by looking for open terminals and exit them with error code before power down. The comment is due to opening an issue or make a central note about knowledge about this behavior.

lib/controller.js
Student007: it is a testing switch

lib/controller.js
Student007: the `t2 root` command isn't interested in whether a Tessel 2 is authorized. It would be able to use a different `id_rsa` by `t2 root -i http://local/key_from_cloud` (e.g.) and login to a Tessel. In the case the key is wrong ... this error code is responded. 

At this point: Once more ... the solution with a flag "authorized" isn't helpful for creating a menu or do other things like replacing id_rsa ...

I think at the moment I need to create
```
controller.ssh = {
// ....
  checkAuthorized = function(ip,id_rsa){
   var ch = require('child_process')
        .spawn('ssh', ['-i',
         id_rsa,
          'root@' + ip
        ], {
          stdio: 'inherit'
        });
    ch.once('error', function(e) {
        if (e === 255) {
          return false;
       }
      });
    ch.once('exit', function(e) {
     if (e === 255) {
      return false;   
      } else if(e === 0){
      return true;
    }
 });
ch.kill('SIGHUP');
  }
}

```
And call it within 
```
controller.menu = {
// ...
 makeMenu: function(tessel, index, rtm) {
    if (tessel.lanConnection) {
var auth = controller.ssh.checkAuthorized(tessel.lanConnection.ip,opts.path)
// if (!tessel.lanConnection.authorized) {
if (!auth) {
rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (not authorized) \n');
      } else {
        rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (authorized) \n');
      }
// ...
```

johnnyman727: Oh I understand - so this is if a user provides a key file but once the SSH process runs, T2 decides the keys aren't legit?

Student007: If I use the `t2-cli`, I will do this multiple times - each time for a single job. And if I (would... ...) own multiple Tessels, it is possible I like to have multiple keys (one for each Tessel). So if I run a command like `t2 root` or `t2 list`, I will use `-i key_of_tesselXYZ`or `-i key_of_Customer123s_Tessels` ...(e.g.). Based on this, the design of being authorized should be reviewed. 

I think I didn't got it all together in my mind (especial about USB is never `not authorized`) - but I feel something isn't perfect at this point.

bin/tessel-2.js
rwaldron: https://github.com/tessel/t2-cli/blob/master/.jscsrc

But I don't think there is a rule that says "no new empty lines"

lib/controller.js
rwaldron: > I think it is more useful to get an info about where and what is searched... 

Any tessel-centric command allows for `--usb` or `--lan` flags, so this message won't make sense if one of those is used.

bin/tessel-2.js
Student007: I will do :+1:

lib/controller.js
Student007: @rwaldron didn't notice this until I added the current experimental `t2 select` command. But in this case this could be a switch around `opts.lan`:
```
if(opts.lan) {
  logs.info('Searching nearby LAN (netmask) for Tessels...');
} else if(opts.usb) {
  logs.info('Searching USB for your Tessel...');
} else {
  logs.info('Searching nearby LAN (netmask) and USB for Tessels...');
} 
```
**Question**: Because I do not own a single Tessel 2 .. ... . I don't know whether `var seeker = new Seeker ..` also discovers USB Tessels (I assume) ... but does it differ between `--lan`and `--usb` or is this done by filtering later ?

lib/controller.js
Student007: @johnnyman727 do you agree to remove testing flags and only write tests for things what are not blowing up code by lines like this ? Also the stupid testing related organizing code parts like `controller.ssh.seek` are only necessary for testing and looks bad for someone don't write tests... In fact this feels stupid and terminal-menu integration isn't really testable but I think the normal case would be to write unit tests for terminal-menu itself.
What do you prefer ?
@rwaldron maybe your meaning is interesting here (did you wrote the most of the tests - or who has this done (Jon?) ?)

lib/controller.js
Student007: @johnnyman727 got it while writing `t2 select` ;)

johnnyman727:
@Student007 I was away for the weekend but will try to provide more feedback tomorrow. Still digging through emails.

lib/controller.js
johnnyman727: @Student007 I get where you are coming from but 90% of users won't care what types of connections the CLI is looking for. They will just want to deploy their code and it should work regardless of the connection type. Let's remove it from the PR because it's not related to `t2 root` and if you feel strongly about it, let's open an issue so more people can discuss. :)

lib/controller.js
johnnyman727: I think TESSEL_AUTH_KEY and TESSEL_AUTH_PATH should be set in the same place. For someone looking to maintain this code in the future, it would be confusing that one is set in `controller.js` and the other is set in `provision.js`. 

@Student007 I think the better option would be to not set the default in `tessel-2.js` and just substitute the standard Tessel key path in `controller.js`. 

I'm always partially thinking about how we can make this CLI eventually usable for other types of devices and to do that we would need to separate the device key location from the command parsing.

lib/controller.js
johnnyman727: @Student007 with Tessel 2, we presume that if it's connected over USB, then that user has full authorization to use it. All they have to do is run `t2 provision` to now have control over it even if it's not plugged in via USB.

lib/controller.js
johnnyman727: @Student007 that works too! I just want to make sure the grammar is correct.

lib/controller.js
johnnyman727: Yep! Just like if you only have one Tessel and you do `t2 run`, it will automatically deploy to that Tessel. It provides for the best user experience because most people will only have one Tessel.

lib/controller.js
johnnyman727: Ah well, we'll need to address this before this PR can be merged. I'm surprised the child process doesn't try to quit after the SSH session is merged...

lib/controller.js
johnnyman727: @Student007 I would prefer to impact the code as little as possible with unit test modifications. You should be able to create stubs for any of these methods within your unit tests to change the behavior however you need. [Here](https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js#L238-L242) is an example for how to do that.

lib/controller.js
IgitDanny: @johnnyman727 it would be fixed by using `t2 reboot` what is an [other pull request](https://github.com/tessel/t2-cli/pull/328)

lib/controller.js
johnnyman727: @Student007 our model for T2 only involves one SSH key pair for all of your Tessels. I'm not sure what the point of having multiple keys would be.

I do understand that if someone passes in an alternate key path to this particular command then it's possible that it won't be correct and the command will fail.

You'll want to set the Tessel.TESSEL_AUTH_PATH with the provided key before calling `Tessel.get` so that we attempt to use that key to open an SSH connection with any discovered Tessels. That way you will know by this point whether the keys work or not.

johnnyman727:
@Student007 I went through and added more feedback! Let me know if you need help rebasing or making suggested changes.

lib/controller.js
Student007: @johnnyman727 as mentioned I opened an issue https://github.com/tessel/t2-cli/issues/340 for further discussions.

lib/controller.js
Student007: > I'm always partially thinking about how we can make this CLI eventually usable for other types of devices and to do that we would need to separate the device key location from the command parsing.

@johnnyman727  That is a good idea. In that case we should have a second and third VM to be discovered. That way we can hold this in mind.

By writing this, I got awash by TODOs like
- replace all "Tessel" by DEVICETYPENAME
- TESSEL_AUTH_PATH should become '.'+DEVICETYPENAME+AUTH_PATH
- ...++

What type of devices ? Tessel based or also Ra. PI, Intel an others ?
Or do you mean to be able to fork and easy re-engineering this CLI for e.g. Ra. PI for running OpenWRT on it ?

lib/controller.js
Student007: @johnnyman727 OK, thanks, now I understand the security idea. A single Tessel is bond to a single computer after provisioning. I hear ringing the :bell: and see a high frequented FAQ entry:
# Tessel 2 FAQ
> ### My Tessel doesn't work on my second Computer
> For reasons of security your Tessel is only able to `run` and `push` your scripts from a single computer - that one you use `t2 provision`. No other computers are able to connect to your Tessel. But you are able to connect also by LAN from that single computer so you are able to `push` scripts through the air :ghost: :v: 
*-You are visitor 1999*

:see_no_evil: :hear_no_evil: :speak_no_evil: :wink: only my two cents ...

But this is not the worst idea :+1: 

Maybe it would be an idea to provide a solution about using a key-server what accepts sURL based authorization for putting generated keys and use them remotely.

---
Btw: I didn't got this until now, because .... I ... didn't ... own ... a single Tessel 2 ...

lib/controller.js
Student007: @johnnyman727 I will do a new PR but will use this one for discussion. I am waiting to prepare a new PR until the PR for `controller.menu`, `TESSEL_AUTH_KEY-option` and the local branch about `t2 select` is ready and able to merge and are merged - due to the parts I like to use from that PRs.

lib/controller.js
johnnyman727: Yes, devices like RasPi and Edison which also run some version of Linux. It will definitely take some refactoring but it would be an interesting goal.

lib/controller.js
johnnyman727: Yes, it's possible some folks will be confused about that. We were initially planning on making `t2 provision` run automatically but ultimately decided against it so that users don't get confused about trying to deploy to LAN devices that they've never connected over USB before.

lib/controller.js
johnnyman727: I suppose we should break out that logic into it's own functionality that can be called both in `deploy.js` and here.

lib/controller.js
Student007: :+1: I will have this in mind - but I will ignore my current wishes to look over everything at this point - due to you would wait much longer to merge a useful result causing from discussion and feedback :smile:

lib/controller.js
Student007: ok - thanks for talking about backgrounds and internal history.

lib/controller.js
Student007: > @Student007 our model for T2 only involves one SSH key pair for all of your Tessels. I'm not sure what the point of having multiple keys would be.

@johnnyman727 I don't like this because it would be the same as not being able to change PASSWORDs... 
I vote for having `--key` options and review the code for necessary changes as I have done as first step in https://github.com/tessel/t2-cli/pull/330 
Also if you gave the key to someone or got hacked whatever ... **changing and having different keys for different Tessels** should be possible.
My two cents ...
Opened an issue for further discussion https://github.com/tessel/t2-cli/issues/347

johnnyman727: Thanks for making a separate issue. That will allow us to discuss in more detail with more people and merge PRs in the meantime.

johnnyman727:
@Student007 what else do we need to get this PR merged?

johnnyman727:
Closing in favor of #453.

PR #309, opened 2015-09-15
controller.menu

Student007:
as discussed here:
https://github.com/tessel/t2-cli/pull/307/files#r39476687

This pull request is part of the t2 root puzzle with a kind of reusable menu if multiple Tessels are found. They will be shown as a ncurses like menu.

You start to use the menu by 
`controller.menu.prepareMenu(opts, tessels, resolve, reject, title, callback)` 
the `callback` receives `name, index, resolve, reject` where name is the selected Tessels name and Index within Tessels collection.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -8,6 +8,7 @@ var cp = require('child_process');
 var async = require('async');
 var updates = require('./update-fetch');
 var controller = {};
+var Menu = require('terminal-menu');
 
 Tessel.list = function(opts) {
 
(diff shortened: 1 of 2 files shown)


Student007:
@ 08:57:03  http://time.is/GMT ... All checks have passed and the branch is up-to-date with the base branch !

lib/controller.js
```diff
@@ -641,7 +642,105 @@ controller.tesselFirmwareVerion = function(opts) {
...
+      if (!tessel.usbConnection.authorized) {
+        rtm.add(index + ') ' + tessel.name + ': [USB] (not authorized) \n');
+      } else {
+        rtm.add(index + ') ' + tessel.name + ': [USB] (authorized) \n');
+      }
+    }
+    // FIXME: Workaround for trouble with dynamic added menu elements
+    rtm._draw();
+  },
+  showMenu: function(rtm, resolve, reject) {
```
rwaldron: As a method of the `menu` property, this can be called `show` (ie. `controller.menu.show(...)`

lib/controller.js
```diff
@@ -641,7 +642,105 @@ controller.tesselFirmwareVerion = function(opts) {
...
+        // controller.ssh.runSSH(index, opts, tessels, resolve, reject);
+      } else {
+        // going to clear screen and calling exit to resolve promise
+        controller.menu.clear();
+        controller.menu.exit(resolve);
+      }
+    });
+    //  });
+  },
+  makeMenu: function(tessel, index, rtm) {
```
rwaldron: As a method of the `menu` property, this can be called `make` (ie. `controller.menu.make(...)`

lib/controller.js
```diff
@@ -641,7 +642,105 @@ controller.tesselFirmwareVerion = function(opts) {
         });
     });
 };
+controller.menu = {
 
+  prepareMenu: function(opts, tessels, resolve, reject, title, callback) {
```
rwaldron: As a method of the `menu` property, this can be called `prepare` (ie. `controller.menu.prepare(...)`

lib/controller.js
```diff
@@ -641,7 +642,105 @@ controller.tesselFirmwareVerion = function(opts) {
...
+      bg: 'red'
+    });
+    rtm.reset();
+    rtm.write(title + '\n');
+    rtm.write('                                                \n');
+
+    // create menu entries
+    for (var i in tessels) {
+      controller.menu.makeMenu(tessels[i], i, rtm);
+    }
```
rwaldron: Don't use `for-in` for iterating arrays.

```js
tessels.forEach(function(tessel, index) {
  controller.menu.make(tessel, index, rtm); 
});
```

lib/controller.js
Student007: :+1:

lib/controller.js
Student007: :+1:

lib/controller.js
Student007: :+1:

lib/controller.js
Student007: @rwaldron oh, thanks for this edit - didn't know that! :sun_with_face:

Student007:
@johnnyman727 @rwaldron keep the pull request pending. I will rework the methods to be promises.

Student007:
I am out of time today. I will not commit my result because there wasn't time for testing it asynchronously.

Student007:
### API
method | return | info
-----|------|-----
| controller.menu.create(tessels, opts, title) | promise | calls `make`and `show` before returning the selected single `Tessel` object to the next promise. In the case `EXIT` is selected, the `process.exit()` will be executed due to the reason there is not reason to work with nothing. **The params `opts` and `title` are optional !** while `opts` in fact is useless at the moment but `title` will become useful to create an information line above the options to select (e.g. `*You stupid customer forgot to use the awesome command line param TESSEL=WhatTheHellYouCallYourPoorTessel*`).
| controller.menu.make(tessels, rtm, title) | promise | Of cause you may use `make` if you own a `rtm` object. The returned promise will contain the reference to the made menu what is to use with `show` to become useful.
| controller.menu.show(rtm) | promise | Of cause you may use only that to show a menu. The promises resolve will pipe the index like this: `controller.menu.make(tessels, rtm, title).then(controller.menu.show(rtm)).then(function(index) {resolve(tessels[index]);});` 
| controller.menu.clear() | promise | it simple clears the terminal and resolves the returned promise

### Usage
```
controller.menu.create(tessels)
  .then(function(tessel) {
    // do something with the selected Tessel
  });

```

**Note:** This pull request changes a dependencies version: async needs to be updated from `0.9` to `1.4` due to proper implemented method `async.forEachOf` what in difference to `async.each` will provide the arrays key (index) what is necessary to create the menu in the right order.

johnnyman727:
@Student007 this is looking good so far!

I don't see how it connects to the `Tessel.get` process though. [This branch of code](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L162-L167) will be hit if it's ambiguous which Tessel should be selected. This is where the menu should pop up so users can choose.

Student007:
@johnnyman727 yes, I simple prevented to change existing code like `Tessel.get` what could damage behaviors you know better. As shown in the API & Usage you can add this directly in the master after merging. I will smash the commits and then you can go.

lib/controller.js
```diff
@@ -641,7 +642,114 @@ controller.tesselFirmwareVerion = function(opts) {
...
+          if (tessel.lanConnection) {
+
+            if (!tessel.lanConnection.authorized) {
+              rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (not authorized) \n');
+            } else {
+              rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (authorized) \n');
+            }
+
+          } else if (tessel.usbConnection) {
+            if (!tessel.usbConnection.authorized) {
```
johnnyman727: USB Connections are always authorized.

lib/controller.js
```diff
@@ -641,7 +642,114 @@ controller.tesselFirmwareVerion = function(opts) {
...
+              rtm.add(index + ') ' + tessel.name + ': [USB] (authorized) \n');
+            }
+          } else {
+            reject('Type of Tessel unkonwn: ', tessel);
+          }
+          done();
+        },
+        function closed(err) {
+          if (err) {
+            logs.warn(err);
```
johnnyman727: Just pass the error string into `reject` and it will be WARNed automatically by `tessel-2.js`.

lib/controller.js
```diff
@@ -641,7 +642,114 @@ controller.tesselFirmwareVerion = function(opts) {
...
+            reject();
+          } else {
+            rtm.add('EXIT\n');
+            //resolve();
```
johnnyman727: Why not resolve here? If we're not going to use it, let's delete the comment.

johnnyman727:
@Student007 fair enough! I've added a few comments to be addressed.

Student007:
close due to new PR https://github.com/tessel/t2-cli/pull/348 for cleaning up commits.

Issue #272, opened 2015-09-01
t2 run|push <file> -s (single file run/push)


rwaldron:
When the `-s` is present, only the file is deployed.

- Skips the folder delete
- Skips the folder create

Frijol:
is `-s` standard for this?

rwaldron:
I was told this was the flag used for T1 cli?

johnnyman727:
[Indeed!](https://github.com/tessel/cli/blob/master/bin/tessel-run.js#L73-L77)

Frijol:
cool, just checking!

wprater:
I don't mind taking this on, if @rwaldron isn't.

johnnyman727:
I believe @rwaldron is traveling the globe for a few more days. If you've got time time before then, feel free to steal this one!

wprater:
is there a spec somewhere I can look at?  from v1?

johnnyman727:
@wprater unfortunately, I don't think we have a spec for this command. It was tacked on for the T1 CLI and it has been tacked on for the T2 CLI as well. :/ 

My take is that it takes the path to a single file, places that file in `/tmp/remote-script/FILE` or `/app/FILE` (depending on if deployed with `run` or `push`) and runs that file. I don't think it needs to be any more complicated than that.

wprater:
ok, that's what I figured as well!

rwaldron:
Fixed by https://github.com/tessel/t2-cli/pull/311

PR #311, opened 2015-09-16
[WIP] Add singlefile push run

wprater:
A WIP to push a single file.

- you cannot push a single directory
- push a single file should follow project path

not currently writing to the remote store, but wanted to get this up if it's helpful for #272  as I may be offline for a couple days.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -110,7 +110,7 @@ makeCommand('run')
   .option('single', {
     flag: true,
     abbr: 's',
-    help: 'Push only the entryPoint'
+    help: 'Run only the specified entry point file'
   })
   .option('verbose', {
     flag: true,
@@ -132,7 +132,7 @@ makeCommand('push')
   .option('single', {
     flag: true,
     abbr: 's',
-    help: 'Push only the entryPoint'
+    help: 'Push only the specified entry point file'
   })
   .option('verbose', {
     flag: true,
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -107,6 +107,11 @@ makeCommand('run')
     required: true,
     help: 'The entry point file to deploy to Tessel'
   })
+  .option('single', {
+    flag: true,
+    abbr: 's',
+    help: 'Push only the entryPoint'
+  })
   .option('verbose', {
     flag: true,
     abbr: 'v',
@@ -123,6 +128,11 @@ makeCommand('push')
     position: 1,
     required: true,
     help: 'The entry point file to deploy to Tessel'
+  })
+  .option('single', {
+    flag: true,
+    abbr: 's',
+    help: 'Push only the entryPoint'
   })
   .option('verbose', {
     flag: true,
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -48,7 +48,7 @@ Tessel.prototype.deployScript = function(opts) {
         var prom = Promise.resolve();
 
         if (opts.single) {
-          // Always be sure the appropraite dir is created
+          // Always be sure the appropriate dir is created
           prom.then(actions.execRemoteCommand(self, 'createFolder', filepath));
         } else {
           // Delete any code that was previously at this file path
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -111,7 +111,7 @@ actions.execRemoteCommand = function(tessel, command, filepath) {
 actions.findProject = function(opts) {
   return new Promise(function(resolve, reject) {
     var single = opts.single;
-    var file = opts.file;
+    var file = opts.entryPoint;
     var home = process.env[(process.platform === 'win32') ? 'USERPROFILE' : 'HOME'];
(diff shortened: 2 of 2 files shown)


lib/tessel/deploy.js
```diff
@@ -100,7 +108,7 @@ actions.execRemoteCommand = function(tessel, command, filepath) {
   });
 };
 
-actions.findProject = function(file) {
+actions.findProject = function(file, single) {
```
rwaldron: Change this to: 

```js
actions.findProject = function(file, options) {
```

lib/tessel/deploy.js
```diff
@@ -177,16 +189,25 @@ actions.sendBundle = function(t, filepath, opts) {
       .then(function(remoteProcess) {
         var error = '';
 
-        actions.findProject(opts.entryPoint).then(function(project) {
-          var resolved = path.resolve(process.cwd(), project.pushdir);
+        actions.findProject(opts.entryPoint, opts.single).then(function(project) {
```
rwaldron: Change to: 

```js
actions.findProject(opts)
```

rwaldron:
In addtion to the inline review notes, please review the travis-ci failures here: https://travis-ci.org/tessel/t2-cli/builds/80551104 (be sure to always run `grunt`)

lib/tessel/deploy.js
wprater: good call, I went with `opts` since that's what been used in the rest of the file.

bin/tessel-2.js
```diff
@@ -124,6 +129,11 @@ makeCommand('push')
     required: true,
     help: 'The entry point file to deploy to Tessel'
   })
+  .option('single', {
+    flag: true,
+    abbr: 's',
+    help: 'Push only the entryPoint'
```
rwaldron: `entryPoint` => `entry point`

bin/tessel-2.js
```diff
@@ -107,6 +107,11 @@ makeCommand('run')
...
```
rwaldron: `Push only the entryPoint` => `Run only the specified entry point file.`

lib/tessel/deploy.js
```diff
@@ -100,8 +108,10 @@ actions.execRemoteCommand = function(tessel, command, filepath) {
   });
 };
 
-actions.findProject = function(file) {
+actions.findProject = function(opts) {
   return new Promise(function(resolve, reject) {
+    var single = opts.single;
+    var file = opts.file;
```
rwaldron: `var file = opts.file;` => `var file = opts.entryPoint;`

wprater: durr, sorry.  I need to write test for this, but could use help to get myself situated with the testing stubs and methodology we're using.

rwaldron:
To answer your question in the comment, regarding tests: take a look at https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js to get started.

- Create new tests after this block: https://github.com/tessel/t2-cli/blob/master/test/unit/deploy.js#L81-L114 that _almost_ replicate those two, however ensuring that the `deleteFolder` remote command is not called.
- Using the fixtures in test/unit/fixtures, capture the size of a deployment in the "whole project" case, then compare that to single file run/push (there might be better ways to test this)

If I think of more I will post here. I will be back at my regular work area tomorrow and can be more useful then.

wprater:
@rwaldron thanks.  currently a single file does not seem to be writing [here](https://github.com/wprater/t2-cli/blob/add-singlefile-push-run/lib/tessel/deploy.js#L235). will look to tests in the next couple days.

rwaldron:
> currently a single file does not seem to be writing here. will look to tests in the next couple days.

I'll also take a look

rwaldron:
The problem with the current solution is that it's trying to send a buffer containing the contents of a file, but the other end expects a buffer for a tarball. The solution doesn't require two separate reading paths, because fstream-ignore can have new rules added dynamically. I will make a pull request to your branch with the latest changes.

rwaldron:
Actually, it's easier to land this and then merge the modification.

wprater:
OK.. I can make some updates now.. or I can rebase this into one WIP commit?

wprater:
we can just have it use a new command for the single file vs the `untarStdin`?

rwaldron:
> we can just have it use a new command for the single file vs the untarStdin?

Then we have to maintain two code paths that effectively do the same thing.

rwaldron:
> OK.. I can make some updates now.. or I can rebase this into one WIP commit?

Sorry, I missed this after my last comment. I'd already landed yours locally and was working on the tests when this came through.

wprater:
>> we can just have it use a new command for the single file vs the untarStdin?

> Then we have to maintain two code paths that effectively do the same thing.

OK.  will check out what you come up with! thanks

rwaldron:
We still need tests that ensure that the `deleteFolder` command is not called when `tessel.deployScript({ single: true, ... })`

wprater:
will add that test.

PR #316, opened 2015-09-18
use fully resolved path when push|run single file

wprater:
if you push a file from a nested dir, it should untar into correct location

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -193,6 +193,7 @@ actions.sendBundle = function(t, filepath, opts) {
 
         actions.findProject(opts).then(function(project) {
           opts.target = path.resolve(process.cwd(), project.pushdir);
+          opts.resolvedEntryPoint = project.entryPoint;
 
           // Collect error data as received
           remoteProcess.stderr.on('data', function(data) {
@@ -241,7 +242,7 @@ actions.tarBundle = function(opts) {
     });
 
     if (opts.single) {
-      fstream.addIgnoreRules(['*', '!' + opts.entryPoint]);
+      fstream.addIgnoreRules(['*', '!' + opts.resolvedEntryPoint]);
     }
 
     fstream.basename = '';
--- a/test/unit/deploy.js
+++ b/test/unit/deploy.js
@@ -136,7 +136,7 @@ exports['tarBundle'] = {
       target: target
     }).then(function(bundle) {
       test.equal(this.addIgnoreRules.callCount, 0);
-      test.equal(bundle.length, 3072);
+      test.equal(bundle.length, 4608);
       test.done();
     }.bind(this));
   },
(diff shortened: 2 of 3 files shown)


wprater:
not sure why the test is failing now..

wprater:
fixed the test.  I added the `resolvedEntryPoint` versus overwriting the `entryPoint` which was provided initially by user/cli.

wprater:
or perhaps we should have the resolve of `findProject` add this new property and leave along the user provided `entryPoint` one?

wprater:
we could also leave the opts alone, and just pass the `project` dict into `tarBundle`.  I feel changing the `entryPoint` in project was confusing a bit, which is why I chose `resolvedEntryPoint`.

test/unit/deploy.js
```diff
@@ -136,7 +136,7 @@ exports['tarBundle'] = {
       target: target
     }).then(function(bundle) {
       test.equal(this.addIgnoreRules.callCount, 0);
-      test.equal(bundle.length, 3072);
```
rwaldron: Why did this change?

rwaldron:
I pulled this locally and reviewed it, ran tests and they passed, so I landed the patch. Just now I checked out to another branch to rebase, but tried `t2 run nested/index.js -s` before I did and it worked exactly as it should—without this patch.

wprater:
the issue occurs when you try to push a file while in a nested directory.
 the ignored path does not included the cwd portion, so it will default to
the relpath

rwaldron:
Output captured from branch that doesn't have this patch: 

```
rwaldron at nova in ~/robotz/tessel-pir
$ t2 run nested/index.js -s
INFO Connected to bishop over USB
INFO Writing nested/index.js to RAM on bishop (2.56 kB)...
INFO Deployed.
INFO Running nested/index.js...
/tmp/remote-script/nested/index.js
1440599263655 1
1440599263726 1
1440599263733 1
1440599263737 1
1440599263741 1
INFO Stopping script...
rwaldron at nova in ~/robotz/tessel-pir
$ t2 run index.js -s
INFO Connected to bishop over USB
INFO Writing index.js to RAM on bishop (2.048 kB)...
INFO Deployed.
INFO Running index.js...
/tmp/remote-script/index.js
1440599279774 1
1440599279843 1
1440599279850 1
1440599279854 1
1440599279858 1
1440599279861 1
1440599279864 1
INFO Stopping script...
```

rwaldron:
> the issue occurs when you try to push a file while in a nested directory.
> the ignored path does not included the cwd portion, so it will default to
> the relpath

That doesn't appear to be the case in my output above.

wprater:
`cd nested; t2 run index.js -s` works for you?

rwaldron:
I used the same code, in two different files:

- index.js
- nested/index.js


```js
var tessel = require('tessel');

console.log(process.argv[1]);

tessel.port.A.pin[7].input();

function read() {
  tessel.port.A.pin[7].read(function(error, data) {
    console.log(Date.now(), data);
    read();
  });
}

read();


setTimeout(function() {
  process.exit(0);
}, 200);

```

wprater:
perhaps because the file is still there?

rwaldron:
> perhaps because the file is still there?

What file is still where? Look at the paths that are printed out.

lib/tessel/deploy.js
```diff
@@ -241,7 +242,7 @@ actions.tarBundle = function(opts) {
     });
 
     if (opts.single) {
-      fstream.addIgnoreRules(['*', '!' + opts.entryPoint]);
```
wprater: @rwaldron `opts.entryPoint` would be the entry point set for the cwd.  so it's ignoring at the relPath, not the resolved path of the actual file.  because of this, it was still ignoring the file in a nested dir when I was cd'd into that dir.

rwaldron:
> cd nested; t2 run index.js -s works for you?

This is the difference I was looking for.

PR #326, opened 2015-09-18
Fix hueristics bug

### LINKED ISSUE #317: Hueristics not working properly
I get the following even when I have a USB Tessel connected:

```
➜  t2-demo  t2 run index.js --name Frank
INFO Please specify a Tessel by name [--name <tessel name>]
```

johnnyman727:
Fixes #317.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -9,7 +9,9 @@ var async = require('async');
 var updates = require('./update-fetch');
 var inquirer = require('inquirer');
 var colors = require('colors');
+var util = require('util');
 var controller = {};
+
 
 Tessel.list = function(opts) {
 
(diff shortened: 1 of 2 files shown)


lib/controller.js
```diff
@@ -61,21 +61,24 @@ Tessel.list = function(opts) {
...
+          // Figure out which Tessel will be selected
+          controller.runHeuristics(opts, reconciledTessels)
+            .then(function logSelected(tessel) {
+              // Report that selected Tessel to the user
+              logs.info('Multiple Tessels found.');
+              if (tessel) {
+                logs.info('Will default to', tessel.name, '.');
+              }
+              // Helpful instructions on how to switch
+              logs.info('Set default Tessel with environment variable (e.g. \'export TESSEL=bulbasaur\') or use the --name flag.');
```
rwaldron: nit: Instead of `\'`, just use `"`

johnnyman727:
@rwaldron could you review when you get a chance?

This PR fixes two things:
1. If multiple LAN Tessels were discovered, the heuristics function would claim that it couldn't deduce which Tessel to use even if a usb Tessel was connected.
2. The heuristics function weirdly returned nothing if it could deduce which Tessel to use. This PR fixes change those cases so that an error is `Promise.reject`ed which I believe is a bit more sanitary.

rwaldron:
LGTM!

(Great tests!)

PR #328, opened 2015-09-18
Reboot

Student007:
due to @flaki mention on `slack review channel` here is a suggestion for
a possible `t2 reboot` command.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -86,7 +86,7 @@ parser.command('reboot')
   .option('force', {
     abbr: 'f',
     required: false,
-    help: 'don\'t go through init'
+    help: 'Do not go through init'
   })
   .option('nosync', {
     abbr: 'n',
@@ -94,9 +94,14 @@ parser.command('reboot')
     help: 'Do not sync'
   })
   .option('name',   {
-    required: true,
+    required: false,
     metavar: 'TESSEL',
-    help: 'Tessels name or IP'
+    help: 'Tessels name'
+  })
+  .option('ip', {
+    required: false,
+    metavar: 'IP-ADDRESS',
+    help: 'Tessels IP-ADDRESS (no authorization check!)'
   })
   .option('key', {
     abbr: 'i',
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -69,6 +69,44 @@ parser.command('provision')
   })
   .help('Authorize your computer to control the USB-connected Tessel');
 
+parser.command('reboot')
+  .callback(callControllerCallback('reboot'))
+  .option('time', {
+    abbr: 't',
+    metavar: 'SEC',
+    required: false,
+    help: 'Delay interval until reboot (cannot be stopped!)'
+  })
+  .option('delay', {
+    abbr: 'd',
+    metavar: 'SEC',
+    required: false,
+    help: 'Same as --time'
+  })
+  .option('force', {
+    abbr: 'f',
+    required: false,
+    help: 'don\'t go through init'
+  })
+  .option('nosync', {
+    abbr: 'n',
+    required: false,
+    help: 'Do not sync'
+  })
+  .option('name', {
+    required: true,
+    metavar: 'TESSEL',
+    help: 'Tessels name or IP'
+  })
+  .option('key', {
+    abbr: 'i',
+    default: '~/.tessel/id_rsa', // TODO: replace this by global var
+    required: false,
+    metavar: 'PATH',
+    help: 'SSH-Key to login into your Tessel. Default is:'
+  })
+  .help('Reboot your Tessel immediately or with delay');
+
 makeCommand('restart')
   .callback(function(opts) {
     var packageJson;
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -93,7 +93,7 @@ parser.command('reboot')
     required: false,
     help: 'Do not sync'
   })
-  .option('name', {
+  .option('name',   {
     required: true,
     metavar: 'TESSEL',
     help: 'Tessels name or IP'
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -8,6 +8,7 @@ var cp = require('child_process');
 var async = require('async');
 var updates = require('./update-fetch');
 var controller = {};
+var debug = require('debug')('controller');
 
 Tessel.list = function(opts) {
 
@@ -176,6 +177,63 @@ may or may not be open and closes them
 
 */
 
+/*
+Reboots your tessel
+*/
+
+controller.reboot = function(opts) {
+  return new Promise(function(resolve, reject) {
+    Tessel.get(opts)
+    .then(function(tessel){
+      var command = 'ssh root@'+tessel.lanConnection.ip+' -i '+opts.key+' \'reboot ';
+      if (opts.time || opts.delay) {
+        command = command+' -d '+opts.time || opts.delay;
+      }
+      if(opts.force) {
+        command = command+' -f ';
+      }
+      if(opts.nosync) {
+        command = command+' -n ';
+      }
+      command = command+'\'';
+      debug(command);
+      var reboot_cp = cp.exec(command);
+
+      reboot_cp.on('error', function(e){
+        if(e === 255){
+          logs.warn('Not authorized (maybe wrong key)');
+          reject();
+        } else if (e === 127) {
+          // command not found
+          logs.warn('Command "reboot" not found!');
+          reject();
+        }
+        else {
+          reject(e);  
+        }
+      });
+      reboot_cp.on('exit', function(e){
+        if(e === 255){
+          logs.warn('Not authorized (maybe wrong key)');
+          reject();
+        } else if (e === 127) {
+          // command not found
+          logs.warn('Command "reboot" not found!');
+          reject();
+        } else if (e === 0) {
+          logs.info(tessel.name +' is going to reboot!');
+          resolve();
+        } else {
+          reject(e);  
+        }
+      });
+
+  }).catch(function (err) {
+    logs.err(err+' ('+opts.name+')');
+  });
+
+  });
+};
 
 controller.closeTesselConnections = function(tessels) {
   return new Promise(function(resolve, reject) {
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -176,59 +176,70 @@ Takes a list of Tessels with connections that
 may or may not be open and closes them
 
 */
-
 /*
 Reboots your tessel
 */
-
 controller.reboot = function(opts) {
   return new Promise(function(resolve, reject) {
+    var host;
+    if (opts.ip) {
+      Tessel.get = function(opts) {
+        return new Promise(function(resolve) {
+          debug('Rebooting remote IP: ', opts.ip);
+          resolve();
+        });
+      };
+    } else if (!opts.name) {
+      logs.err('Either --name or --ip is required!');
+      resolve();
+    }
     Tessel.get(opts)
       .then(function(tessel) {
-        var command = 'ssh root@' + tessel.lanConnection.ip + ' -i ' + opts.key + ' \'reboot ';
+        if (!tessel && opts.ip) {
+          debug('Prevented looking for Tessels!');
+          host = opts.ip;
+        } else if (tessel && !opts.ip) {
+          debug('Tessel found!');
+          host = tessel.lanConnection.ip;
+        }
+        var command = 'ssh root@' + host + ' -i ' + opts.key + ' \'reboot ';
         if (opts.time ||  opts.delay)  {
-          command = command + ' -d ' + opts.time || opts.delay;
+          command += ' -d ' + opts.time || opts.delay;
         }
         if (opts.force) {
-          command = command + ' -f ';
+          command += ' -f ';
         }
         if (opts.nosync) {
-          command = command + ' -n ';
+          command += ' -n ';
         }
-        command = command + '\'';
+        command += '\'';
         debug(command);
-        var reboot_cp = cp.exec(command);
+        var reboot = cp.exec(command);
 
-        reboot_cp.on('error', function(e) {
+        function rebootHandler(e) {
           if (e === 255) {
             logs.warn('Not authorized (maybe wrong key)');
-            reject();
+            resolve();
           } else if (e === 127) {
             // command not found
             logs.warn('Command "reboot" not found!');
-            reject();
-          } else {
-            reject(e);
-          }
-        });
-        reboot_cp.on('exit', function(e) {
-          if (e === 255) {
-            logs.warn('Not authorized (maybe wrong key)');
-            reject();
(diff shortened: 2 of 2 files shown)


bin/tessel-2.js
```diff
@@ -69,6 +69,44 @@ parser.command('provision')
   })
   .help('Authorize your computer to control the USB-connected Tessel');
 
+parser.command('reboot')
```
rwaldron: `parser.command('reboot')` => `makeCommand('reboot')`

bin/tessel-2.js
```diff
@@ -69,6 +69,44 @@ parser.command('provision')
...
+  .option('delay', {
+    abbr: 'd',
+    metavar: 'SEC',
+    required: false,
+    help: 'Same as --time'
+  })
+  .option('force', {
+    abbr: 'f',
+    required: false,
+    help: 'don\'t go through init'
```
rwaldron: As you've done below: 

`don\'t` => `Do not`

bin/tessel-2.js
```diff
@@ -69,6 +69,44 @@ parser.command('provision')
...
+  .option('nosync', {
+    abbr: 'n',
+    required: false,
+    help: 'Do not sync'
+  })
+  .option('name',   {
+    required: true,
+    metavar: 'TESSEL',
+    help: 'Tessels name or IP'
+  })
```
rwaldron: After you switch to `makeCommand`, this option can be removed.

bin/tessel-2.js
```diff
@@ -69,6 +69,44 @@ parser.command('provision')
...
+  .option('key', {
+    abbr: 'i',
+    default: '~/.tessel/id_rsa', // TODO: replace this by global var
+    required: false,
+    metavar: 'PATH',
+    help: 'SSH-Key to login into your Tessel. Default is:'
+  })
```
rwaldron: Maybe this should be added into `makeCommand`?

lib/controller.js
```diff
@@ -176,6 +177,62 @@ may or may not be open and closes them
...
+        }
+        if (opts.force) {
+          command = command + ' -f ';
+        }
+        if (opts.nosync) {
+          command = command + ' -n ';
+        }
+        command = command + '\'';
+        debug(command);
+        var reboot_cp = cp.exec(command);
```
rwaldron: nit: `reboot_cp` => `reboot`

lib/controller.js
```diff
@@ -176,6 +177,62 @@ may or may not be open and closes them
...
+      .then(function(tessel) {
+        var command = 'ssh root@' + tessel.lanConnection.ip + ' -i ' + opts.key + ' \'reboot ';
+        if (opts.time ||  opts.delay)  {
+          command = command + ' -d ' + opts.time || opts.delay;
+        }
+        if (opts.force) {
+          command = command + ' -f ';
+        }
+        if (opts.nosync) {
+          command = command + ' -n ';
```
rwaldron: nit: `command = command + ' -n ';` => `command += ' -n ';` (and the others as well)

lib/controller.js
```diff
@@ -176,6 +177,62 @@ may or may not be open and closes them
...
+          } else if (e === 127) {
+            // command not found
+            logs.warn('Command "reboot" not found!');
+            reject();
+          } else if (e === 0) {
+            logs.info(tessel.name + ' is going to reboot!');
+            resolve();
+          } else {
+            reject(e);
+          }
```
rwaldron: Could these be consolidated into a single handler that gets share by both events?

lib/controller.js
```diff
@@ -176,6 +177,62 @@ may or may not be open and closes them
...
+/*
+Reboots your tessel
+*/
+
+controller.reboot = function(opts) {
+  return new Promise(function(resolve, reject) {
+    Tessel.get(opts)
+      .then(function(tessel) {
+        var command = 'ssh root@' + tessel.lanConnection.ip + ' -i ' + opts.key + ' \'reboot ';
+        if (opts.time ||  opts.delay)  {
```
rwaldron: Looks like an extra space here? Please run `grunt`

bin/tessel-2.js
wprater: the `key` option?  not a bad idea.

Student007: :+1:

lib/controller.js
Student007: @rwaldron You are absolutely right! Maybe the the `controller.ssh` could get a `controller.ssh.handler` what is used by `t2 reboot` as well ?

bin/tessel-2.js
Student007: @rwaldron did that first try, but the default options are too much ...

bin/tessel-2.js
Student007: About the `key` option to become a default option in `makeCommand` is worth to think about, because the discussed things behind `t2 select` make this a fitting solution.

Student007:
@rwaldron @johnnyman727 if/when nothing has to be changed, I would smash the commits. While trying to work around the `--ip` flag (see last commit) I discovered parts of code would be usable to execute a single cmd. I forgot the reason not to use this (I think because it didn't work well with Tessels outside of netmask)

johnnyman727:
@Student007 thanks for working on this. I'm still not convinced of the use case. When would you want to reboot your T2 and wait for it to come up? It seems like this is a patch over other issues and we should actually be solving those other issues.

@rwaldron @flaki would like to hear your thoughts as well.

Student007:
@johnnyman727 I you would use your Tessel somewhere outside of the netmask of your provision computer you are currently not able to access it. By adding this pull request you would be able to reboot a Tessel remotely by using `--ip` instead of `--name`. On the other hand it will hide a freezing terminal when logged in to the Tessel by real `ssh -i ~/.tessel/id_rsa root@myLocalTesselWithinNetmaskHostname`.

johnnyman727:
@Student007 Thanks for the clarification but I still don't understand why a user would want to reboot their Tessel. It seems like they should never be in a position where a reboot should be necessary.

Student007:
>  It seems like they should never be in a position where a reboot should be necessary.

@johnnyman727  there is a used case: I think you agree the command `t2 restart myEntryPoint.js --ip 123.123.123.123` could fail because a user didn't test the `hot fix` or a module produces unexpected values to the `myEntryPoint.js` what freeze the Tessel. In that case he likes to use `t2 reboot --ip 123.123.123.123` because it is V8 what gets frozen not the entire Tessel. There is no `--ip` ? ... OK you are right :wink: but I think it should be added. If it isn't there, your only capability to restart a script remotely would be the `t2 reboot --ip 123.123.123.123` command what could save the day from finding someone is taking kids from kindergarten because you are required to travel to the Tessel :baby: :bomb: 

**Every service technician would miss that feature.**

Of cause you could say, in production mode, a technician would be able to use the `ssh` but in reality and thinking about who is buying and using the Tessel, that people don't got in mind this nice piece of electronics is a full stack linux computer of cause has a ssh server. Also a lot beginners never would get the idea the `t2 key generate` is generating `ssh keys` ... they do not think up to that point. It is not because they are stupid - it simple abstracts complexity without the need of asking deeper questions when following a goal to hack a cool idea for optimizing a process in their company or at their customer.

Another used case would be for example you hacked together a garage door remote control APP with your Tessel to open that door but for some reason the Tessel-APP thinks it is still open. Because you are a hacker sitting in your car before the closed door, you take your laptop boot it up while letting the cars engine running and reboot your Tessel but not over your WiFi (for security it is only reachable in your house) ... over the internet of cause :jack_o_lantern: :ghost:

johnnyman727:
> @johnnyman727 there is a used case: I think you agree the command t2 restart myEntryPoint.js --ip 123.123.123.123 could fail because a user didn't test the hot fix or a module produces unexpected values to the myEntryPoint.js what freeze the Tessel. 

@Student007 I don't think it's very likely that a user script will completely freeze the Tessel. I've never seen that happen even with the most intensive scripts. 

If there are issues that are causing the T2 to freeze or become unresponsive, we should fix those issues instead of rebooting the machine.

Additionally, this command might confuse people because you're not actually rebooting the co-processor (which facilitates USB communications) and could be causing the T2 to appear unresponsive.

I can see the use of it at this stage while we're still flushing out the bugs but I'd like to hear from @rwaldron, @wprater , @Frijol, or @flaki for their opinions of this command.

Sorry to be providing friction, @Student007 - I just think that every new command we add takes away from the simplicity of Tessel.

Student007:
@johnnyman727 @rwaldron @wprater @Frijol @flaki as a compromise we could make the command `t2 reboot` a hidden command as well as the `--ip` option.
> you're not actually rebooting the co-processor (which facilitates USB communications) 

`--ip` is no USB :wink: 
But of cause you are right: It would be more interesting to implement a `t2 reset --ip` what would reset the hole system. In that case it would be necessary to write a very little [daemon](http://wiki.openwrt.org/doc/techref/initscripts) for OpenWRT what is listening to service port and needs authorization with private key as well as ssh does. That daemon could reset and reboot things more useful.

> Sorry to be providing friction[...]

that is no problem if it feels to be democracy at the end of the day :+1: 

> [...] while we're still flushing out the bugs [...]

I belief the not reported bug about `terminal freeze when rebooting Tessel out of ssh login` will never be fixed. It would be necessary to rewrite the command `reboot` of the OpenWRT to check whether existing ssh connections are still open and send them a kill signal to shut down (and wait for it is done) before reboot is initialed. (The reason there is not bug report is about the `t2 root` command is pending in development due to discussion and additional requirements like `--key` and `--ip` as well as the open pull request `controller.menu` what I tried to smash commits but fails ...)

> [...]every new command we add takes away from the simplicity of Tessel.

In this case it is a simplification of 
- login via ssh
- type reboot
- close the terminal
- open a new terminal
- change directory to your current Tessel script development what froze V8-Engine because you are such a stupid Tessel customer tried to force the awesome Tessel 2 to do things your local machine never would do 4 you 

`t2 reboot # because I had been a fool`

johnnyman727:
> I belief the not reported bug about terminal freeze when rebooting Tessel out of ssh login will never be fixed.

If we can't make a command not freeze, then we should not have it at all. :/ I would bet that we can make `t2 root` work w/o issues - I imagine lots of people have done it with Node.

I still think `reboot` as a command would only cover up issues instead of encouraging developers to fix the root cause.

Frijol:
Not sure I follow the whole argument... I can see a use case for `t2 reboot` as long as it's the same as completely power cycling the Tessel. Great for testing stuff out, especially while developing for Tessel. However, @Student007  it sounds like you're proposing something different with the same wording, which is confusing.

Seems fairly contentious; should we boot up the [RFC process](https://github.com/tessel/rfcs)?

johnnyman727:
@Frijol As much as I dislike inconsistency, I feel like the RFC process might be too heavyweight for this situation. The only difference is that we would post it on the Forums - but most of the people that would have an opinion at the moment frequent Github more than the Forums and have already been tagged.

Student007:
@Frijol 
> Not sure I follow the whole argument... I can see a use case for t2 reboot as long as it's the same as completely power cycling the Tessel. Great for testing stuff out, especially while developing for Tessel. 

I think it is more I wrote too much :wink: 
**But you are right.
@kevinmehall or @rwaldron Would you give me a T2 Firmware command what replaces `reboot` for resetting the hole Tessel2 (Instead of simple rebooting the Linux)?

@johnnyman727 
Btw: As I discovered `nomnom` only allows to hide *options* but no *commands* :cry: 

> If we can't make a command not freeze, then we should not have it at all. :/ 

It is not a CLI-Command what freezes - it is `ssh root@tesselname 'reboot'` what freeze. This issue is due to OpenWRT. So this argument don't fit here.

> I would bet that we can make t2 root work w/o issues 

It is no `t2 root` issue - people simple will run into that problem same as `ssh root@tesselname 'reboot'` would do. You cannot expect someone types `reboot` and hits enter. 

> I imagine lots of people have done it with Node.

YES - and then they / @flaki  did: 
- close the terminal
- open a new terminal
- change directory to the current Tessel script development

I think I should create a fan page for people got frozen terminals :+1: who lost their hole history.
# FAQ
### Lost history by frozen terminal
If you try to reboot your Tessel by `ssh root@tesselname 'reboot'` or while logged in by `t2 root` and typing `reboot` and hit enter your will got a frozen terminal. 
**Solution:** It would be better to unplug your Tessel and reconnect it to power. That way you will save your hole history of the day 
*-visitors: Only you (and the other anonymous idiot tried to write a `t2 reboot` cmd to increase usability)*

kevinmehall:
You can hit `<enter> ~ .` to break out of a frozen SSH session.

Student007:
@kevinmehall :fire: your joke :bomb:  destroyed my 6-Day history :boom: ...
oh ... whats that ... it is :raised_hands: magic ... it came back a few seconds later.

I would accept to hit `<enter>` but `<enter> ~ .` I cannot accept. Before a few month I didn't know how to make the `~` on a German keyboard. Next year I plan to do Tessel-Workshops at Albert-Ludwigs University Freiburg - and they all need to learn this trick (with German Keyboards). Of cause all will read this thread when I need to apologize... 
Also if there will not be a `t2 reset` or `t2 reboot` command in the master - I want to prepare a patch people could use if they need it.

Kevin, please tell me what I asked for :smile_cat:

kevinmehall:
You can change the `EscapeChar` SSH option, but you probably don't want SSH to disconnect every time you press enter.

I agree with Jon -- fix the other issues to make this unnecessary. That includes figuring out why the `reboot` command on Tessel OpenWrt doesn't disconnect your SSH session cleanly and fix that.

Student007:
@kevinmehall 
> You can change the EscapeChar SSH option, but you probably don't want SSH to disconnect every time you press enter.

this wasn't what I asked for.

>  figuring out why the reboot command on Tessel OpenWrt doesn't disconnect your SSH session cleanly 

I know why (wrote above).

> and fix that

:-1: will not do that. I am not dived in `t2 runtime` and have no additional time doing that - thanks.

@ all Student XXX: I did my job - no reason to apologize :smile:

kevinmehall:
I doubt it would require "rewriting the reboot command". `reboot` just tells init to perform a shutdown, for one. I believe init does run shutdown scripts to stop services. The first step would be to confirm that `/etc/init.d/dropbear stop` is called on shutdown, the second would be to see if that command executed over serial closes SSH connections.

I'd expect it to be a very simple bug, given that the routers OpenWrt designs for are rarely rebooted.

johnnyman727:
I'm going to go ahead and close this PR because we weren't able to come to a consensus on the `reboot` command. Thanks for suggesting the change and starting the discussion @Student007!

Student007:
@flaki that is so nice :smile_cat: you ask for this ... but I gave up to fight against windmill vanes ... 

I remembered about some situations on issue discussions I followed by reading, where I though I should link this PR for having fun (since discussion ended).

Maybe more people will ask for it. Then its done fast :+1: 
For me it is OK if we didn't got an agreement - it is democracy.

PR #330, opened 2015-09-21
`Key` option and TESSEL_AUTH_KEY

Student007:
As mentioned by @johnnyman727 at
https://github.com/tessel/t2-cli/pull/290/files#r39475328 and liked by
@wprater at https://github.com/tessel/t2-cli/pull/328/files#r39893445 I
am going to make a suggestion how to do by this pull request.

### Notes
It was necessary to change a test about `update` command (assume
@rwaldron wants to check this).

### Usage
Each command what likes to have a default value for the `option('key')`
would be able to use `TESSEL_AUTH_KEY`.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -6,7 +6,7 @@ var controller = require('../lib/controller');
 var key = require('../lib/key');
 var init = require('../lib/init');
 var logs = require('../lib/logs');
-
+var TESSEL_AUTH_KEY = require('../lib/tessel/provision.js').TESSEL_AUTH_KEY;
 
 function closeSuccessfulCommand() {
   process.exit(0);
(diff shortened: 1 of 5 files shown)


johnnyman727:
@Student007 Thanks for putting this together. It would be great if we could set the default provision path in `provision.js` from the `controller.js`. That way, when we discover Tessels, we can try connecting with that provided key. Something like:

```.js
if (opts.key) {
  provision.setDefaultPath(opts.key);
}

Tessel.get()
.then(function(tessel) {
  ....
});
```

What do you think? Then we don't have to be pulling the key into `tessel-2.js` and it cleans up the code quite a bit.

Student007:
closed due to PR https://github.com/tessel/t2-cli/pull/349 what replaces this one

Issue #336, opened 2015-09-24
Deploy doesn't send the project contents over sometimes


johnnyman727:
I attempted to deploy the blinky example:
```
➜  t2-demo  t2 run index.js
INFO Looking for your Tessel...
INFO Connected to Harold over LAN
INFO Writing index.js to RAM on Harold (3.072 kB)...
INFO Deployed.
INFO Running index.js...
module.js:334
    throw err;
          ^
Error: Cannot find module '/tmp/remote-script/index.js'
    at Function.Module._resolveFilename (module.js:332:15)
    at Function.Module._load (module.js:282:25)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:117:18)
    at node.js:948:3
INFO Stopping script...
```

When I checked the T2 directly, there `/tmp/remote-script` existed but had no file contents.

johnnyman727:
I then immediately experienced #319.

johnnyman727:
Here is a clue - it looks like the code starts to untar the contents of the project before the process that creates the folder has been completed:

```
➜  t2-demo  DEBUG=deploy t2 run index.js --timeout 3
INFO Looking for your Tessel...
INFO Connected to Harold over USB
  deploy Executing command: +0ms stopRunningScript
  deploy Command: +64ms stopRunningScript Completed
  deploy Executing command: +2ms deleteFolder
  deploy Command: +11ms deleteFolder Completed
  deploy Executing command: +1ms createFolder
  deploy Beginning processing of untarring stdin +1ms
  deploy Command: +36ms createFolder Completed
INFO Writing index.js to RAM on Harold (3.072 kB)...
  deploy Process opened. Writing stdin +1ms
INFO Deployed.
INFO Running index.js...
I'm blinking! (Press CTRL + C to stop)
```

johnnyman727:
Fixed with #337.

Issue #319, opened 2015-09-18
t2 run can die with "/tmp/remote-script" not found


tcr:
The solution is to wait, but also "killall node" also seems to work, even though it doesn't kill anything!

</armchair debugging>

johnnyman727:
Note: this occurs at least as early as 4c773cf.

johnnyman727:
Another note: fixed by deleting `/tmp/remote-script` on the T2. Note sure why.

johnnyman727:
Yeah, this one is really killer and will be at the top of my to-do list.

johnnyman727:
Fixed with #337.

PR #337, opened 2015-09-24
[Bug Fix] Wait for deploy folder to be created before deploying to it

johnnyman727:
@wprater could you review?

I believe we had a small regression in #311 that is causing #336 and #319.

This change ensures that the `createFolder` command completes before moving on to deploying to that folder.

### DIFF
--- a/test/unit/deploy.js
+++ b/test/unit/deploy.js
@@ -160,6 +160,57 @@ exports['Tessel.prototype.deployScript'] = {
       test.done();
     }.bind(this));
   },
+
+  processCompletionOrder: function(test) {
+
+    var self = this;
+
+    // Array of processes we've started but haven't completed yet
+    var processesAwaitingCompletion = [];
+    self.tessel._rps.on('control', function(data) {
+      // Push new commands into the queue
+      processesAwaitingCompletion.push(data);
+    });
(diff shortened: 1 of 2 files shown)


wprater:
@johnnyman727 will take a look in a few!

test/unit/deploy.js
```diff
@@ -160,6 +160,57 @@ exports['Tessel.prototype.deployScript'] = {
       test.done();
     }.bind(this));
   },
+
+  processCompletionOrder: function(test) {
```
wprater: is this method for another commit / fix?

johnnyman727: No, this is a test that will fail if we ever have a regression back to the issue that this PR fixes.

The issue was that the project contents were being written to `/tmp/remote-script` before the process that created that folder completes. This test ensures that each step of the deployment process (including folder creation) completes before the previous starts.

rwaldron:
This is a pretty brilliant fix—nice work @johnnyman727!!

wprater:
nice work, I learned something new.  had no idea of the `tessel._rps` events emitted.  thanks!

johnnyman727:
Thanks!

PR #342, opened 2015-09-27
Ssh client on close

Student007:
When `seeker.on('tessel', ....)` is discovering a Tessel and the SSH client didn't got further responses from the found Tessel, there are additional events we have to listen to.
**This issue is hard to be reproduced** but from my point of view this will be prevented by adding new event handlers to `lib/lan_connection.js` `LAN.Connection.prototype.open`

### Error
```
Unhandled rejection Error: No response from server
    at Socket.<anonymous> (/Users/{anonym}/github/t2-cli/node_modules/ssh2/lib/client.js:226:15)
    at emitOne (events.js:82:20)
    at Socket.emit (events.js:166:7)
    at TCP.close (net.js:464:12)
```
### Solution based on
As [described](https://github.com/mscdex/ssh2#client-events) I added two additional event handlers:

- end() - The socket was disconnected.
- close(< boolean >hadError) - The socket was closed. hadError is set to true if this was due to error.

**This pull request adds this both event handlers within `lib/lan_connection.js` `LAN.Connection.prototype.open` !**


Student007:
@johnnyman727 sorry for the mixed in TESSEL_AUTH_KEY pull stuff. The new version of GitHub-GUI for Mac seams to fool me about syncing with my master instead of HEAD...

johnnyman727:
@Student007 it looks like you created this new branch off of your key branch. You can run `git fetch origin; git rebase -i origin/master` and delete all commits except the last one to fix that. As an aside, I recommend moving off of Github-GUI. I used it when I first started learning git but I found that it abstracted too much and I didn't learn the details of how to use the `git` tool well.

Which of these two events will be called when you get the `Unhandled rejection Error: No response from server` error?

Student007:
@johnnyman727 Not to use GitHub-GUI for some time would be a good idea maybe. @Frijol Maybe it is a good idea to add this to the howto-contribute session (for cmd lovers as a git-cmd.rd) showing a command line workflow about the following problem:
**When I like to be up to date with master - what is the cmd workflow to:**
- make a branch
- sync branch with HEAD while working (especially to test the solution with actual HEAD) 
- squash multiple commits to have a pull request with only one commit
- remove commits (if possible with no risk?)
- also point on workflow while discussion of pull requests (additional commits what will be squashed once more finally)
- common rebasing cases

@LinusU helped me at https://github.com/tessel/t2-cli/pull/243 to get it done and this style helped me much times (looking onto again and again). But due to updating GitHub GUI I got confused and lost overview - the new features promised to be helpful for syncing... but ... it doesn't make clear whether you are syncing with your forks master or the heads master.
I think @johnnyman727 is right - I should do all open pull requests by command line based on a clear Linus-like description matching the Tessel guidlines.

Student007:
> Which of these two events will be called when you get the Unhandled rejection Error: No response from server error? 

Out of the [ssh2 module](https://github.com/mscdex/ssh2/blob/master/lib/client.js#L226) there is a error event is fired.
The 'close' event fixes this error.
> When does end get called as opposed to close?

I think the only difference is the ability to notice it is due to an error (hadError = true/false). Maybe `end` was added earlier but stay due to be compatible to older versions.

### I think there is additional need for discussion how to handle this by the seeker
An additional `self.emit('close', hadError);` what will be listened on within the tessel object should be the right solution.
```
self.ssh.on('close', function(hadError) {
      // due to a closed connection is no more authorized
      self.emit('close', hadError);
      if (hadError) {
        debug('No response from server', self.host);
        logs.warn('Tessel seams to stall ssh connection!');
        resolve(self);
      } else {
        debug('SSH connection to Tessel ' + self.host + ' closed!');
      }
      resolve(self);
    }).connect(self.auth); // trying to reconnect
```
and `discover.js` connectionHandler` will get additional listener like (e.g.):
```
function connectionHandler(conn) {
    var tessel = new Tessel(conn);
    tessel.on('close', function(hadError) {
      if (hadError) {
        debug('Tessel lost connection due to an error');
        self.emit('stalled',tessel);
        pendingOpen.splice(pendingOpen.indexOf(tessel),1);
      } else {
        debug('Tessels connection had been closed');
        self.emit('close',tessel);
        pendingOpen.splice(pendingOpen.indexOf(tessel),1);
      }
    });
//...
```
`controller.js` needs listeners on `stalled` (e.g.):
```
 seeker.on('stalled', function(tessel) {
       foundTessels.splice(foundTessels.indexOf(tessel),1);
    });
```

Student007:
> @Student007 does this event get called instead of ready? Could it be called after ready if, for example, the network drops out?

YES - this is what happens I think.
>  Fortunately, Promises natively restrict resolve from being called twice but it would be nice to have some clarity on the order of operations.

I assume the `close` event is only fired after a `ready` event talks about the ssh connection got authorized:
- ready() - Authentication was successful.

Otherwise the `error` event would be fired I think:
- error(< Error >err) - An error occurred. A 'level' property indicates 'client-socket' for socket-level errors and 'client-ssh' for SSH disconnection messages. In the case of 'client-ssh' messages, there may be a 'description' property that provides more detail.


About the promises I hadn't got it at this case. I simple thought they are necessary.

Student007:
@johnnyman727 due to `tessel.on('close', function(hadError) {...})` I need some help where and whether to catch the cause of seekers ssh client event `close`. 
In my opinion it needs to be listened on the referenced Tessel-Object itself, but this doesn't work and cause in calling a method of undefined error.
Maybe you have in idea while reading my last comment how to solve this better than my first idea.

lib/tessel/provision.js
```diff
@@ -258,6 +258,7 @@ util.inherits(AlreadyAuthenticatedError, Error);
 actions.AlreadyAuthenticatedError = AlreadyAuthenticatedError;
 
 module.exports = actions;
+module.exports.TESSEL_AUTH_KEY = filepath;
```
rwaldron: This should be defined as an accessor property on `Tessel`:

```js
Object.defineProperty(Tessel, 'TESSEL_AUTH_KEY', {
  get: function() {
    return filepath;
  },
  set: function(value) {
    filepath = value;
  }
});
```

Student007: @rwaldron it is in an other [PR](https://github.com/tessel/t2-cli/pull/349/files#diff-464d0e2fb3c59e8716aa5cacbb54b6b8R18) (forgot to remove it here)

johnnyman727:
@Student007 I recommend placing the event listener in `tessel.js` (on Tessel creation) as you suggested:

```.js
if (this.connection.connectionType === 'LAN') {
  this.closed = true;
}
```

And then I recommend adding a check to the `exec` function (in both `usb_connection.js` and `lan_connection.js`) to make sure the connection is not closed before executing a command.

Student007:
@johnnyman727 could you please make a PR to tessel/master tessel.js and add a tested event listener ?
I am not able to get it done and got sick of this task.
I will do a new PR after rebasing when your PR is merged.

johnnyman727:
@Student007 I haven't forgotten about this! Just haven't had time in the past week. I'll get to it sometime this week.

Student007:
@johnnyman727 OK - then I am triggered by this and will work on it when you've done.

johnnyman727:
@Student007 how do you feel about https://github.com/tessel/t2-cli/pull/392 as an alternative?

Student007:
:+1:

PR #344, opened 2015-09-27
Emits warning if name option provided but Tessel with that name is not found

### LINKED ISSUE #338: Confusing interaction with --name flag if none found by that name
If you pass in the `--name` flag to the CLI but no Tessels are found by that name, the CLI will happily default to the next available Tessel.

I think if a `--name` is explicitly provided, we should display a warning that no Tessels were found by that name and not use any other discovered Tessels.

johnnyman727:
Fixes #338.

@Frijol could you review?

### DIFF
--- a/test/unit/tessel.js
+++ b/test/unit/tessel.js
@@ -107,6 +107,25 @@ exports['Tessel (get)'] = {
     Tessel.get({
         timeout: 0.01
       })
+      // If Tessels were returned, this test should fail because we're
+      // not emitting any Tessels to the seeker
+      .then(function(tessels) {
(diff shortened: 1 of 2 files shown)


test/unit/tessel.js
```diff
@@ -107,6 +107,24 @@ exports['Tessel (get)'] = {
     Tessel.get({
         timeout: 0.01
       })
+      // If Tessels were returned, this test should fail
```
Frijol: Can you explain this a little more?

Frijol:
Works great!

johnnyman727:
Done!

PR #348, opened 2015-09-28
controller.menu cleaned up

Student007:
### API
method | return | info
-----|------|-----
| controller.menu.create(tessels, opts, title) | promise | calls `make`and `show` before returning the selected single `Tessel` object to the next promise. In the case `EXIT` is selected, the `process.exit()` will be executed due to the reason there is not reason to work with nothing. **The params `opts` and `title` are optional !** while `opts` in fact is useless at the moment but `title` will become useful to create an information line above the options to select (e.g. `*You stupid customer forgot to use the awesome command line param TESSEL=WhatTheHellYouCallYourPoorTessel*`).
| controller.menu.make(tessels, rtm, title) | promise | Of cause you may use `make` if you own a `rtm` object. The returned promise will contain the reference to the made menu what is to use with `show` to become useful. 
| controller.menu.show(rtm) | promise | Of cause you may use only that to show a menu. The promises resolve will pipe the index like this: `controller.menu.make(tessels, rtm, title).then(controller.menu.show(rtm)).then(function(index) {resolve(tessels[index]);});` 
| controller.menu.clear() | promise | it simple clears the terminal and resolves the returned promise

### Usage
```
controller.menu.create(tessels)
.then(function(tessel) {
// do something with the selected Tessel
});

```

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -127,7 +127,23 @@ Tessel.get = function(opts) {
             tessels = reconciledTessels;
             // Run the heuristics to pick which Tessel to use
             return controller.runHeuristics(opts, tessels)
-              .then(logAndFinish);
+              .then(function finalSection(tessel) {
+                // If the heuristics were able to narrow it down to a single Tessel
+                if (tessel) {
+                  // Log we found it and return it to the caller
+                  return logAndFinish(tessel);
+                }
+                // If we couldn't narrow it down
+                else {
+                  // Open up an interactive menu for the user to choose
+                  return controller.menu.create(tessels, {}, 'Which Tessel did you mean to use?')
+                    // Once they choose
+                    .then(function(tessel) {
+                      // Log we found it and return it to the caller
+                      return logAndFinish(tessel);
+                    });
+                }
+              });
           });
       }
     }
@@ -688,15 +704,15 @@ controller.menu = {
           if (tessel.lanConnection) {
 
             if (!tessel.lanConnection.authorized) {
-              rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (not authorized) \n');
+              rtm.add(index + ') ' + tessel.name + ': ' + '[LAN] (not authorized) \n');
             } else {
-              rtm.add(index + ') ' + tessel.name + ': ' + tessel.lanConnection.ip + ' (authorized) \n');
+              rtm.add(index + ') ' + tessel.name + ': [LAN] \n');
             }
 
           } else if (tessel.usbConnection) {
             rtm.add(index + ') ' + tessel.name + ': [USB] \n');
           } else {
-            reject('Type of Tessel unkonwn: ', tessel);
+            reject('Type of Tessel unknown: ', tessel);
           }
           done();
         },
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -663,56 +663,74 @@ controller.menu = {
   create: function(tessels, opts, title) {
     return new Promise(function(resolve) {
       var rtm = new Menu({
-        //  width: process.stdout.columns - 4,
-        width: 55,
+        width: process.stdout.columns - 4,
+        //width: 46,
         x: 1,
         y: 2,
         bg: 'red'
       });
-
       // create menu entries
-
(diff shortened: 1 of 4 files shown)


rwaldron:
This is a great PR message :)

Student007:
:blush: thanks

johnnyman727:
@Student007 This is great! I opened a PR on your branch to integrate this service into `Tessel.get`: https://github.com/Student007/t2-cli/pull/2. Note that I changed some of the output to be more consistent with the rest of the CLI (ie removed IP addresses).

Remaining items to finish:
* [ ] Merge my PR into your PR
* [ ] Fix existing tests. Currently, running `grunt` (the test runner) will pop up the menu. We need a way for it to run w/o using the menu.
* [ ] Create new unit tests. I'm not sure the best way to do this for this kind of PR that requires interactivity. I suppose it should be as easy as writing to `process.stdin` but I haven't tried it yet. @rwaldron do you have any suggestions here?

Student007:
@johnnyman727 I will write unit tests as easy as writing API test but without chaining the hole used case. Testing `terminal-menu` itself would be a pull to the project `terminal-menu`
About fixing other tests, I will add a testing flag. I remember about asking for this and didn't got an answer so I removed the last solution for testing interactive mode.
**Is there a best practice about identifying test mode ?**

Student007:
@rwaldron Is there a best practice about identifying test mode ?

Student007:
@rwaldron [I got it](https://github.com/rwaldron/nino-io/blob/master/Gruntfile.js)

rwaldron:
> Create new unit tests. I'm not sure the best way to do this for this kind of PR that requires interactivity. I suppose it should be as easy as writing to process.stdin but I haven't tried it yet. @rwaldron do you have any suggestions here?

We don't need to test the menu itself, that's for the author of that module to do. We just need to test that the functions it calls do the right thing for both good and bad inputs.

rwaldron:
@Student007 that's correct, and it's already available in t2-cli: https://github.com/tessel/t2-cli/blob/master/test/common/bootstrap.js also, it's not necessary to change any of the actual cli code, just stub the menu's API in the tests

rwaldron:
FFS. terminal-menu doesn't export the actual Menu class, which means you can't EVER stub the methods. More importantly: **we absolutely do not want conditional code interwoven throughout the actual cli code.**

rwaldron:
After you merge @johnnyman727's PR, please also merge this: https://github.com/Student007/t2-cli/pull/3

Don't worry about new tests for now, it's not clear how/if it will even be possible. We can follow up later.

Student007:
@rwaldron does it mean I shell stop writing tests for `controller.menu` at all ?
To fix the problem `Tessel.get` is tested in `test/unit/tessel.js` needs the `controller.menu.create` to response with a single tessel.
For that reason I added **conditional code within the CLI** like that:
```
// within controller.menu.create
      if (!global.IS_TEST_ENV) {
        controller.menu.make(tessels, rtm, title)
          .then(controller.menu.show(rtm))
          .then(function(index) {
            resolve(tessels[index]);
          });
      } else {
        resolve(tessels[0]); // due to old tests what expect a single tessel returned by Tessel.get
      }
```
It isn't possible to come over that need because `controller.menu.create` is used in `Tessel.get`.
There would be a simple `cleaner` workaround:
```
// at the bottom of controller.js
if (global.IS_TEST_ENV) {
controller.menu = {
  create = function(tessels) {
     return new Promise(function(resolve) {
        resolve(tessels[0]);
     }
  });
}
```

}
```

Student007:
> After you merge @johnnyman727's PR, please also merge this: Student007#3

@rwaldron :raised_hands: magic ... :+1: understand. Will roll back my local changes.

rwaldron:
I know you put effort into writing tests and I appreciate that, but for now I'd rather we avoid conditional code paths. I'm working on writing you a "mock Menu" class, that's derived from `Menu`. We can land it in a follow up.

Student007:
@rwaldron :+1:

johnnyman727:
Looks good everyone! Will merge it now.

PR #349, opened 2015-09-28
Global --key option

Student007:
# Usage
You can verify the use of the global option is working well by
`t2 list --key ~/.ssh/id_rsa`
instead *~/.tessel/id_rsa* what is default.
The result should be (in compression to `t2 list`) your discovered
Tessels will show as non provisioned with note `(USB connect and run
`tessel provision` to authorize)`

This PR is a part of others will base on it. It should work without
conflicts.

### DIFF
--- a/lib/tessel/provision.js
+++ b/lib/tessel/provision.js
@@ -13,6 +13,16 @@ var directory = path.join(osenv.home(), '.tessel');
 var filename = 'id_rsa';
 var filepath = path.join(directory, filename);
 var remoteAuthFile = '/etc/dropbear/authorized_keys';
+var debug = require('debug')('provision');
+
+Object.defineProperty(Tessel, 'TESSEL_AUTH_KEY', {
+  get: function() {
+    return filepath;
+  },
+  set: function(value) {
+    filepath = value;
+  }
+});
 
 Object.defineProperty(Tessel, 'TESSEL_AUTH_PATH', {
   get: function() {
@@ -170,6 +180,27 @@ actions.checkAuthFileExists = function(tessel, authFile) {
   });
 };
 
+actions.setDefaultKey = function(opts) {
+
+  if (opts.key && typeof opts.key === 'string') {
+    debug('Using --key ' + opts.key + ' after checking existens');
+    filepath = opts.key;
+  } else if (opts.key && typeof opts.key !== 'string') {
+    logs.err('No valid path for private key');
+    process.exit();
+  }
+  try {
+    if (fs.statSync(filepath).isFile() &&
+      fs.statSync(filepath + '.pub').isFile()) {
+      return;
+    }
+  } catch (e) {
+    debug('Keyfiles not found', e);
+    logs.err('Could not find ' + filepath + ' and its public key (.pub)');
+    process.exit();
+  }
+};
+
 function checkIfKeyInFile(tessel, authFile, pubKey) {
   return new Promise(function(resolve, reject) {
     // Read the public keys from the auth file
@@ -256,7 +287,7 @@ function AlreadyAuthenticatedError() {
 util.inherits(AlreadyAuthenticatedError, Error);
 
 actions.AlreadyAuthenticatedError = AlreadyAuthenticatedError;
-
+module.exports.TESSEL_AUTH_KEY = filepath;
 module.exports = actions;
 
 if (global.IS_TEST_ENV) {
--- a/lib/tessel/provision.js
+++ b/lib/tessel/provision.js
@@ -183,7 +183,7 @@ actions.checkAuthFileExists = function(tessel, authFile) {
 actions.setDefaultKey = function(opts) {
 
   if (opts.key && typeof opts.key === 'string') {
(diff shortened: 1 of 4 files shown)


lib/tessel/provision.js
```diff
@@ -170,6 +180,27 @@ actions.checkAuthFileExists = function(tessel, authFile) {
   });
 };
 
+actions.setDefaultKey = function(opts) {
+
+  if (opts.key && typeof opts.key === 'string') {
+    debug('Using --key ' + opts.key + ' after checking existens');
```
rwaldron: Typo: existens => existence

lib/tessel/provision.js
```diff
@@ -256,7 +287,7 @@ function AlreadyAuthenticatedError() {
 util.inherits(AlreadyAuthenticatedError, Error);
 
 actions.AlreadyAuthenticatedError = AlreadyAuthenticatedError;
-
+module.exports.TESSEL_AUTH_KEY = filepath;
```
rwaldron: Remove this line, the data should only be exposed from the Tessel accessor property

lib/tessel/provision.js
```diff
@@ -170,6 +180,27 @@ actions.checkAuthFileExists = function(tessel, authFile) {
...
+    debug('Using --key ' + opts.key + ' after checking existens');
+    filepath = opts.key;
+  } else if (opts.key && typeof opts.key !== 'string') {
+    logs.err('No valid path for private key');
+    process.exit();
+  }
+  try {
+    if (fs.statSync(filepath).isFile() &&
+      fs.statSync(filepath + '.pub').isFile()) {
+      return;
```
johnnyman727: @Student007 should this function actually be setting the `TESSEL_AUTH_KEY`?

lib/tessel/provision.js
```diff
@@ -170,6 +180,27 @@ actions.checkAuthFileExists = function(tessel, authFile) {
   });
 };
 
+actions.setDefaultKey = function(opts) {
+
+  if (opts.key && typeof opts.key === 'string') {
+    debug('Using --key ' + opts.key + ' after checking existens');
+    filepath = opts.key;
+  } else if (opts.key && typeof opts.key !== 'string') {
+    logs.err('No valid path for private key');
```
johnnyman727: Sorry to go back and forth on this one but we should have a consistent way of handling errors. In the rest of the CLI, the error is handled by calling `reject` in a promise with an error. I know I originally suggested this function by synchronous but in an effort to keep the error handling process consistent, let's move this function to a Promise and use `resolve` and `reject`.

lib/tessel/provision.js
```diff
@@ -13,6 +13,16 @@ var directory = path.join(osenv.home(), '.tessel');
 var filename = 'id_rsa';
 var filepath = path.join(directory, filename);
 var remoteAuthFile = '/etc/dropbear/authorized_keys';
+var debug = require('debug')('provision');
+
+Object.defineProperty(Tessel, 'TESSEL_AUTH_KEY', {
```
johnnyman727: What is this set to by default?

Student007: @johnnyman727 it will stay on filepath because this is only called when active called I thought.

lib/tessel/provision.js
Student007: :scream:

lib/tessel/provision.js
Student007: it is setting the `namepath` what is used when creating an instance - also `namepath` is used in different cases within provisioning.js so using the setter wouldn't be consistent. Using the setter would be usable on the object itself out of other places in the CLI.

lib/tessel/provision.js
Student007: @rwaldron I am unsure this is what @johnnyman727 likes to get. This is a public accessible global used as base for the running CLI. The instantiated objects also can be changed by the setter - this one has another usage.

lib/tessel/provision.js
Student007: :blush:

lib/tessel/provision.js
johnnyman727: Ah okay - sorry I was confused! I understand now.

lib/tessel/provision.js
johnnyman727: I don't follow your logic. It looks like the first branch of the if statement sets the key if one was provided in the options. If a key wasn't provided, it checks to make sure the current key exists? It seems like you shouldn't be able to call this function without a new key.

lib/tessel/provision.js
johnnyman727: @rwaldron the reason we have the default filepath exported is so that it can be used as the default key on line 42 of bin/tessel-2.js (see above). I agree that it's kind of hacky though. Do you have another suggestion?

lib/tessel/provision.js
Student007: @johnnyman727 you are right - I think I forgot to review this.

Student007:
closed due to https://github.com/tessel/t2-cli/pull/363

Issue #289, opened 2015-09-08
Tessel remote commands: add "run command and give me all output" functionality


MrNice:
## The Problem:
We can execute arbitrary shell commands over USB and LAN through SSH, but the ssh interface returns an remote process, which has stdout and stderr event emitters.

This is great for complicated use cases, but it's a bit of a hassle when you just want to run a command to get its stdout output.

```js
return self.connection.exec(commands.command())
  .then(function(rp) {
    var err = '';
    var result = '';

    rp.stderr.on('data', function(d) {
      err += d.toString();
    });
    rp.stdout.on('data', function(d) {
      result += d.toString();
    });

    rp.once('close', function() {
      if (err.length) reject(new Error(err));
      // DO STUFF HERE
    });
  });
```

## A Solution
It would be nice to be able to instead write something like

```js
return self.connection.execOnce(commands.command())
  .then(function(result) {
    // DO STUFF HERE
  })
  .catch(function(err) {
    reject(err);
  });
```

johnnyman727:
Agreed! There are lots of places the code could use this format.

wprater:
I like this approach as well!

PR #351, opened 2015-09-30
Refactor many process executions

johnnyman727:
Fixes https://github.com/tessel/t2-cli/issues/289.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -461,7 +461,7 @@ controller.renameTessel = function(opts) {
       if (!opts.reset && !opts.newName) {
         reject('A new name must be provided.');
       } else {
-        if (!Tessel.isValidName(opts.newName)) {
+        if (!opts.reset && !Tessel.isValidName(opts.newName)) {
           reject('Invalid name: ' + opts.newName + '. The name must be a valid hostname string. See http://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_host_names.');
         } else {
           resolve();
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -68,44 +68,31 @@ Tessel.prototype.restartScript = function(opts) {
   var filepath = isPush ? PUSH_PATH : RUN_PATH;
 
   return new Promise(function(resolve, reject) {
-    return self.connection.exec(commands.readFile(filepath + opts.entryPoint))
-      .then(function(remoteProcess) {
-        var error = '';
+    return self.simpleExec(commands.readFile(filepath + opts.entryPoint))
+      .then(function() {
+        if (isPush) {
+          // Start the script from flash memory
+          return actions.startPushedScript(self, opts.entryPoint)
+            .then(resolve).catch(reject);
+        } else {
+          // Start the script in RAM
+          return actions.runScript(self, filepath, opts.entryPoint)
+            .then(resolve).catch(reject);
+        }
+      })
+      .catch(function(error) {
+        if (error.message.indexOf('No such file or directory') !== -1) {
+          error = '"' + opts.entryPoint + '" not found on ' + self.name;
+        }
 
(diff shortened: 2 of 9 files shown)


lib/controller.js
```diff
@@ -461,7 +461,7 @@ controller.renameTessel = function(opts) {
       if (!opts.reset && !opts.newName) {
         reject('A new name must be provided.');
       } else {
-        if (!Tessel.isValidName(opts.newName)) {
+        if (!opts.reset && !Tessel.isValidName(opts.newName)) {
```
johnnyman727: This was actually an unknown bug in `t2 rename` where the `--reset` flag would not work properly (because this if else statement rejected the command before it got to the point where it could reset)

johnnyman727:
@MrNice could you give this a code review?

MrNice:
Sure

lib/tessel/deploy.js
```diff
@@ -68,44 +68,31 @@ Tessel.prototype.restartScript = function(opts) {
...
-        remoteProcess.stderr.on('data', function(data) {
-          error += data.toString();
-        });
-
-        remoteProcess.once('close', function() {
-          if (error.length) {
-            if (error.indexOf('No such file or directory') !== -1) {
-              error = '"' + opts.entryPoint + '" not found on ' + self.name;
-            }
+    return self.simpleExec(commands.readFile(filepath + opts.entryPoint))
```
MrNice: Should we have a command "checkFileExistance"?
```bash
[ -f $FILE ] && echo "File exists" || echo "File does not exists"
```

Just to be super clear that we don't care about the contents of the file.

lib/tessel/deploy.js
```diff
@@ -68,44 +68,31 @@ Tessel.prototype.restartScript = function(opts) {
...
 
 actions.execRemoteCommand = function(tessel, command, filepath) {
-  return new Promise(function(resolve) {
-    return tessel.connection.exec(commands[command](filepath))
-      .then(function(remoteProcess) {
-        remoteProcess.once('close', resolve);
-      });
-  });
+  return tessel.simpleExec(commands[command](filepath))
+    .catch(function() {});
```
MrNice: We swallow errors here by design?

lib/tessel/name.js
```diff
@@ -127,40 +97,17 @@ Tessel.prototype.rename = function(opts) {
...
-        });
-
-        // When we receive data on stdout
-        remoteProc.stdout.on('data', function(data) {
-          // add it to the contatenated string
-          info += data.toString();
-        });
-      });
-  });
+  // var self = this;
```
MrNice: Unnecessary comment can be removed

MrNice:
@johnnyman727 I went over this change, and it looks good. Made a few comments above **^^** but they're minor things that can be quickly fixed.

lib/tessel/deploy.js
johnnyman727: @MrNice I don't follow you here. What would we use that command for?

lib/tessel/deploy.js
johnnyman727: The previous implementation did so b/c it calls a command to stop the service that starts a tessel script on boot (if you `push` a project) regardless of whether the service has been started. If it hasn't been started, it will return an error. 

However, swallowing errors is almost never a good thing so I'll revise this a bit.

Good catch!

lib/tessel/name.js
johnnyman727: Whoops!

lib/tessel/deploy.js
johnnyman727: On second thought, I'm going to make an issue that `deploy` is swallowing errors and go ahead and land this patch. It's something that's broken in master so this PR isn't necessarily responsible for fixing it.

PR #354, opened 2015-10-01
Fixes tessel erase

### LINKED ISSUE #313: t2 erase doesn't actually work
It does properly tell OpenWRT to stop running the service but it appears that OpenWRT still runs whatever is in `/app` at startup. This is exacerbated by the fact that `t2 erase` makes no effort to delete that folder.

johnnyman727:
Fixes #313.

- [x] Add tests

### DIFF
--- a/lib/tessel/commands.js
+++ b/lib/tessel/commands.js
@@ -9,7 +9,7 @@ module.exports.stopRunningScript = function() {
 };
 module.exports.disablePushedScript = function() {
   return ['/etc/init.d/tessel-app', 'disable'];
-}
+};
 module.exports.deleteFolder = function(filepath) {
   return ['rm', '-rf', filepath];
 };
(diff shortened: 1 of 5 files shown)


lib/tessel/commands.js
```diff
@@ -7,6 +7,9 @@ module.exports.scanWiFi = function() {
 module.exports.stopRunningScript = function() {
   return ['/etc/init.d/tessel-app', 'stop'];
 };
+module.exports.disablePushedScript = function() {
+  return ['/etc/init.d/tessel-app', 'disable'];
```
rwaldron: If the command is `erase`, is this misleading?

johnnyman727: @rwaldron the name is `erase` for legacy reasons. On Tessel 1, we literally just erased the code that was in flash to clear out a `tessel push`. I'm open to other suggestions. `tessel unpush`?

johnnyman727:
@rwaldron could you give this one a review?

lib/tessel/commands.js
rwaldron: We should think more on this and plan a follow up.

rwaldron:
Works Great!

Issue #355, opened 2015-10-01
inquirer might be more appropriate than terminal-menu


rwaldron:
I'm going to investigate. 

https://www.npmjs.com/package/inquirer

rwaldron:
particle-cli uses inquirer, this is what a "select from list" experience looks like, vs current t2-cli: 

![](https://i.gyazo.com/66439d476e43415dd0b9d1211d978534.png)

![](https://i.gyazo.com/19b87db3450b05aa3bb0c29207d00c9c.png)

Problems: 

- The UI change (colored background, colored text) is disorienting
- The loss of previously printed stdout is confusing (no indication of what steps were reported before the prompt)


As shown in the inquirer example above, lists can be displayed inline.


Here is the same functionality, using inquirer: 

![](https://i.gyazo.com/f93259366e9b28c480d648b80050a65e.png)

![](https://i.gyazo.com/d9b7ec2208393e7937e259775b52fdba.png)

![](https://i.gyazo.com/9e58f3c95dc077b6e49876ceaadf00e7.png)

PR #357, opened 2015-10-01
Refactor controller.menu (Replace terminal-menu with inquirer.prompt). Closes gh-355

rwaldron:
See #355

Previously: 

After refactor:

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -7,8 +7,9 @@ var sprintf = require('sprintf-js').sprintf;
 var cp = require('child_process');
 var async = require('async');
 var updates = require('./update-fetch');
+var inquirer = require('inquirer');
+var colors = require('colors');
 var controller = {};
-var Menu = require('terminal-menu');
 
 Tessel.list = function(opts) {
 
@@ -108,7 +109,7 @@ Tessel.get = function(opts) {
         reject('No Tessels Found.');
         return;
       }
-      // The name match for a given Tessel happens upon discovery, not at 
+      // The name match for a given Tessel happens upon discovery, not at
       // the completion of discovery. So if we got to this point, no Tessel
       // was found with that name
       else if (opts.name !== undefined) {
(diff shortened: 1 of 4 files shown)


lib/controller.js
```diff
@@ -136,12 +138,43 @@ Tessel.get = function(opts) {
...
+                        map[display] = i;
+
+                        return display;
+                      })
+                    },
+                    translate: function(answer) {
+                      return tessels[map[answer.selected]];
+                    }
+                  }).then(function(tessel) {
+                    if (!tessel) {
```
johnnyman727: If a single Tessel was chosen, we should close all the non-selected Tessels.

johnnyman727:
@rwaldron it looks good! I made a PR onto your PR to fix the tests and attend to my comment above about closing unused connections: https://github.com/rwaldron/t2-cli/pull/1

lib/controller.js
johnnyman727: Nvm, that's handled in `logAndClose`.

Student007:
@rwaldron @johnnyman727 :cry: my nice red menu :crying_cat_face: 

But if its better it is :+1: 

Once when I was thinking about a lot of different menu solutions I also took a look onto inquirer as a lot others like my favorite [Blessed](https://github.com/chjj/blessed#scrollablebox-from-box).
Also I thought about using nodes readline as you can see in this [example](https://nodejs.org/api/readline.html#readline_example_tiny_cli).

**The reason I decided to go with `terminal-menu` in version `2.x` had been to create a cooperate identity matching solution (what feels like to promise `t2-cli` to be a user friendly tool)**. Blessed would be to much but same as [terminal-kit](https://github.com/cronvel/terminal-kit). All other solutions I took a look onto either were beta or experimental.

There is also a fork of `terminal-menu` what provides selection and switches (see: [terminal-menu-program](https://github.com/Pomax/terminal-menu-program)). If the goal is to use selections this would be an option if the need of brand conform cooperate identity is needed.

About the problem of look and feel like @rwaldron shows up, it could be simple be solved by calling `controller.menu.clear().then(function(){// go forwards})`

But as I said: If majority is for replacing terminal-menu I didn't have a problem with this - the best solution should be in - not the hardest work :smile: 

Maybe @Frijol also would put her two cents in (as a marketing professional).

johnnyman727:
@Student007 don't be :crying_cat_face:! Be :joy_cat:! Every contribution builds on the work of the previous. Without your help, it would have been much more work to improve on the menu experience. It's important to get used to your code being overwritten because, in a successful project, that will happen all the time.

Issue #358, opened 2015-10-02
provision needs to prefer usb connections


johnnyman727:
Currently, the `provision` command doesn't specify that `Tessel.get` *must* return a USB connection. This can be easily fixed by changing the options dict to include `usb = true`.

johnnyman727:
I'm going to close this for now b/c I can't reproduce this. Somehow, `tessel.provisionTessel` always gets a Tessel with a `usbConnection` if it's possible. The strange thing is that I have no idea how this is happening...

johnnyman727:
Not true! My Tessel was just slow to find the LAN connection for some reason. PR coming up...

Issue #238, opened 2015-07-25
--name flag doesn't prefer LAN connections


johnnyman727:
Because USB will almost always be detected before a LAN Tessel, the CLI will choose the USB Tessel and stop searching for LAN connections as soon as the USB connection enumerates. That means if you have  Tessel connected via USB and LAN and you pass the --name flag, you will only be able to use the USB connection.

Relevant code: https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L109-L115

PR #360, opened 2015-10-02
Uses LAN and USB flags properly

johnnyman727:
Turns out this was the root cause of #358 and #238. Probably more issues too, they just weren't reported.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -18,15 +18,19 @@ Tessel.list = function(opts) {
 
     // Keep a list of all the Tessels we discovered
     var foundTessels = [];
+
+    // Options for  Tessel discovery
+    var seekerOpts = {
+      timeout: opts.timeout * 1000,
+      usb: opts.usb,
+      lan: opts.lan
+    };
+
     // Start looking for Tessels
-    var seeker = new discover.TesselSeeker().start(opts.timeout * 1000);
+    var seeker = new discover.TesselSeeker().start(seekerOpts);
 
     // When a Tessel is found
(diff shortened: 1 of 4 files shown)


lib/controller.js
```diff
@@ -68,7 +72,7 @@ Tessel.list = function(opts) {
             // Report that selected Tessel to the user
             logs.info('Multiple Tessels found.');
             if (tessel) {
-              logs.info('Will default to', tessel.name, '.');
+              logs.info('Will default to', tessel.name + '.');
```
johnnyman727: This was bugging me because it was print out as: `Will default to Frank .` with a space before the period.

LinusU: Now you use both `,` and `+` on the same line

johnnyman727: Yeah, perhaps `sprintf` is the reasonable way to do this?

LinusU: This should work and looks nice:

```javascript
logs.info('Will default to %s.', tessel.name);
```

It uses `util.format` which is a builtin in Node that does this. That works with `console.log` as well :+1:

johnnyman727:
@LinusU since you already started, do you want to give it a full review?

LinusU:
Already on it ;)

LinusU:
Code looks good, tests pass locally and on CI.

The only thing is the one comment I made:

```diff
-              logs.info('Will default to', tessel.name + '.');
+              logs.info('Will default to %s.', tessel.name);
```

LGTM :+1:

johnnyman727:
Awesome, thanks a million! Will make that change and merge.

PR #362, opened 2015-10-04
Adds base wifi command and tests

### LINKED ISSUE #199: "t2 wifi" alone does nothing
```
➜  t2-cli git:(master) t2 wifi
➜
```

I assume this is when there is no wifi network connected.

johnnyman727:
Fixes #199.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -12,18 +12,20 @@ function closeSuccessfulCommand() {
   process.exit(0);
 }
 
-function closeFailedCommand(err) {
-  // If the returned value is an error
+// Allow options to be partially applied
+function closeFailedCommand(opts, err) {
+  if (!err) {
+    err = opts;
+    opts = {};
+  }
   if (err instanceof Error) {
-    // Throw it
     throw err;
+  } else {
+    // Print a stern warning by default
+    opts.type = opts.type || 'warn';
+    logs[opts.type](err);
   }
-  // Otherwise
-  else {
-    // Print a stern warning
-    logs.warn(err);
-  }
-  // Exit with non-zero code
+  // NOTE: Exit code is non-zero
   process.exit(1);
 }
 
@@ -165,11 +167,14 @@ parser.command('init')
 
 makeCommand('wifi')
   .callback(function(opts) {
-    //TODO: Refactor switch case into controller.wifi
+    // TODO: Refactor switch case into controller.wifi
     if (opts.list) {
       callControllerWith('printAvailableNetworks', opts);
     } else if (opts.ssid && opts.password) {
       callControllerWith('connectToNetwork', opts);
+    }
+    else {
+      callControllerWith('getWifiInfo', opts);
     }
   })
   .option('list', {
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -7,28 +7,6 @@ var key = require('../lib/key');
 var init = require('../lib/init');
 var logs = require('../lib/logs');
 
-
-function closeSuccessfulCommand() {
-  process.exit(0);
-}
-
-// Allow options to be partially applied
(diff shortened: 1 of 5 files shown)


bin/tessel-2.js
```diff
@@ -252,6 +234,28 @@ module.exports = function(args) {
   parser.parse(args);
 };
 
+module.exports.closeSuccessfulCommand = function() {
+  process.exit(0);
+};
+
+// Allow options to be partially applied
+module.exports.closeFailedCommand = function(opts, err) {
```
johnnyman727: It's kind of ugly but I had to break `closeSuccessfulCommand` and `closeFailedCommand` into the module exports (and call them as `module.exports.X` elsewhere) to use them properly in tests. Other suggestions welcome.

test/unit/bin-tessel-2.js
```diff
@@ -87,12 +88,11 @@ exports['Tessel (cli: restart)'] = {
   noError: function(test) {
     test.expect(1);
 
-    cli(['restart']);
+    cli(['restart', '--type=ram']);
```
johnnyman727: For some reason, the `restart` command doesn't get the default `type` property assigned to it by `nomnom` in the unit tests. The default param shows up properly w/o tests.

test/unit/bin-tessel-2.js
```diff
@@ -239,9 +238,12 @@ exports['Tessel (cli: wifi)'] = {
   setUp: function(done) {
     this.sandbox = sinon.sandbox.create();
     this.warn = this.sandbox.stub(logs, 'warn');
+    this.info = this.sandbox.stub(logs, 'info');
     this.printAvailableNetworks = this.sandbox.stub(controller, 'printAvailableNetworks').returns(Promise.resolve());
     this.connectToNetwork = this.sandbox.stub(controller, 'connectToNetwork').returns(Promise.resolve());
-    this.processExit = this.sandbox.stub(process, 'exit');
```
johnnyman727: For some reason, `process.exit` wasn't being stubbed over properly here. The actual `process.exit` would get called and break the tests. I opted to use `closeSuccessfulCommand` insteaad.

johnnyman727: @rwaldron any idea why this might be the case?

johnnyman727:
@MrNice could you review?

MrNice:
@johnnyman727 I'll get it early tmw morning, sorry. Just finished a reversi app that I was assigned for a job process. https://arcane-lake-3833.herokuapp.com/ Probably the second largest job application piece I've had to do.

johnnyman727:
@MrNice I don't really know how reversi works but the premise seems to be that you can never beat the computer.

MrNice:
@johnnyman727 It looks good to me, but I'm not 100% clear on what the errors were. There's a few ugly things like the double `catch` that should be looked into.

Is there a Wiki Page on best practices for writing tests for the CLI? This PR opened my eyes to a few of the oddities therein.

johnnyman727:
@MrNice The root cause appears to be that `process.exit` wasn't properly being stubbed out so the actual `process.exit` was being called which killed the test. I'm not sure why that is the case.

Where is the double `catch` that you are referring to?

We don't have documentation up for CLI tests yet. I've mostly been going off of [`nodeunit`](https://github.com/caolan/nodeunit) and [`sinon`](http://sinonjs.org/) documentation.

test/unit/bin-tessel-2.js
```diff
@@ -285,7 +289,7 @@ exports['Tessel (cli: wifi)'] = {
     reject.catch(function() {
```
MrNice: @johnnyman727 Sorry! I was referring to these double catches over here, they are outside of the scope of the PR

johnnyman727: :+1:

PR #363, opened 2015-10-05
Global --key option (patch 1)

Student007:
Fixes useless conditional branch of if statement.

@johnnyman727 
I am wondering about where I answered with :scream: about recover `actions.setDefaultKey` to be asynchronous as I did first time.
Did you erase this part or am I to stupid to find it at the moment (I have a cold).

I can create an additional patch to rework this asynchronous. Should I do this based on the current master or are you able to merge them quickly (due to msg about conflicts) ? 

> You can verify the use of the global option is working well by
t2 list --key ~/.ssh/id_rsa
instead ~/.tessel/id_rsa what is default.
The result should be (in compression to t2 list) your discovered
Tessels will show as non provisioned with note (USB connect and run tessel provisionto authorize)

This PR is a part of others will base on it.
I need this PR merged for `t2 root` PR.

### DIFF
--- a/lib/tessel/provision.js
+++ b/lib/tessel/provision.js
@@ -13,6 +13,16 @@ var directory = path.join(osenv.home(), '.tessel');
 var filename = 'id_rsa';
 var filepath = path.join(directory, filename);
 var remoteAuthFile = '/etc/dropbear/authorized_keys';
+var debug = require('debug')('provision');
+
+Object.defineProperty(Tessel, 'TESSEL_AUTH_KEY', {
+  get: function() {
+    return filepath;
+  },
+  set: function(value) {
+    filepath = value;
+  }
+});
 
 Object.defineProperty(Tessel, 'TESSEL_AUTH_PATH', {
   get: function() {
@@ -170,6 +180,27 @@ actions.checkAuthFileExists = function(tessel, authFile) {
   });
 };
 
+actions.setDefaultKey = function(opts) {
+
+  if (opts.key && typeof opts.key === 'string') {
+    debug('Using --key ' + opts.key + ' after checking existens');
+    filepath = opts.key;
+  } else if (opts.key && typeof opts.key !== 'string') {
+    logs.err('No valid path for private key');
+    process.exit();
+  }
+  try {
+    if (fs.statSync(filepath).isFile() &&
+      fs.statSync(filepath + '.pub').isFile()) {
+      return;
+    }
+  } catch (e) {
+    debug('Keyfiles not found', e);
+    logs.err('Could not find ' + filepath + ' and its public key (.pub)');
+    process.exit();
+  }
+};
+
 function checkIfKeyInFile(tessel, authFile, pubKey) {
   return new Promise(function(resolve, reject) {
     // Read the public keys from the auth file
@@ -256,7 +287,7 @@ function AlreadyAuthenticatedError() {
 util.inherits(AlreadyAuthenticatedError, Error);
 
 actions.AlreadyAuthenticatedError = AlreadyAuthenticatedError;
-
+module.exports.TESSEL_AUTH_KEY = filepath;
 module.exports = actions;
 
 if (global.IS_TEST_ENV) {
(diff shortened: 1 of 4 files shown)


Student007:
@johnnyman727 I will `rebase` and open a new PR after clarification about asynchronously or not. Async would change `Tessel.get` and `Tessel.list` as shown earlier. I would use the old working parts of code (in the case you are going forward to like the asynchronous version).

johnnyman727:
@Student007 I think this function should be async to be consistent with the rest of the code base.

Student007:
@johnnyman727 I have big trouble to fix tests when using asynchronous `setDefaultKey` within `Tessel.get` or `Tessel.list`.

The key option itself works on terminal. But the tests are ending within test/unit/controller.js with `Fatal error: Not running`

Firstly I tried to fixed tests by adding `authKey` to the options next to where each `timeout` is located - but it didn't fix all problems.

Also playing around with longer timeout didn't help.

Thought it could be the latency caused by checking `isFile()` within provision.js `setDefaultKey`. (of cause this was not causing the problems - it is only a file is present check ...)

I assumed also it is depending on the inner resolve of `provision.setDefaultKey` what ends the unit test case. For that reason I created a stub for this method. But returning causes in `Cannot read property 'then' of undefined` of cause. But returning `Promize.resolve()` doesn't really make sense. It causes same behavior as not stubbing.

Any idea ? Or maybe @rwaldron has and idea ? I am out of ideas ... and need a break.

My last idea while reading my comment:
```
Tessel.get({
        timeout: 0.01
      })
      .then(function() {
// handling the first resolve
}).then(function() {
// all tests ...
test.done();
});
```

johnnyman727:
Will check this out later this week.

Student007:
@johnnyman727 would be really cool :smile:

Student007:
@johnnyman727 I am blocked with `t2 root` due to #363 ...

I noticed this while writing tests and now the weekend is over :cry: - I hoped to finish `t2 root` today.

Hope you can help me with #363 before next weekend :smile: - Elsewhere we should talk about to keep it synchronous - so tests will not explode.

lib/tessel/provision.js
```diff
@@ -256,7 +287,7 @@ function AlreadyAuthenticatedError() {
 util.inherits(AlreadyAuthenticatedError, Error);
 
 actions.AlreadyAuthenticatedError = AlreadyAuthenticatedError;
-
+module.exports.TESSEL_AUTH_KEY = filepath;
```
johnnyman727: @Student007 if the filepath variable is ever set by another part of the program, it seems like this export will store an old variable. Is that understanding correct?

johnnyman727: Also, you will need to place this line below the next because `module.exports = actions` overwrites this variable.

johnnyman727:
@Student007 you'll need to rebase and update the `TESSEL_AUTH_KEY` variable with the new `authKey` var.

Student007:
@johnnyman727 no this isn't the problem. But I think due to prevent further confusions here is a commit you can run tests and see grunt exploding tests :boom: 
https://github.com/student007/t2-cli/tree/key

The problem is only I didn't get tests running. The features work fine - after installation you can try:

`t2 list --key ~/.ssh/id_rsa` to see a message about non provisioned tessels instead provisioned shown by using `t2 list` without `--key`.
Same for `t2 version --key ~/.ssh/id_rsa` or others use `Tessel.get` ...

When I did the same with no promise surround the logic of `Tessel.get` or `Tessel.list` it works with tests for weeks... 
I am still blocked about `t2 root` about this at the moment.

johnnyman727:
Understood! Well, since the Milestone for this week is spec completion and `t2 root` is part of the spec we'll try to pump this out this week!

Student007:
@johnnyman727 cool - will have time tomorrow!

Student007:
@johnnyman727 looking at the clock: I will not be able to finish it that evening. Next earliest time frame will be Thursday.

johnnyman727:
Closing in favor of #432. I could try to merge it into this branch but it's proving difficult.

PR #365, opened 2015-10-05
Remote: abstract receiving from remote process; to replace for all remote process error/data handlers

rwaldron:
As t2-cli has evolved, we've built up some technical debt w/r to handling remote process io. Specifically, the code that handled receiving errors, data and closing was repeated in several places. This PR abstracts the operation into a Promise returning method of the current Tessel instance.

### DIFF
--- a/lib/tessel/tessel.js
+++ b/lib/tessel/tessel.js
@@ -69,40 +69,38 @@ Object.defineProperty(Tessel.prototype, 'connection', {
   }
 });
 
+Tessel.prototype.receive = function(remote) {
+  var error = '';
+  var received = new Buffer(0);
+
+  remote.stderr.on('data', function(buffer) {
+    error += buffer.toString();
+  });
+
+  remote.stdout.on('data', function(buffer) {
+    received = Buffer.concat([received, buffer]);
+  });
+
+  return new Promise(function(resolve, reject) {
+    remote.once('close', function() {
+      if (error) {
+        return reject(new Error(error));
+      } else {
+        return resolve(received);
+      }
+    });
+  });
+};
+
 Tessel.prototype.simpleExec = function(command) {
   var self = this;
   return new Promise(function(resolve, reject) {
     // Stop processes and delete everything in the folder
     return self.connection.exec(command)
       .then(function(remoteProcess) {
-
-        // Buffer to store incoming error data
-        var errBuf = new Buffer(0);
-        // If we receive error data
-        remoteProcess.stderr.on('data', function(e) {
-          // Concatenate the data
-          errBuf = Buffer.concat([errBuf, e]);
-        });
-
-        // Buffer to store incoming stdout data
-        var dataBuf = new Buffer(0);
-        // If we receive stdout data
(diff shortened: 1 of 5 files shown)


lib/tessel/tessel.js
```diff
@@ -69,40 +69,38 @@ Object.defineProperty(Tessel.prototype, 'connection', {
...
+    remote.once('close', function() {
+      if (error) {
+        return reject(new Error(error));
+      } else {
+        return resolve(received);
+      }
+    });
+  });
+};
+
```
rwaldron: @johnnyman727 start here.

lib/tessel/tessel.js
```diff
@@ -69,40 +69,38 @@ Object.defineProperty(Tessel.prototype, 'connection', {
   }
 });
 
+Tessel.prototype.forward = function(remote) {
```
rwaldron: Todo: rename to `receive`

rwaldron:
I'll get those failures fixed in the morning

johnnyman727:
@rwaldron I don't quite understand the purpose of this PR. It seems like it abstracts out part of `tessel.simpleExec` so that it's less encapsulated for instances where `simpleExec` was too abstract?

rwaldron:
>  I don't quite understand the purpose of this PR. It seems like it abstracts out part of tessel.simpleExec so that it's less encapsulated for instances where simpleExec was too abstract

No, `simpleExec` resolves with `remoteProcess`, what is then done with that `remoteProcess` is arbitrary, but always involved: 

1. listening to `remoteProcess.stderr` for "data" events
  - capturing any error data as it's received
2. listening to `remoteProcess.stdout` for "data" events
  - capturing any data as it's received
3. sometimes writing to `remoteProcess.stdin`
4. listening to `remoteProcess` for "close" events
  - If any error data was captured, `reject(...)`
  - Otherwise, `resolve(...)`, sometimes with the data captured in step 2.
  - Sometimes log some message

There were 5 places were either all 4 steps were taken, or some combination eg. 1 & 4 or 1, 3, & 4 or just 4. All 5 places had slightly different variable names and slightly different comments explaining the operation—but all were doing either exactly (or very nearly) the same thing. Abstracting the intersection semantics into a single operation (steps 1, 2, 4), that returns a promise which is either resolved with any received data or rejected with any error, provides a sound consistency across all occurrences. This _cannot_ be rolled into `simpleExec`, because of step 3, where sometimes `remoteProcess.stdin` needs to be written to.

All of the affected operations were smoke tested many times, with different T2 devices.

johnnyman727:
@rwaldron so, do I understand it correctly that `receive` is essentially "let me create the process and do something to it but just let me know when it has completed" whereas "simpleExec" is equivalent to "just run the command and tell me when it completes - I'm not going to write to the process"?

johnnyman727:
I think it was the name `receive` that through me off but now I get it. It's when you're done with a process and want to receive data and close. 

In any case, the PR looks good to me so feel free to merge!

Issue #366, opened 2015-10-06
Error message when renaming unauthorized Tessels is confusing


Student007:
will create a pull request fixes this next minutes.

johnnyman727:
Fixed with https://github.com/tessel/t2-cli/pull/367.

PR #367, opened 2015-10-06
Rename check authorized

Student007:
fixes https://github.com/tessel/t2-cli/issues/366

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -533,7 +533,12 @@ controller.renameTessel = function(opts) {
     })
     .then(function executeRename() {
       return controller.standardTesselCommand(opts, function(tessel) {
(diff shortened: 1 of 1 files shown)


Student007:
@rwaldron ah .. see - you also have failing tests. Is there something broken in the master ?

johnnyman727:
@Student007 we still have one weird elusive bug that sometimes causes the entire test suite to fail. I've restarted the tests on Travis.

johnnyman727:
@Student007 I'm glad you caught this. The proper way to fix it would be the way I've outlined in [the issue you created](https://github.com/tessel/t2-cli/issues/368). I'll close this patch in favor of one that fixes that overarching issue.

Issue #368, opened 2015-10-06
Should `Tessel.get` have a conditional auth filter ?


Student007:
Some internal commands would like to show only authorized Tessels if multiple devices are discovered and the menu comes up. At the moment it will show also not authorized Tessels.

johnnyman727:
@Student007 great find! I didn't realize that if you have only an unauthorized Tessel, it will be the one returned by `Tessel.get`. 

We should make `authorized` be an optional [parameter passed to the seeker](https://github.com/tessel/t2-cli/blob/8848a73c1c0b0d27c2aa291e8390c2e0e4200b36/lib/controller.js#L113-L120). Then check that flag [in the seeker after attempting to open](https://github.com/tessel/t2-cli/blob/8848a73c1c0b0d27c2aa291e8390c2e0e4200b36/lib/discover.js#L52). If the Tessel is not authorized, do not emit it.

@Student007 are you up for this patch?

Student007:
@johnnyman727 yes I can do this.

PR #369, opened 2015-10-06
Tessel get seeker option

Student007:
`Tessel.get` and `Tessel.list` accept additional option
`opts.authorized` for filtering seekers result.

fixes https://github.com/tessel/t2-cli/issues/368

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -32,6 +32,9 @@ Tessel.list = function(opts) {
 
     // Start looking for Tessels
     var seeker = new discover.TesselSeeker().start(seekerOpts);
+    var noTessels = opts.authorized ?
+      'No Authorized Tessels Found.' :
+      'No Tessels Found.';
 
     // When a Tessel is found
     seeker.on('tessel', function displayResults(tessel) {
@@ -55,7 +58,7 @@ Tessel.list = function(opts) {
       // If there were no Tessels found
       if (foundTessels.length === 0) {
         // Report the sadness
-        reject('No Tessels Found');
+        reject(noTessels);
       } else if (foundTessels.length === 1) {
         // Close all opened connections and resolve
         controller.closeTesselConnections(foundTessels)
@@ -117,17 +120,22 @@ Tessel.get = function(opts) {
       lan: opts.lan,
       authorized: true
     };
+
     if (opts.authorized !== undefined) {
       seekerOpts.authorized = opts.authorized;
     }
+
     // Create a seeker object and start detecting any Tessels
     var seeker = new discover.TesselSeeker().start(seekerOpts);
+    var noTessels = seekerOpts.authorized ?
+      'No Authorized Tessels Found.' :
+      'No Tessels Found.';
 
     function searchComplete() {
       // If we found no Tessels
       if (tessels.length === 0) {
         // Report it
-        reject('No Tessels Found.');
+        reject(noTessels);
         return;
       }
       // The name match for a given Tessel happens upon discovery, not at
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -115,9 +115,11 @@ Tessel.get = function(opts) {
       timeout: (opts.timeout || 2) * 1000,
       usb: opts.usb,
       lan: opts.lan,
-      authorized: opts.authorized || true
+      authorized: true
     };
-
+    if (opts.authorized !== undefined) {
+      seekerOpts.authorized = opts.authorized;
+    }
     // Create a seeker object and start detecting any Tessels
     var seeker = new discover.TesselSeeker().start(seekerOpts);
 
@@ -490,6 +492,7 @@ controller.provisionTessel = function(opts) {
     .then(function executeProvision() {
       // We should only be using a USB connection
       opts.usb = true;
+      opts.authorized = false;
       // Fetch a Tessel
       return controller.standardTesselCommand(opts, function(tessel) {
         // Provision Tessel with SSH keys
@@ -499,6 +502,7 @@ controller.provisionTessel = function(opts) {
 };
 
 controller.deployScript = function(opts) {
+  opts.authorized = true;
   return controller.standardTesselCommand(opts, function(tessel) {
     // Deploy a path to Tessel
     return tessel.deployScript(opts);
@@ -506,6 +510,7 @@ controller.deployScript = function(opts) {
 };
 
 controller.restartScript = function(opts) {
+  opts.authorized = true;
   return controller.standardTesselCommand(opts, function(tessel) {
     // Tell Tessel to restart an existing script
     return tessel.restartScript(opts);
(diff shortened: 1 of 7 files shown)


lib/controller.js
```diff
@@ -113,7 +114,8 @@ Tessel.get = function(opts) {
     var seekerOpts = {
       timeout: (opts.timeout || 2) * 1000,
       usb: opts.usb,
-      lan: opts.lan
+      lan: opts.lan,
+      authorized: opts.authorized || true
```
rwaldron: If `opts.authorized` is explicitly set to `false`, this expression will evaluate to `true`, ie. `false || 1; // 1`

Student007: @rwaldron :scream: ... you are right - any idea ?

Student007: maybe 
```
authorized: opts.authorized || "yes" // or "no"
```

johnnyman727: I recommend passing through the option as is. Then, inside the [seeker.start](https://github.com/tessel/t2-cli/blob/master/lib/discover.js#L26-L47) function, you can sanitize the option. I don't think you need should set it to true by default.

lib/controller.js
```diff
@@ -26,7 +26,8 @@ Tessel.list = function(opts) {
     var seekerOpts = {
       timeout: opts.timeout * 1000,
       usb: opts.usb,
-      lan: opts.lan
+      lan: opts.lan,
+      authorized: undefined
```
johnnyman727: I recommend setting it to `false` explicitly.

lib/discover.js
```diff
@@ -50,6 +50,18 @@ TesselSeeker.prototype.start = function(opts) {
     var tessel = new Tessel(conn);
     var opening = tessel.connection.open()
       .then(function() {
+        if (tessel.lanConnection && opts.authorized !== undefined) {
```
johnnyman727: I think what we should do here is just [change the Tessel initialization](https://github.com/tessel/t2-cli/blob/master/lib/tessel/tessel.js#L5) to have `authorized = true` by default if it is a USB Connection. That way this can just be a simple if statement:

```
if (opts.authorized && !tessel.connection.authorized) {
  return;
}
```

Student007: @johnnyman727 will do it tomorrow :sleeping:

lib/controller.js
rwaldron:
```js
authorized: Boolean(opts.authorized),
```

Will make... 

```js

Boolean(true);
true

Boolean(1);
true

Boolean(false);
false

Boolean(undefined);
false

Boolean(null);
false

Boolean(0);
false
```

lib/discover.js
rwaldron: > I think what we should do here is just change the Tessel initialization to have authorized = true

+1—this is semantically sound.

lib/controller.js
Student007: @rwaldron thanks for this idea - never used `Boolean()` before.
I am used to have three (3) states to work with: 
- undefined (by `x === "undefined"`)
- true (everything with a value what is neither "undefined" nor false `if (x)`)
- false (if it is false: `if (!x)` what only should match if `x === false` means it will **NOT** match if `x === 0`)

The stupid expression `opts.authorized || true` is really a bad case because the `||` is triggered by `"undefined"` and `false`

lib/discover.js
Student007: :+1:  - did this.

Student007:
@johnnyman727 @rwaldron 
Changed default value and also added a reasonable `opts.authorized` to
each command what calls `Tessel.get` or
`controller.standardTesselCommand`` - also the CASE of Tessel.js
`connection.connectionType === USB` defaults to  `connection.authorized
=== true`

rwaldron:
I just did smoke testing with an unauthorized T2 and this nicely prevents the `Unhandled rejection Error: Not connected` issue. 

Once [this is merged](https://github.com/Student007/t2-cli/pull/4), then I'm +1 on this PR.

Student007:
@rwaldron @johnnyman727 my local tests pass - don't know why Travis-CI is angry ... :sleeping: now

lib/controller.js
```diff
@@ -113,17 +117,25 @@ Tessel.get = function(opts) {
...
 
+    if (opts.authorized !== undefined) {
+      seekerOpts.authorized = opts.authorized;
+    }
+
     // Create a seeker object and start detecting any Tessels
     var seeker = new discover.TesselSeeker().start(seekerOpts);
+    var noTessels = seekerOpts.authorized ?
+      'No Authorized Tessels Found.' :
+      'No Tessels Found.';
```
johnnyman727: @Student007 can we try to make this cleaner? It's not good that this is copied and pasted in `Tessel.get` and `Tessel.list`.

lib/controller.js
```diff
@@ -497,6 +510,7 @@ controller.provisionTessel = function(opts) {
 };
 
 controller.deployScript = function(opts) {
+  opts.authorized = true;
```
johnnyman727: It's not great that we have this `opts.authorized = true` all over the place. I understand that you didn't want to make it an option for each command b/c the user doesn't need access to it. But perhaps we can place this line in [`callControllWith`](https://github.com/tessel/t2-cli/blob/8848a73c1c0b0d27c2aa291e8390c2e0e4200b36/bin/tessel-2.js#L34)?

lib/tessel/tessel.js
```diff
@@ -32,6 +32,7 @@ function Tessel(connection) {
     // Set the connection var so we have an abstract interface to relay comms
     switch (connection.connectionType) {
       case 'USB':
+        connection.authorized = true;
```
johnnyman727: Nice!

lib/controller.js
rwaldron: That was my patch ;)

I figured we could do a follow up, but might as well just store them in vars now

lib/controller.js
rwaldron: I did a test earlier where I put it in put it in standardTesselCommand, but there's one place where it's a `false`, and I had other stuff to do so I didn't press it

Student007: this was the reason I also didn't put it in `standardTesselCommand`

Student007: the static way I did this at the moment is done with the purpose it will be changed by other PRs after deeper thinking about each command. For example - @johnnyman727 is right - some could be an option within the CLI `bin/tessel-2.js`

lib/controller.js
Student007: @rwaldron @johnnyman727 is in `Seeker.start` now - usable like this:
```
var seeker = new discover.TesselSeeker().start(seekerOpts);
    var noTessels = opts.authorized ?
      seeker.msg.noAuth :
      seeker.msg.auth;
```
Tests are fixed by adding `msg` to stub of `Seeker.start`

Student007:
:tada: Travis-CI error fixed also - before :tea: + :egg: (breakfast) :yum:

rwaldron:
@Student007 great work!

rwaldron:
Landed

Student007:
:notes: :dancers: :notes: 
great teamwork!
:bow:

PR #370, opened 2015-10-08
Update messages to use "t2"

dtex:


rwaldron:
Thanks!

rwaldron:
There is a goal to merge this with `tessel` cli at some point in the future, but until then it makes sense for users to see the right messages.

PR #374, opened 2015-10-14
Refuse to update over USB because it's definitely broken right now

johnnyman727:


kevinmehall:
If you do want to update over USB and don't care about preserving the overlay, you can use the [factory test flash script](https://github.com/tessel/t2-test-rig-sw/blob/master/client/tests/flash.py) to make the samd21 forcibly overwrite the SPI flash -- CLI should at some point offer this as a "factory reset" for if if you break openwrt.

rwaldron:
Works for me.

PR #376, opened 2015-10-16
If the user explicitly specified the usb flag, then we need to ignore authorization filtering.

rwaldron:
If a user has a Tessel 2 connected via USB and runs a command with the "usb" flag, the "authorized" filter should not interfere with surfacing a USB connected device.

This is blocking us from working in a LAN-less environment.

### DIFF
--- a/lib/discover.js
+++ b/lib/discover.js
@@ -23,6 +23,12 @@ TesselSeeker.prototype.start = function(opts) {
   // An array of pending open connections
   var pendingOpen = [];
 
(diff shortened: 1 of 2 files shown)


johnnyman727:
This is a much more elegant fix than what I had in my head :)

Works as expected. Feel free to merge.

johnnyman727:
Actually, going to break protocol and merge this myself so we can make use of it at the JQuery event. :fearful:

PR #380, opened 2015-10-18
[WIP] after update, set name of tessel to old name; add a few unit tests ar…

aricearice:
…ound that

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -747,18 +747,28 @@ controller.update = function(opts) {
 };
 
 controller.updateTesselWithVersion = function(force, tessel, currentVersion, build) {
+  function logCompletion() {
+    if (!force) {
+      logs.info('Updated', tessel.name, 'from ', currentVersion, ' to ', build.version);
+    } else {
+      logs.info('Force updated', tessel.name, 'to version', build.version);
+    }
+  }
 
   // Fetch the requested build
   return updates.fetchBuild(build)
     .then(function startUpdate(image) {
       // Update Tessel with it
       return tessel.update(image)
-        // Log that the update completed
-        .then(function logCompletion() {
-          if (!force) {
-            logs.info('Updated', tessel.name, 'from ', currentVersion, ' to ', build.version);
+        .then(function retainName() {
+          if (tessel.name) {
+            // If Tessel originally had a name before the update, set Tessel name to its original name
+            tessel.setName(tessel.name)
+              // Log that the update completed
+              .then(logCompletion);
           } else {
-            logs.info('Force updated', tessel.name, 'to version', build.version);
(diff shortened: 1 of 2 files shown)


johnnyman727:
@aricearice I see that this is still a work in progress but I just wanted to say that your implementation looks great so far!

aricearice:
turns out my implementation doesn't fix the problem. The log statements in controller.updateTesselWithVersion are never even printed. 

```
t2 update --force                              ~/src/t2-cli
INFO Connected to psyduck over LAN
INFO Updating psyduck to latest version (0.0.4)...
INFO Download complete!
INFO Updating OpenWRT (1/2)
INFO Transferring image of size 19.92 MB
INFO Starting OpenWRT update. Please do not remove power from Tessel.
INFO Updating firwmare (2/2)
INFO Firmware update complete!
INFO Tessel is rebooting. Please wait about 30 seconds before using Tessel.
```

lib/controller.js
```diff
@@ -747,18 +747,28 @@ controller.update = function(opts) {
 };
 
 controller.updateTesselWithVersion = function(force, tessel, currentVersion, build) {
+  function logCompletion() {
+    if (!force) {
+      logs.info('Updated', tessel.name, 'from ', currentVersion, ' to ', build.version);
+    } else {
+      logs.info('Force updated', tessel.name, 'to version', build.version);
+    }
+  }
```
rwaldron: All this needs to work correctly is a line here that captures the present value of `tessel.name`, to be used further down in your call to `tessel.setName(...)`

johnnyman727: @rwaldron that's what @aricearice is doing here. Are you suggested she explicitly assign `tessel.name` to a new variable prior to initiating restart in case `tessel.name` gets set automatically somehow in the future?

The real hard part of this PR is that the Tessel continues to boot for another 30 seconds or so after the update and there is currently no easy way to figure out when it finished (so you can set the name back).

rwaldron:
> The real hard part of this PR is that the Tessel continues to boot for another 30 seconds or so after the update and there is currently no easy way to figure out when it finished (so you can set the name back).

Ah, yikes.

johnnyman727:
Just a note, that I [discovered why] the hostname was being reset. It looks like the same config file we created to set a unique hostname on the first boot is also being run after every system upgrade.

johnnyman727:
@aricearice it looks like this solution will not be optimal (partially because it would not be fun for the command to wait until the system reboots to set the name). Instead I've outlined the proper way to set the name [here](https://github.com/tessel/openwrt-tessel/issues/19#issuecomment-156745814). I'm happy to help if you're interested in tackling that route instead! It's a little more Linux-heavy than this attempt.

PR #381, opened 2015-10-18
Colorize `t2 list` by USB, LAN, PROVISIONED and NOT AUTHORIZED

Student007:
As you could see at summit a lot of listed Tessel 2 will make it difficult to find your own.
t2listcolored3
So I made a difference from not authorized Tessels when no `~/.tessel/id_rsa` is present. This will help the customer to find own Tessel-Name faster.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -40,6 +40,8 @@ Tessel.list = function(opts) {
       responses.noAuth :
       responses.auth;
 
+    var provisioned = Tessel.isProvisioned();
+
     // When a Tessel is found
     seeker.on('tessel', function displayResults(tessel) {
 
@@ -51,10 +53,15 @@ Tessel.list = function(opts) {
       // Add a note if the user isn't authorized to use it yet
       if (tessel.connection.connectionType === 'LAN' && !tessel.connection.authorized) {
         note = '(USB connect and run `t2 provision` to authorize)';
+        logs.basic(sprintf('\t%s\t%s\t%s'.red, tessel.name, tessel.connection.connectionType, note));
+      } else if (tessel.connection.connectionType === 'USB' && !provisioned) {
+        // if the user never used the t2 provision at this place
+        note = '(USB connect and run `t2 provision` to authorize)';
+        logs.basic(sprintf('\t%s\t%s\t%s'.blue, tessel.name, tessel.connection.connectionType, note));
+      } else {
+        // Print out details...
+        logs.basic(sprintf('\t%s\t%s\t%s', tessel.name, tessel.connection.connectionType, note));
       }
-
-      // Print out details...
-      logs.basic(sprintf('\t%s\t%s\t%s', tessel.name, tessel.connection.connectionType, note));
     });
 
     // Called after CTRL+C or timeout


Student007:
Tessel instructors can tell the users what `red` or `blue` means.

Frijol:
@Student007 Is this PR made to fix https://github.com/tessel/t2-start/issues/52? A better process for this would be to discuss a solution on the issue before implementing (I'm not sure that this solution fixes the problem).

Student007:
@Frijol it is ineffective to wait for people sleep so long :laughing: so I provide a possible solution - either you take it or you change it - as I try to discuss in Slack-Chat.

Student007:
@Frijol by the way: This is a quick fix, of cause - If I would take time to rework and discuss this, it wouldn't help the show at summit. For that reason I double set the same phrase for `note`. Also there are some things have to be done about `t2 list` and `Tessel.isProvisioned()` what causes in a totally different solution maybe (like caching known Tessels and a note if they could be found). But there are other major tasks to finish. Like `t2 root` I also stumbled over some things I have to work around.
Finally I couldn't test it with a USB Tessel - but it looks to me you could need this at the summit. So I wrote on the review call you also could rework and/or merge.

Frijol:
Hm, this fix doesn't make sense to me; I'm not sure what usability it is fixing. Can you explain? I can't tell what the red and blue denote.

Student007:
https://github.com/tessel/t2-start/issues/52
> New user here. I was following the instructions for "Find Your Tessel" and was a **little confused when it returned multiple Tessels** [...]

This will happen at events with free WLAN all the time ...
> I didn't realize that other Tessels on the LAN would be returned. 

@Frijol this is covered by your doc fix suggestion (I agree).

> **Maybe clarify that the USB connection is the one new users are looking for.**

**Mr. Speaker**: *the blue one is yours* 

(the red Tessels found are either others or your second one)

I though about to make the provisioned and authorized Tessels **green** but while testing several terminals I though it would better stay native.

**`t2 list` should only show a `blue` USB-Tessel (not provisioned) and show not authorized Tessels `red` in general (e.g. LAN-Tessels). All others (short: your own Tessels) will stay native.**

Frijol:
I think it's not a usability fix if the speaker has to say "the blue one is yours". It might even be more confusing: "why are these different colors?" "these red ones look like errors" etc.

Knowing that the USB one is the one connected to your computer is useful (good problem identification), but I think we should put more thought into the design.

There was some long discussion about this here: https://github.com/tessel/t2-cli/issues/86

adkron:
I like the idea of coloring them. The words next to the readout should help clarify any confusion and after using it a few times the colors will become quite obvious, and useful. I think this is a nice addition.

Student007:
@adkron me too - don't understand why @Frijol didn't love it. I vote for merging like it is - but if someone has a better wording, this could be done quickly.
Maybe @johnnyman727 and @rwaldron put their two cents in before we maybe have a bit more usability for events :heart_eyes: 

At the moment voting is 2:1 for merging I think ?

johnnyman727:
@Student007 I think coloring the text *could* be useful but it's hard to do it properly with the way `t2 list` works as is. What I mean is that it would be great if we could color the `default` Tessel green, unprovisioned LAN devices as grey (not red because that symbolizes error), and all other devices some other unassuming color. Unfortunately, we don't know what the `default` Tessel is until the discovery process is complete.

Perhaps the best we can do for now is mark USB and authorized LAN devices as green and unauthorized devices as grey?

Frijol:
I think @johnnyman727 's proposed solution is good. Good in hackathon settings and home use cases alike.

I like the suggestion of green & gray as well; gray is a good "you're not authorized" indicator, as opposed to "you have an error" red

LinusU:
I very much agree on the "no red". Red makes me think that something went wrong when it didn't.

Even if we don't know which one is the default until the discovery process is complete, we could easily amend the list as soon as discovery is completed. [`log-update`](https://www.npmjs.com/package/log-update) is a great package for this which I have used before.

Any time a tessel is discovered it would be added to a list, we would then call `logUpdate(...)` to print that list to the screen. When discovery is completed, we color one of them green and then call `logUpdate` again. Finally we call `logUpdate.done()` to signal that we are happy with the output and want to move on.

johnnyman727:
@LinusU +1

We also print out [what the default Tessel is](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L85) after listing them and we can color that name green as well.

Frijol:
by the way re @Student007 

> don't understand why @Frijol didn't love it

I love the enthusiasm, and the ideas behind it; I just want to make sure we fully think through the user-facing design decisions, since they have such a big impact.

Student007:
I agree - red is misleading (I simple decide to use red because of best contrast on my tested terminals - but I also would agree, the user could change the terminal layout if the contrast is bad).

> [...] log-update is a great package for this which I have used before.

@LinusU this would fix the contrast problem - great idea :+1: 

@Frijol  as you wrote #86 
>            Rafael        LAN
>            Bulbasaur     USB       (run `tessel provision` to authorize)
>            Mewcat        LAN       (USB connect and run `tessel provision` to authorize)
>            Jigglypuff    USB
>            Jigglypuff    LAN

I would use `(run 'tessel provision' to authorize)` but in the case we stay using colors instead of @LinusU idea, colored blue.

> Perhaps the best we can do for now is mark USB and authorized LAN devices as green and unauthorized devices as grey?

:+1: 
I don't know if a USB connected Tessel causes to filter out its detected LAN-Version (I couldn't test yet) I would agree. I think a not provisioned USB Tessel should be blue (it will be used seldom but event visitors who use a Tessel the first time would get the point faster).

Finally I think a combination of less colors (but colors) and `log-output` would be a contrast save but event supporting solution.

adkron:
As another story we could allow the user to pass --no-color to turn off color too

Student007:
@adkron hmm... thinking about this the implementation needs to be made different. But this also could be reasonable for a presenting situation - the projectors silver screen and sun could crash the experience guests should have by following the course.

I like this option because I am not fine with forcing to use green for valid things - it simple could have problems with contrast (like I figured out with different terminals).

@Frijol I think you did the most public presentations - could you put in your two cents here (maybe there also had been problems with native colorization) ?

Blue I feel was the best color to accent things to point peoples nose on. But only on lighted terminals - no dark. Blue on black ... there would be a option "inversed" what will make white on blue per line.

@johnnyman727 because this PR was meant as a quick fix for the summit, we can close the PR now and reference it from a late milestone description. If I am right you are thinking about new milestone structure at the moment ?

I am focused on a different feature at the moment (t2 root and --/key).

Frijol:
@adkron I'd lean against adding another flag – better to make a design decision here and keep it standard to reduce confusion.

@Student007 I don't have strong knowledge of the effects of forcing colors on different terminals (I use a black terminal with transparency, but at workshops a lot of beginners use white terminals). I'm guessing there exists a canned solution for this.

However, I think it would be better to table this issue for now instead of taking time to discuss it further; it's not of central importance compared to building out the core functionalities of Tessel 2 and I'm wary of [bikeshedding](https://en.wikipedia.org/wiki/Parkinson%27s_law_of_triviality).

Student007:
@Frijol :+1: .. but it has less priority on my list. First `t2 key` then `t2 root`... than this one

Issue #96, opened 2015-05-05
Merge support for binary modules


johnnyman727:
@allanca has done quite a bit of work on this already [and has an outstanding PR](https://github.com/tessel/cli/pull/179) on the original CLI. We need to get this merged into the Tessel 2 CLI and the remote server set up.

johnnyman727:
@wprater Here's the order of operations:
- [ ] The CLI needs to iterate through node_module dependencies and find those with binary dependencies. You can do that either by finding a `bindings.gyp` or checking for `nan` or `bindings` in the  `package.json` dependencies. [Here is an example](https://github.com/websockets/bufferutil/blob/master/package.json).
- [ ] Ping an S3 server to fetch each of the dependency binaries. I can set this up when you're ready or you can use a test server you have access to until this is about ready to merge. We'll need to figure out how to make the server accessible by all Team Members.
- [ ] Inject them into the CLI tarball. It will need to happen between [when we locate the entry point to the project and when we deploy it](https://github.com/tessel/t2-cli/blob/476f16763f1435f19d8d501cfde98a9a72460cfc/lib/tessel/deploy.js#L194-L203).

Bonus Points:
- [ ] If the binary dependency is not on the server already, have the t2-cli send a request to another server to run `t2-compiler` to create that dependency and then save it on the S3 instance with the rest of the deps. Once it's done, inject that binary into the tarball like normal.

You'll need to to use the `t2-compiler` for creating binary dependencies to test with. Feel free to ping @tcr if you have any issues with that.

rwaldron:
@wprater any updates?

rwaldron:
I'm going to look at the deploy routines and see where I can expose the data needed for this

rwaldron:
@johnnyman727 I've updated your list above to indicate items that function (which is only the first). The Third item is in progress, using dummy tarballs.

johnnyman727:
:+1:

Student007:
@johnnyman727, @tcr  or @rwaldron when I use [t2 compile](https://github.com/tessel/t2-compiler) I get a **tgz** what includes a folder `Release` containing binding.node

If I like to use a binary on the current implementation of the `t2-cli` to push/run code on the Tessel 2 - what is necessary to do with the result of the `t2 compiler` ?

I would assume, manually edit the package.json what is generated by `t2 init` and copy the **tgz** to ... ?? 
[...]

PR #385, opened 2015-10-18
Generates slim bundle for running on Tessel

### LINKED ISSUE #383: Reduce 'push' or 'run' bundle size by following 'require' paths
At the moment, running and pushing code to Tessel 2 takes the `entryPoint` file and everything in `node_modules/` to build a tarball to send over. [Reference](https://github.com/tessel/t2-cli/blob/dac8975498b3d39e7e9646d805502b2ee1a4b107/lib/tessel/deploy.js#L171)

I'm aware of the [`.tesselignore`](https://github.com/tessel/t2-cli/blob/dac8975498b3d39e7e9646d805502b2ee1a4b107/lib/tessel/deploy.js#L214) file, which is used to blacklist files from being used in tarball. That workflow would require developers to modify that file every time they install a new module and blacklist everything they don't use in their program. 

It would be nice to have an option to pass into the bundle process for only following the `require` paths of the `entryPoint`, similar to how Browserify works, to create a bundle of only the code needed for the program to run. It could be a flag in the cli, i.e. `t2 run index.js --follow-requires` or `t2 run index.js --slim`.

HipsterBrown:
Fixes #383 

~~**Blockers:**~~
~~I'm not sure if I am able to pipe the `browserify.bundle()` output stream into `tar.Pack()`, so it is currently pushing an un-tarred bundle to Tessel, which results in the error `invalid tar magic`.~~

~~Running the entry point file through `browserify` from the command line and running that as a single file to Tessel works with a greatly reduced bundle. I would love some feedback on how to better integrate this process into the cli workflow.~~

This is now a working implementation! :raised_hands: 

There is still some refactoring and testing left to do before this PR is complete but the idea works! Now, after Browserify is done bundling, the resulting buffer is written to a temporary `build.js` (under the default value for `options.slimPath`) and run through the typical fstream process before piping to `tar.Pack()` and resolving the Promise.


lib/tessel/deploy.js
```diff
@@ -205,43 +207,67 @@ actions.sendBundle = function(tessel, filepath, opts) {
...
+  var buffers = [];
+
+  if (opts.slim) {
+    console.log('Running slim...');
+    return new Promise(function(resolve, reject) {
+      var b = browserify(target, {
+        node: true,
+        ignoreMissing: true
+      });
+      b.transform(uglify);
```
rwaldron: .tesselignore mechanism has to be respected in both cases.

rwaldron:
I'm trying to think about issues that might arise once support for compiled modules arrives. Any thoughts here?

HipsterBrown:
Can you give me an example of what you mean by "compiled modules"? I'm not familiar with that terminology.

LinusU:
Module that uses C++ code, e.g. [fs-xattr](https://github.com/LinusU/fs-xattr).

HipsterBrown:
Hmmm I guess that is when the user would use the `.tesselignore` with the default bundling process. I'm not sure how to tell the `slim` bundling process to include non-JS code unless there was a custom dotfile or package.json config for referencing those packages.

johnnyman727:
@rwaldron that's a good point - I hadn't thought about modules with binary dependencies here.

@HipsterBrown we haven't built it yet (it's on my to-do) but the CLI will probably look for a `binding.gyp` in a node module root directory as an indicator of whether it has a binary dependency. If that's true, the CLI should fetch a pre-compiled binary from one of our servers and inject it into the tarball. I believe @wprater is working on it. Would that not work with how you've built this option?

LinusU:
> Hmmm I guess that is when the user would use the .tesselignore with the default bundling process. 

I don't think so since the user want that module actually running on the tessel.

Is there a reason to hide this behind a switch? Couldn't we make it the default behaviour?

HipsterBrown:
@johnnyman727 During the `browserify` bundling process we can check for 'file' events and trace the modules root directory for a `binding.gyp` and then do the fetching process if one if found. 

> Is there a reason to hide this behind a switch? Couldn't we make it the default behaviour?

@LinusU I think this would be a nice default behavior, however I didn't want to make assumptions before more people were able to take a look at the implementation.

rwaldron:
> I don't think so since the user want that module actually running on the tessel.

This is where it gets strange—it won't actually be **that** module, it will be somehow replaced by a special 
"pre-compiled-for-Tessel-2" version (that might be a naive explanation).

HipsterBrown:
That process of pulling in the pre-compiled binaries is still being worked, correct? [Reference Issue](https://github.com/tessel/t2-cli/issues/96)

rwaldron:
@HipsterBrown yessir

lib/tessel/deploy.js
```diff
@@ -205,49 +211,98 @@ actions.sendBundle = function(tessel, filepath, opts) {
 
 actions.tarBundle = function(opts) {
   var target = opts.target || process.cwd();
+  var pack = tar.Pack();
+  var buffers = [];
+
+  if (opts.slim) {
+    console.log('Running slim...');
```
johnnyman727: @HipsterBrown if you feel output is necessary, can you use `logs.basic` for consistency with the rest of the library? (Other options are `logs.warn` and `logs.err`)

lib/tessel/deploy.js
```diff
@@ -205,49 +211,98 @@ actions.sendBundle = function(tessel, filepath, opts) {
...
+    return new Promise(function(resolve, reject) {
+      var b = browserify(opts.resolvedEntryPoint, {
+        node: true,
+        ignoreMissing: true
+      });
+      b.transform(uglify);
```
johnnyman727: Lazy Jon wonders what `uglify` does.

johnnyman727:
It sounds like we can probably get the `--slim` option working once we have precompiled binaries but if it gets too tricky, we might have to take it back out at that point. @HipsterBrown are you okay with that?

@rwaldron could you review? You're much more familiar with the bundling logic than I am.

lib/tessel/deploy.js
HipsterBrown: Sure thing. Those were mostly for development, but while refactoring, I'll use the `logs` utility instead.

lib/tessel/deploy.js
HipsterBrown: It is a Browserify transform for minifying the code being bundled. [Reference](https://www.npmjs.com/package/uglifyify). I added it as another step to reduce the size of the tarball.

HipsterBrown:
@johnnyman727 That sounds fine to me. I'll do my best to make sure it works with the precompiled binaries (once there is a working implementation).

rwaldron:
> @rwaldron could you review? 

Will do.

rwaldron:
Despite the changes, this still doesn't respect .tesselignore rules. With the following script, I did a `t2 run index.js` and a `t2 run index.js --slim` to compare the results. 

```js
var tessel = require("tessel");
var lodash;

try {
  lodash = require("lodash");
} catch (error) {}

var result = lodash ? "FAIL" : "PASS";

console.log(".tesselignore blocked lodash from being included: ", result);

process.exit(0);
```

Here's the local directory: 

```
$ ls
.
├── [drwxr-xr-x rwaldron staff     102 Oct 21 16:35]  node_modules/
├── [-rw-r--r-- rwaldron staff      14 Oct 21 16:35]  .tesselignore
├── [-rw-r--r-- rwaldron staff     234 Oct 21 16:40]  index.js
└── [-rw-r--r-- rwaldron staff     362 Oct 21 16:46]  package.json

1 directory, 3 files
```

And here's the npm dependency graph

```
rwaldron at nova in ~/robotz/tessel-ignore-bundling
$ npm ls
tessel-ignore-bundling@0.0.0 /Users/rwaldron/robotz/tessel-ignore-bundling
└── lodash@3.10.1
```

Here's the .tesselignore

```
$ cat .tesselignore
node_modules/
```

(Which means that nothing from the node_modules directory will be included in the bundle that's pushed to the T2)

------------

Compare the results: 






There are two major issues: 

1. `t2 run index.js --slim` includes the entirety of `node_modules/`, when it should not. 
2. Since Browserify is intended to be used as a build/bundling tool for JS applications that will run in a browser environment, it has "shimmed" `process.exit()` in ["a browser-appropriate way"](https://github.com/substack/node-browserify#compatibility). It's been a while since I used Browserify for anything, so I went back and re-read the [handbook](https://github.com/substack/browserify-handbook#how-browserify-works). After doing so, it's clear to me that this is an inappropriate tool for this project to use, as it's mechanism for resolving the dependency graph (by tracing dependencies through `require()` calls encountered in an AST) precludes the use of a per-file filtering system, which is far more valuable for controlling the bundled contents in a non-restrictive way.

HipsterBrown:
This is great, constructive feedback!

> Despite the changes, this still doesn't respect .tesselignore rules.

I'm still not completely understanding the use case for a `.tesselignore` if the slim bundling processes follows the dependency tree of the entry point, i.e. `index.js`, and only bundling what it needs. Why would a user ignore a dependency needed by the program?

> Since Browserify is intended to be used as a build/bundling tool for JS applications that will run in a browser environment, it has "shimmed" process.exit() in "a browser-appropriate way".

This is a great point. I thought the `node` option would eliminate those "shimmed" areas since it is supposed to bundle the code in a way to still be run in Node. Thanks for clearing that up. 

---

Do you think it would be practical to replace this Browserify bundling (which clearly needs to be replaced) with a process of following the dependency graph to generate the ignore rules for the `fstream`? The rules would end up being similar to when passing the `--single` flag, ignore everything that's not the files in the dependency graph. This should still result in a smaller tarball being sent over to the Tessel.

LinusU:
I actually still think that browserify should work, something must have gone wrong when applying the node option. When that option is correctly applied, it won't look at the `browser` field, and it won't shim the builtins (e.g. `require('fs')`).

rwaldron:
> I'm still not completely understanding the use case for a .tesselignore if the slim bundling processes follows the dependency tree of the entry point, i.e. index.js, and only bundling what it needs. Why would a user ignore a dependency needed by the program?

I had some unconfirmed concerns about conditional `require()` calls, where a conditionally loaded module would be included and shouldn't be (ie. mocks and similar). After working through an exercise this afternoon I'm satisfied that this is not an issue. What remains: fixing the shims problem.

HipsterBrown:
> What remains: fixing the shims problem.

I got a mostly working implementation of following the dependency graph and generating the `ignoreRules` array based off those results. The only issue I've run into so far is with node_modules where the `main` is not an `index.js`. The solution to this issue is to find the `package.json` for that module and include that in the tarball. I've done some deep diving into the `module` package in Node to research how to do this efficiently in the build process. I should have this PR updated with that implementation sometime this weekend.

rwaldron:
@HipsterBrown is all of that necessary? After exhaustively testing refuting my concerns, your current solution is sound, it's just a matter of telling browserify to _not_ mess with the built-ins.

lib/tessel/deploy.js
```diff
@@ -205,49 +211,98 @@ actions.sendBundle = function(tessel, filepath, opts) {
...
+      b.bundle(function(err, results) {
+        if (err) {
+          console.log('Error');
```
rwaldron: Remove this line

HipsterBrown:
@rwaldron It might be. After seeing @LinusU's suggestion that the options were not correctly set, I dug into Browserify's cli source and found [this](https://github.com/substack/node-browserify/blob/master/bin/args.js#L67). I was setting `node` as an option when it's really just a cli shorthand for several other options to run through node. 

I set the options correctly (including setting `builtins` to false) and got an even smaller bundle than before, however there were issues running the bundle on Tessel. I had an error related to something called `_process`, which I believe may be an internal browserify module. Still investigating that challenge while consider other options. 

If anyone knows more about the internals of browserify to enlighten me, that would be wonderful. I can push my fixed browserify options up for anyone who wants to check it out as well.

HipsterBrown:
So I found where `_process` is referenced in browserify, [reference](https://github.com/substack/node-browserify/blob/b7cf5bef990a6823d14e2d437e15112aa6b3adfa/index.js#L551). It is also referenced in the [exports of `builtins`](https://github.com/substack/node-browserify/blob/b7cf5bef990a6823d14e2d437e15112aa6b3adfa/lib/builtins.js#L38), so I'm not sure if browserify is fully respecting [that option](https://github.com/substack/node-browserify/blob/b7cf5bef990a6823d14e2d437e15112aa6b3adfa/index.js#L493).

Is `_process` also a node internals thing?

rwaldron:
> I was setting node as an option when it's really just a cli shorthand for several other options to run through node.

Confirm.

> I set the options correctly (including setting builtins to false) and got an even smaller bundle than before, however there were issues running the bundle on Tessel. I had an error related to something called _process, which I believe may be an internal browserify module. Still investigating that challenge while consider other options.

Yep, same here. 

> Is _process also a node internals thing?

Yes

HipsterBrown:
@rwaldron Any reason why the `_process` module could not be found while running the browserify  bundle on Tessel?

rwaldron:
I have to recant the following statement: 

> I had some unconfirmed concerns about conditional require() calls, where a conditionally loaded module would be included and shouldn't be (ie. mocks and similar). After working through an exercise this afternoon I'm satisfied that this is not an issue. What remains: fixing the shims problem.

As I was working through this issue, I discovered that:

1. My previous observations were incorrect. Conditional require() calls **will result in inclusion**. Here's a concrete example:
  - index.js
  ```js
  var tessel = require("tessel");
  var UART = global.IS_TEST_MODE ?
    require("./mocks/uart-mock") :
    require("./uart");
  ```
  - ./mocks/uart-mock.js
  ```js
  "THIS IS THE MOCK";
  ```
  - ./uart.js
  ```js
  "THIS IS THE REAL API";
  ```
  - Resulting "slim" bundle: 
  ```js
  (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof require=="function"&&require;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof require=="function"&&require;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(require,module,exports){
  var tessel=require("tessel"),UART=global.IS_TEST_MODE?require("./mocks/uart-mock"):require("./uart");
  },{"./mocks/uart-mock":2,"./uart":3,"tessel":undefined}],2:[function(require,module,exports){
  "THIS IS THE MOCK";

  },{}],3:[function(require,module,exports){
  "THIS IS THE REAL API";

  },{}]},{},[1]);
  ```
  This example is trivial, but consider that this means Johnny-Five programs result in the inclusion of the entire node-serialport module—which is never going to be used by Johnny-Five on this platform (this is a good example of why .tesselignore files exist.)

2. The correct combination of options (discovered by @reconbot) to produce node-friendly bundles is: 
  ```js
  {
    builtins: false,
    commondir: false,
    browserField: false,
    detectGlobals: false,
    ignoreMissing: true,
  }
  ```
  https://github.com/HipsterBrown/t2-cli/pull/1

rwaldron:
@HipsterBrown this got me thinking...

> I got a mostly working implementation of following the dependency graph and generating the ignoreRules array based off those results. The only issue I've run into so far is with node_modules where the main is not an index.js. The solution to this issue is to find the package.json for that module and include that in the tarball. I've done some deep diving into the module package in Node to research how to do this efficiently in the build process.

And it inspired me to take one last crack at our problem and I think I've got it. Here's a run down of the semantics that I've come up with: 

1. If `opts.slim` is `true`, 
  1. `glob` for all `./**/*.tesselignore` files; 
    1. let `exclusions` be the result of combining the contents of each .tesselignore into a single `{pattern}`; 
  1. let `b` be a new `browserify` bundler object with the existing options that @reconbot discovered
  1. let `files` be `glob.sync(exclusions)`
    1. for each `file` of `files`
      1. call `b.exclude(file)`
  1. call `b.bundle(...)`
  1. on completion of `bundle`, minify and write bundled program code. 

With this algorithm, --slim deployments:

1. respect all the .tesselignore rules
1. respect the Node.js built-ins
1. benefit from browserify dependency graph tracing
1. **_dramatically_** reduce the deployment size


The results are utterly astounding. I brought a Johnny-Five program from ~1mb down to 376kb!

HipsterBrown:
@rwaldron That looks like a very thorough and efficient process with some awesome results. :+1:

> let exclusions be the result of combining the contents of each .tesselignore into a single {pattern};

I'll need some help with this since I don't work with `globs` and pattern matching very often. I'm not sure what to research or where to begin with that point. I guess I could look at the source for `fstream-ignore` to see how that does it. 

Everything else looks pretty straightforward. Did you have some of that working already?

rwaldron:
> What files are we bringing in beyond the bundle? Why do we need node_modules at all?

With --slim, only the `bundle.js` ships

> Everything else looks pretty straightforward. Did you have some of that working already?

The whole thing

rwaldron:
@HipsterBrown we're still building this on your patch, here's the modifications: https://github.com/HipsterBrown/t2-cli/pull/2

HipsterBrown:
@rwaldron Rad! Anything left to do?

HipsterBrown:
Merged your PR @rwaldron.

rwaldron:
@HipsterBrown pleasure working with you on this—great teamwork here :)

johnnyman727:
:clap: :clap: :clap: Nice job getting this merged everyone! We did neglect to add unit tests for the option but I'll create a separate issue for it.

HipsterBrown:
@rwaldron I definitely learned a bunch about `globs` and bundling from this build. Thanks for every patience and feedback. :+1:

Frijol:
Should `--slim` be default behavior? Any reason you wouldn't always want this?

johnnyman727:
btw, I love this option. I deployed a folder w/o `--slim` and it was 63k, with `--slim` it was less than 3kb and deployed instantly over USB.

Frijol:
thoughts on showcasing `--slim` [on the Tweet example page of the start experience](http://tessel.github.io/t2-start/tweet.html)? It's the first time we use a big library

johnnyman727:
I'd be for it!

HipsterBrown:
There should probably be some documentation around `--slim` that we can link to if the Tweet example page is updated.

johnnyman727:
+1

I tested out the `twitter` example just in case and it did work as expected. The slim option cuts deployment size to 1/6 of the usual (472kB instead of 2.8MB)

Frijol:
done https://github.com/tessel/t2-start/pull/70

Frijol:
roger @HipsterBrown https://github.com/tessel/t2-docs/issues/35

leopornorecio (approved):

Issue #346, opened 2015-09-27
T2 CLI does not properly terminate USB transfers


Frijol:
Tessel 2 takes a while to boot up. If you run a command while it is still booting, it messes up the USB comms.

rwaldron:
I think there needs to be some way of checking that it's ready to go. I'm not sure how though.

Frijol:
@johnnyman727 was using the existence of USB comms from the Tessel to determine whether it's ready. Does that work in this case?

johnnyman727:
Yes, I was working in [this PR](https://github.com/tessel/t2-cli/pull/291/files). It does work in theory but then the `usb` module seems to crash each subsequent time it's used...

tcr:
See #386 for a fix.

Frijol:
:raised_hands:

PR #386, opened 2015-10-19
Closes USB comms properly to allow communication while booting.

tcr:
See #346, #291.

Only tested with `t2 list` for now.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -6,6 +6,7 @@ var controller = require('../lib/controller');
 var key = require('../lib/key');
 var init = require('../lib/init');
 var logs = require('../lib/logs');
+var usb = require('../lib/usb_connection');
 
 function makeCommand(commandName) {
   return parser.command(commandName)
@@ -235,7 +236,10 @@ module.exports = function(args) {
 };
 
 module.exports.closeSuccessfulCommand = function() {
-  process.exit(0);
+  usb.terminateConnections()
+  .then(function () {
+    process.exit(0);
+  });
 };
 
 // Allow options to be partially applied
@@ -252,7 +256,10 @@ module.exports.closeFailedCommand = function(opts, err) {
     logs[opts.type](err);
   }
   // NOTE: Exit code is non-zero
-  process.exit(1);
+  usb.terminateConnections()
+  .then(function () {
+    process.exit(1);
+  });
 };
 
 
--- a/lib/discover.js
+++ b/lib/discover.js
@@ -49,6 +49,8 @@ TesselSeeker.prototype.start = function(opts) {
     debug('Will scan for USB devices');
     // Start the scan
     self.usbScan = usb.startScan();
+    // Track how many are resolved.
+    self.usbScanCount = 0;
     // When we get a connection, handle it
     self.usbScan.on('connection', connectionHandler.bind(self));
   }
@@ -81,17 +83,16 @@ TesselSeeker.prototype.start = function(opts) {
           debug('Fetching name:', tessel.connection.host);
           tessel.getName()
             .then(function() {
+              self.usbScanCount++;
               self.emit('tessel', tessel);
-              return Promise.resolve();
-            })
-            .catch(function(err) {
-              logs.err('unable to fetch name:', err);
               return Promise.resolve();
             });
         }
-      }).catch(function() {
+      })
+      .catch(function() {
         return Promise.resolve();
       });
+
     // Push this Promise into the pending array
     pendingOpen.push(opening);
   }
@@ -123,6 +124,12 @@ TesselSeeker.prototype.stop = function() {
   }
 
   if (this.usbScan !== undefined) {
+    if (this.usbScanCount != this.usbScan.count) {
+      logs.warn('%d Tessel device%s connected by USB may still be booting.',
+        this.usbScan.count - this.usbScanCount,
+        (this.usbScan.count - this.usbScanCount) == 1 ? '' : 's'); 
+    }
+
     this.usbScan.stop();
 
     this.usbScan = undefined;
--- a/lib/tessel/tessel.js
+++ b/lib/tessel/tessel.js
@@ -97,11 +97,13 @@ Tessel.prototype.simpleExec = function(command) {
   var self = this;
   return new Promise(function(resolve, reject) {
     // Stop processes and delete everything in the folder
-    return self.connection.exec(command)
+    self.connection.exec(command)
       .then(function(remoteProcess) {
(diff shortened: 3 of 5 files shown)


johnnyman727:
@tcr thanks for working on this!

So we have some infrastructure in place to make sure all Tessel connections get closed. Essentially, every command run will end with [`controller.closeTesselConnections`](https://github.com/tessel/t2-cli/blob/bc9c59f16359d68b1d73fd9fbfd13094683b9462/lib/controller.js#L312) which ultimately calls [`tessel.close`](https://github.com/tessel/t2-cli/blob/bc9c59f16359d68b1d73fd9fbfd13094683b9462/lib/tessel/tessel.js#L16-L29) and that function, if it has a USB connection, will ultimately call [`USBConnection._close`](https://github.com/tessel/t2-cli/blob/bc9c59f16359d68b1d73fd9fbfd13094683b9462/lib/usb_connection.js#L170-L184).

It seems as though the significant part of your PR is calling `self.epIn.stopPoll` before releasing the interface. Can we just include that portion of your PR to have the same effect? It seems like the rest of the PR is just making sure connections get closed which they already do.

Additionally, I believe this will throw an error if you have any pending packets to be sent over USB, won't it?

tcr:
@johnnyman727 `controller.closeTesselConnections` will not be called with Tessels that don't identify themselves, i.e. a Tessel while booting does not wind up in the list of Tessels that are closed. How do we deal with that?

As for your second question, look at `lib/usb_connection.js` line 181 in my change. Essentially, given all transfers will timeout, it simply spin-waits until the last transfer is complete to resolve it. This is hacky, but reliable :)

rwaldron:
@tcr need to run `grunt` tests and tooling locally

lib/tessel/tessel.js
```diff
@@ -97,11 +97,13 @@ Tessel.prototype.simpleExec = function(command) {
   var self = this;
   return new Promise(function(resolve, reject) {
     // Stop processes and delete everything in the folder
-    return self.connection.exec(command)
```
johnnyman727: Why did you need to remove this `return`?

lib/tessel/tessel.js
```diff
@@ -97,11 +97,13 @@ Tessel.prototype.simpleExec = function(command) {
...
+    self.connection.exec(command)
       .then(function(remoteProcess) {
-        return self.receive(remoteProcess).then(function(received) {
-          return resolve(received.toString());
-        }).catch(reject);
+        return self.receive(remoteProcess)
```
johnnyman727: Was this just a style change?

lib/usb_connection.js
```diff
@@ -175,12 +178,18 @@ USB.Connection.prototype._close = function(callback) {
...
-    if (typeof callback === 'function') {
-      callback();
-    }
-  }.bind(self));
+  self.epIn.stopPoll(function() {
+    self.intf.release(true, function attempt() {
+      try {
+        self.device.close();
+      } catch (e) {
+        setTimeout(attempt, 100);
```
johnnyman727: It would be great if we had a timeout limit to reduce the chance of an infinite loop.

lib/discover.js
```diff
@@ -81,17 +83,16 @@ TesselSeeker.prototype.start = function(opts) {
...
-            })
-            .catch(function(err) {
-              logs.err('unable to fetch name:', err);
-              return Promise.resolve();
             });
         }
-      }).catch(function() {
+      })
+      .catch(function() {
         return Promise.resolve();
```
johnnyman727: Why not check the connection time and report the booting Tessel here rather than preserving a count of open scans?

bin/tessel-2.js
```diff
@@ -235,7 +236,10 @@ module.exports = function(args) {
 };
 
 module.exports.closeSuccessfulCommand = function() {
-  process.exit(0);
+  usb.terminateConnections()
```
johnnyman727: I'd prefer if we kept all tessel discovery and closing logic in `controller.js`. Could you move it to [`closeTesselConnections`](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L312) which gets called after every command?

lib/usb_connection.js
```diff
@@ -175,12 +178,18 @@ USB.Connection.prototype._close = function(callback) {
...
-  self.intf.release(true, function() {
-    self.device.close();
-    if (typeof callback === 'function') {
-      callback();
-    }
-  }.bind(self));
+  self.epIn.stopPoll(function() {
+    self.intf.release(true, function attempt() {
+      try {
+        self.device.close();
```
johnnyman727: If it's successfully released should we remove it from `usbCollection`?

johnnyman727:
@tcr it works great on my machine! It reported when I had Tessels booting and they continue to work once they were up.

I added a few comments to clean up the code a bit before merging.

Thanks!

lib/tessel/tessel.js
LinusU: `.toString()` is no longer run now, but maybe it wasn't needed

Frijol:
Got this while working from your branch, is it from the branch itself or something else?

```sh
➜  ~  t2 wifi
WARN 1 Tessel device connected by USB may still be booting.
WARN No Authorized Tessels Found.
events.js:141
            ^
Error: Connection was already closed...
    at USB.Connection._write (/Users/frijol/t2-cli/lib/usb_connection.js:70:14)
    at doWrite (_stream_writable.js:291:12)
    at clearBuffer (_stream_writable.js:401:7)
    at onwrite (_stream_writable.js:336:7)
    at OutEndpoint.WritableState.onwrite (_stream_writable.js:88:5)
    at Transfer.callback (/Users/frijol/t2-cli/node_modules/usb/usb.js:343:14)
```

Frijol:
Also, getting false positives from this branch where it says "may still be booting" but really it's booted but I'm having USB issues. Not sure if we want to address this or assume USB should be working properly

johnnyman727:
I also just had the case where it had definitely finished booting but then the CLI reported that it was still booting. This PR will need some thorough tests.

lib/tessel/tessel.js
johnnyman727: @LinusU it looks like it's still run below at 104.

johnnyman727:
Also related https://github.com/tessel/t2-firmware/issues/119

johnnyman727:
@tcr I've superceded this PR with https://github.com/tessel/t2-cli/pull/400/files. It takes some of the logic from this PR, some of the logic from #291 and some new logic to achieve the correct result. It also needs tessel/t2-firmware#118 merged.

Issue #388, opened 2015-10-20
Remote Process: request actual available memory


rwaldron:
- `cat /proc/meminfo`
- ~~Use `MemAvailable`~~
  - Recommended by https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

This is needed for #382

Issue #382, opened 2015-10-18
Limit the size of the bundle to be deployed


johnnyman727:
Tessel has 64MB of RAM but not all 64MB are available because OpenWRT is running other processes and has some files (like `/tmp`) mounted in RAM. 

Currently, you can try to deploy however large a bundle you'd like which could probably really mess up OpenWRT.

For now, we can check that the bundle is 32MB or less but we should eventually ask OpenWRT how much memory is free before deploying and limit it to that amount.

rwaldron:
:+1:

tcr:
:+1:

HipsterBrown:
Is this related to https://github.com/tessel/t2-cli/issues/383? Or this a separate check to make sure there is room for the bundle before deploying?

rwaldron:
There are two parts to this: 

1. [Checking actual available memory](https://github.com/tessel/t2-cli/issues/388)
2. Checking that the bundle byte length doesn't exceed the available memory; this operation calls the remote process command in part 1, then: 
  - **resolves**, then proceeds with [this operation](https://github.com/tessel/t2-cli/blob/master/lib/tessel/deploy.js#L191-L199)   
  - **rejects**, with some message describing the rejecting.

dtex:
What about room for execution? Should we report back available memory after deployment? Maybe color code if it's below a certain threshold?

johnnyman727:
@rwaldron we should probably leave some amount of spare RAM for the user application. I'm not sure what a decent amount is though :/ Maybe a few MB?

rwaldron:
@johnnyman727 @dtex agree with both things. The previous part of the puzzle was just a mechanism to make the data available. All we need to do is decide how much we want to spare and go from there.

johnnyman727:
I think we can set an arbitrary spare RAM limit like 10MB and just see how well that works. Soon after launching, it would be awesome to put some analytics on the CLI and see what the average bundle sizes look like.

Frijol:
The fix for this will be located in t2-cli/deploy.js. Here's the function that deals with memory information: https://github.com/tessel/t2-cli/blob/master/lib/tessel/deploy.js#L37

PR #389, opened 2015-10-20
Commands: support memory info request. Fixes gh-388

rwaldron:
The information contained within the result can be used to implement the checks described in https://github.com/tessel/t2-cli/issues/382


bin/tessel-2.js
```diff
@@ -171,6 +171,11 @@ makeCommand('wifi')
   })
   .help('Configure the wireless connection');
 
+makeCommand('memory')
```
johnnyman727: Do we need to make this accessible to the user? It should hopefully rarely come up so I'd shy away from adding *more* commands.

lib/controller.js
```diff
@@ -616,6 +616,20 @@ controller.getWifiInfo = function(opts) {
...
+controller.getMemoryInfo = function(opts) {
+  opts.authorized = true;
+  return controller.standardTesselCommand(opts, function(tessel) {
+    return tessel.memoryInfo()
+      .then(function(meminfo) {
+        Object.keys(meminfo).forEach(function(key) {
+          logs.basic(sprintf('\t%s\t%d kB', key, meminfo[key] / 1000));
+        });
+
+        return controller.closeTesselConnections([tessel]);
```
johnnyman727: You don't need to close the Tessel connections if you use `standardTesselCommand` - it closes the Tessel connection once the provided function resolves. Nice eh? :stuck_out_tongue:

johnnyman727:
Comments added. It mostly looks good but I don't think we should make this functionality command line accessible, just easy to use by other functions in the CLI. It should rarely come up for users and the output is hard to parse as a human anyway.

bin/tessel-2.js
rwaldron: No, but I figured for your testing purposes it would be useful.

lib/controller.js
rwaldron: Oh, right—totally forgot!

rwaldron:
> I don't think we should make this functionality command line accessible,

WFM.

lib/controller.js
rwaldron: Well, this is unnecessary anyway, since we're not exposing this as a user facing command now :)

rwaldron:
@johnnyman727 updated according to review notes!

johnnyman727:
Looks fantastic! Feel free to merge.

PR #372, opened 2015-10-13
Update to use simplified t2-firmware LED API

rwaldron:
More details here: https://github.com/tessel/t2-firmware/pull/105

PR #398, opened 2015-10-26
fixes https://github.com/tessel/t2-start/issues/53

Frijol:
changes index.js to a shorter blinky example which matches t2-start

r? @johnnyman727 or @HarleyKwyn 

Do I need to change anything other than this one file?


johnnyman727:
This is the same fix as #372 with the exception of the `// blinky` comment. I'm happy merging either as soon as I put out a new build.

PR #400, opened 2015-10-26
Warn when Tessel is booting

johnnyman727:
Note that Tessel USB comms will still be frozen until https://github.com/tessel/t2-firmware/pull/118 is also merged.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -62,7 +62,7 @@ Tessel.list = function(opts) {
       // If there were no Tessels found
       if (foundTessels.length === 0) {
         // Report the sadness
-        reject(noTessels);
+        return reject(noTessels);
       } else if (foundTessels.length === 1) {
         // Close all opened connections and resolve
         controller.closeTesselConnections(foundTessels)
--- a/lib/discover.js
+++ b/lib/discover.js
@@ -79,18 +79,17 @@ TesselSeeker.prototype.start = function(opts) {
           self.emit('tessel', tessel);
         } else {
           debug('Fetching name:', tessel.connection.host);
-          tessel.getName()
+          return tessel.getName()
             .then(function() {
               self.emit('tessel', tessel);
               return Promise.resolve();
-            })
-            .catch(function(err) {
-              logs.err('unable to fetch name:', err);
-              return Promise.resolve();
             });
         }
-      }).catch(function() {
-        return Promise.resolve();
+      }).catch(function(err) {
+        debug('Error opening device', err);
+        if (tessel.connection.connectionType === 'USB') {
+          logs.warn('Detected a Tessel that may be booting.');
+        }
       });
     // Push this Promise into the pending array
     pendingOpen.push(opening);
--- a/lib/tessel/tessel.js
+++ b/lib/tessel/tessel.js
@@ -95,15 +95,13 @@ Tessel.prototype.receive = function(remote) {
 
 Tessel.prototype.simpleExec = function(command) {
   var self = this;
-  return new Promise(function(resolve, reject) {
-    // Stop processes and delete everything in the folder
-    return self.connection.exec(command)
-      .then(function(remoteProcess) {
-        return self.receive(remoteProcess).then(function(received) {
-          return resolve(received.toString());
-        }).catch(reject);
+  // Stop processes and delete everything in the folder
+  return self.connection.exec(command)
+    .then(function(remoteProcess) {
+      return self.receive(remoteProcess).then(function(received) {
+        return Promise.resolve(received.toString());
       });
-  });
+    });
 };
 
 Tessel.PUSH_PATH = '/app/';
--- a/lib/usb_connection.js
+++ b/lib/usb_connection.js
@@ -82,22 +82,8 @@ USB.Connection.prototype._receiveMessages = function() {
   var transferSize = 4096;
   // Start polling
   self.epIn.startPoll(2, transferSize);
-
   // When we get data, push it into the stream
   self.epIn.on('data', self.push.bind(self));
-
-  // Report errors as they arise
-  self.epIn.on('error', function(e) {
-    // If this stream was already closed, just return
-    if (self.closed) {
-      return;
-    } else {
-      // Otherwise print the error
-      logs.err('Error reading USB message endpoint:', e);
-      // Return a non-zero return code
-      process.exit(-5);
-    }
-  });
 };
 
 USB.Connection.prototype.open = function() {
@@ -130,7 +116,7 @@ USB.Connection.prototype.open = function() {
       self.epIn = self.intf.endpoints[0];
       self.epOut = self.intf.endpoints[1];
       if (!self.epIn || !self.epOut) {
-        return reject(new Error('Device endpoints weren not able to be loaded'), self);
+        return reject(new Error('Device endpoints were not able to be loaded'), self);
       }
 
       // Map desciptions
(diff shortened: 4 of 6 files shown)


lib/tessel/tessel.js
```diff
@@ -95,15 +95,13 @@ Tessel.prototype.receive = function(remote) {
...
-    // Stop processes and delete everything in the folder
-    return self.connection.exec(command)
-      .then(function(remoteProcess) {
-        return self.receive(remoteProcess).then(function(received) {
-          return resolve(received.toString());
-        }).catch(reject);
-      });
-  });
+  // Stop processes and delete everything in the folder
+  return self.connection.exec(command)
```
johnnyman727: I just removed an unnecessary Promise wrapping.

lib/usb_connection.js
```diff
@@ -50,12 +53,23 @@ USB.Connection.prototype.exec = function(command) {
...
         // Once the command has been written, call the callback with the resulting process
-        proc.control.once('finish', resolve.bind(self, proc));
+        proc.control.once('finish', function commandWritten() {
+          // Clear the command failed timeout
+          clearTimeout(controlTimeoutHandle);
+          // Resolve with the process
+          resolve(proc);
+        });
+
+        controlTimeoutHandle = setTimeout(function() {
```
johnnyman727: We need this in order to actually get an error returned from `.exec` and raised back into the discovery process.

johnnyman727:
https://github.com/tessel/t2-firmware/pull/118 is a prereq.

lib/controller.js
```diff
@@ -62,7 +62,7 @@ Tessel.list = function(opts) {
       // If there were no Tessels found
       if (foundTessels.length === 0) {
         // Report the sadness
-        reject(noTessels);
+        return reject(noTessels);
```
johnnyman727: Shouldn't change anything because this is an if-else block with nothing after it but it's good to have.

lib/discover.js
```diff
@@ -79,18 +79,17 @@ TesselSeeker.prototype.start = function(opts) {
           self.emit('tessel', tessel);
         } else {
           debug('Fetching name:', tessel.connection.host);
-          tessel.getName()
+          return tessel.getName()
```
johnnyman727: Also shouldn't change anything but at least we don't swallow the error

lib/discover.js
```diff
@@ -79,18 +79,17 @@ TesselSeeker.prototype.start = function(opts) {
...
               return Promise.resolve();
-            })
-            .catch(function(err) {
-              logs.err('unable to fetch name:', err);
-              return Promise.resolve();
             });
         }
-      }).catch(function() {
-        return Promise.resolve();
+      }).catch(function(err) {
```
johnnyman727: `.open` should `reject` if USB Connections aren't working.

lib/tessel/tessel.js
```diff
@@ -95,15 +95,13 @@ Tessel.prototype.receive = function(remote) {
...
   var self = this;
-  return new Promise(function(resolve, reject) {
-    // Stop processes and delete everything in the folder
-    return self.connection.exec(command)
-      .then(function(remoteProcess) {
-        return self.receive(remoteProcess).then(function(received) {
-          return resolve(received.toString());
-        }).catch(reject);
+  // Stop processes and delete everything in the folder
+  return self.connection.exec(command)
```
johnnyman727: Just cleaning up a little bit - there is no need to wrap in the Promise constructor.

lib/usb_connection.js
```diff
@@ -82,22 +82,8 @@ USB.Connection.prototype._receiveMessages = function() {
   var transferSize = 4096;
   // Start polling
   self.epIn.startPoll(2, transferSize);
-
   // When we get data, push it into the stream
   self.epIn.on('data', self.push.bind(self));
-
-  // Report errors as they arise
-  self.epIn.on('error', function(e) {
```
johnnyman727: Removes the event listener that immediately kills the script upon endpoint error. This is bad because all process exiting should be through `tessel-2-bin.js` and because there is no way to handle it. 

However, by removing this, miscellaneous endpoint issues that occur after the `open` function aren't going to be caught. We can revisit this issue later.

lib/usb_connection.js
```diff
@@ -142,6 +128,15 @@ USB.Connection.prototype.open = function() {
         self.serialNumber = data;
         // Register this connection with daemon (keeps track of active remote processes)
         Daemon.register(self);
+
+        // If the USB Pipe isn't enabled on the other end (ie it is booting)
+        self.epIn.on('error', function(err) {
```
johnnyman727: This event will be thrown if Tessel isn't ready to open the USB pipe (ie if OpenWRT isn't booted) once `_receiveMessages` is called..

lib/usb_connection.js
```diff
@@ -175,12 +170,18 @@ USB.Connection.prototype._close = function(callback) {
   }
 
   self.closed = true;
-  self.intf.release(true, function() {
-    self.device.close();
-    if (typeof callback === 'function') {
-      callback();
-    }
-  }.bind(self));
+  self.epIn.stopPoll(function() {
```
johnnyman727: Adding `stopPoll` is apparently the sanitary thing to do when closing the connection.

lib/controller.js
rwaldron: :+1:

lib/discover.js
rwaldron: :+1:

lib/usb_connection.js
```diff
@@ -175,12 +170,18 @@ USB.Connection.prototype._close = function(callback) {
...
+    self.intf.release(true, function attempt() {
+      try {
+        self.device.close();
+      } catch (e) {
+        setTimeout(attempt, 100);
+      }
+      if (typeof callback === 'function') {
+        callback();
+      }
+    }.bind(self));
```
rwaldron: `bind(...)`isn't necessary if you already have a nasty `self` alias floating around ;)

johnnyman727: Good point!

rwaldron:
To review this, I:

- built t2-firmware with this patch: https://github.com/tessel/t2-firmware/pull/118 and did updated via dfu-util.
- applied the proposed patch in this PR to t2-cli

The result was awesome—worked perfectly on first try :)

One thing to ask about...

```
$ t2 list
WARN Discovered USB Tessel that may be booting.
WARN No Tessels Found.
```

I was thinking it might be nice to just kill the operation after the first warning, but then I plugged in two and made sure one was booting: 

```
$ t2 list
WARN Discovered USB Tessel that may be booting.
	bishop	USB
	bishop	LAN
```

`bishop` is the one that's ready to go, but is this warning odd?

rwaldron:
I just went back and tried `t2 wifi -n JLY4Q_EXT -p ...` without the t2-firmware patch and usb commands appear to wait until the T2 is booted.

johnnyman727:
Thanks for reviewing @rwaldron! 

> bishop is the one that's ready to go, but is this warning odd?

You're saying that it's strange that the CLI reports that a USB Tessel is booting but then it shows up anyway and the user might be confused about whether that one is booting?

How about a warning message that just doesn't mention the connection type:
```
WARN Discovered a Tessel that may be booting.
```

rwaldron:
> You're saying that it's strange that the CLI reports that a USB Tessel is booting but then it shows up anyway and the user might be confused about whether that one is booting?

Sort of. In that scenario, `bishop` was already booted, while `chappie` was still booting. It showed me `bishop` (makes sense), but seeing it listed after the warning was strange. I think that maybe all it needs is this: 

```
$ t2 list
WARN Discovered USB Tessel that may be booting.
INFO Detected the following available Tessel(s):
    bishop  USB
    bishop  LAN
```

Or something similar?

rwaldron:
Maybe also, instead of:

```
WARN Discovered USB Tessel that may be booting.
```

This:

```
WARN Detected a Tessel that may be booting.
```

johnnyman727:
Gotcha! I think your concern is valid but I can't see a clean way to know that the booting Tessel warning will print out before the available Tessels (the CLI could in theory find a booting Tessel after a USB Tessel). In the interest of getting the CLI published tonight, I'm not going to invest my time in that change but I will keep it in mind tomorrow during our workshop to see if it trips anyone up (probably not since you need two Tessels to see it).

I will make the second change of not mentioning that the booting Tessel is connected via USB. Thanks!

PR #416, opened 2015-10-27
Fixes installation on Linux

johnnyman727:


bin/tessel-2.js
```diff
@@ -40,6 +40,11 @@ function callControllerCallback(methodName) {
   };
 }
 
+parser.command('install-drivers')
+  .callback(function() {
+    require('./tessel-install-drivers');
```
LinusU: It's considered a very bad practice to run logic just by including a file. Please export a function, move the require to the top of the file, and call the function here :)

rwaldron: Or spawn it from a child process

resources/85-tessel.rules
```diff
@@ -0,0 +1 @@
++ATTRS{idVendor}=="1209", ATTRS{idProduct}=="7551", ENV{ID_MM_DEVICE_IGNORE}="1", MODE="0666"
```
kevinmehall: No `+` at the beginning of the line. (Was this copied from a diff?)

This should have a different name, or contain both T1 and T2 rules. Otherwise, if the user has the T1 CLI installed, it will overwrite the file with the T1 rules.

johnnyman727: Oops - that's what happens when you're jet lagged and writing code while being approached with questions from every angle. Fixed in a subsequent commit. Thanks!

bin/tessel-install-drivers.js
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+
+var fs = require('fs');
+var logs = require('../lib/logs');
+var child_process = require('child_process');
+
```
rwaldron: If this is just for linux, then a condition needs to exist that checks: 

```js
var os = require('os');
var baillist = ['darwin', 'win32'];

if (baillist.indexOf(os.platform()) !== -1) {
  process.exit(0);
}
```

bin/tessel-2.js
rwaldron: Same check I mentioned below.

bin/tessel-install-drivers.js
johnnyman727: @rwaldron isn't that what I have on [line 8](https://github.com/tessel/t2-cli/pull/416/files#diff-f195a6f2706c44203b9d76c77ba72d35R8)?

rwaldron: No, I mean bail out the whole script before anything else runs

rwaldron: You could just move that check out to here

johnnyman727: @rwaldron but I want to keep the check within the `install` function. Otherwise, as soon as `bin/tessel-2.js` imports this file, the CLI will exit.

johnnyman727:
Will merge since I haven't heard any complaints.

bin/tessel-install-drivers.js
rwaldron: The solution to that is: 

```js
var os = require('os');
var baillist = ['darwin', 'win32'];

if (baillist.indexOf(os.platform()) !== -1) {
  return;
}
```

johnnyman727: @rwaldron the first item within `.install` is an OS check. Are you opposed to the way that I did it? (which was copied from the T1 CLI).

Issue #406, opened 2015-10-27
Faster Operations for "low throughput" commands


johnnyman727:
I have a suggestion for making it faster to use Tessel and I'd love to hear other's thoughts on it. I used Tessel 1 recently and was impressed at how fast each command was (`t2 wifi`, etc.) and I thought it was a shame that T2 can be so slow because we spend 5 seconds (by default) "finding" your Tessel. The thought process behind this was that finding a LAN connection will be faster than USB which is true for deploying code bundles. But for simple commands with very small transactions like `rename`, `wifi`, `provision`, `erase` (pretty much everything but `run` and `push`), we might as well just use a USB connection if it exists because we'll default to that Tessel anyway.

**In essence, I'm suggesting a class of "simple commands" in which a USB connection will be selected right away, thus saving almost 5 seconds each time.**

Frijol:
> I used Tessel 1 recently and was impressed at how fast each command was

hahaha

I like this idea but we would need to be careful about messaging to the user; it's already fairly confusing from the user perspective what's going on with USB vs LAN connections, and this would make the automatic selection seem even more stochastic

johnnyman727:
> I like this idea but we would need to be careful about messaging to the user; it's already fairly confusing from the user perspective what's going on with USB vs LAN connections, and this would make the automatic selection seem even more stochastic

Agreed! Perhaps we should not print out the connection method selected when we run operations?

tcr:
I was going to propose we actually make `--lan` NOT the default. Might take a lot more discussion in person, as well as fixing the USB transfer speeds, but I think this makes sense for most of our usecases.

johnnyman727:
@tcr I agree that `-usb` by default would be great if we can get the speed up but I'm not sure how much room for improvement there really is.

MrNice:
@johnnyman727 Where are the notes on what's going on during a deployment?

Intuitively, it doesn't make sense for wireless operations to be faster than wired ones. And also, wireless deployment didn't work easily when I tried a month ago.

I wonder if there's a solution that looks similar to `sudo dd rdisk` - straight copying bytes over in order.

![img_20151027_101219](https://cloud.githubusercontent.com/assets/2587335/10766178/74392a68-7c93-11e5-9abc-6dd263f79714.jpg)

MrNice:
But to fix the problem now, I think we should default to USB and add a temporary console.warn that says "Warning: USB deployment is very slow, please use WLAN deployment if possible until we've fixed this"

[...]

PR #424, opened 2015-10-31
Prefer USB for simple operations

johnnyman727:
I wrote this up on a plane ride back from the UK. It makes all the low data-rate operations execute over USB as soon as it's found (unless the `--lan` flag is provided). Fixes https://github.com/tessel/t2-cli/issues/406. `deploy` and `update` wait for a LAN connection to be found. This makes using the Tessel feel way faster to me.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -35,11 +35,16 @@ function makeCommand(commandName) {
     })
     .option('lan', {
       flag: true,
-      help: 'Use LAN connection'
+      help: 'Use only a LAN connection'
     })
     .option('usb', {
       flag: true,
-      help: 'Use USB connection'
+      help: 'Use only a USB connection'
+    })
+    .option('lan_prefer', {
+      flag: true,
+      default: false,
+      help: 'Prefer a LAN connection if it\'s available, otherwise use USB'
     });
 }
 
@@ -94,7 +99,7 @@ makeCommand('restart')
   })
   .option('entryPoint', {
     position: 1,
-    help: 'The program entry point file to deploy to Tessel.',
+    help: 'The entry point file to deploy to Tessel'
   })
   .option('type', {
     default: 'ram',
@@ -110,7 +115,7 @@ makeCommand('run')
   .option('entryPoint', {
     position: 1,
     required: true,
-    help: 'The program entry point file to deploy to Tessel.'
+    help: 'The entry point file to deploy to Tessel'
   })
   .option('single', {
     flag: true,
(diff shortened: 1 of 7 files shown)


bin/tessel-2.js
```diff
@@ -21,11 +21,16 @@ function makeCommand(commandName) {
...
       flag: true,
-      help: 'Use LAN connection'
+      help: 'Use only a LAN connection'
     })
     .option('usb', {
       flag: true,
-      help: 'Use USB connection'
+      help: 'Use only a USB connection'
+    })
+    .option('lanPrefer', {
```
johnnyman727: This is like the `--lan` option but it's not strict. The CLI will use a LAN connection if one is found but will fallback to USB w/o complaining.

rwaldron: This is just nit-picky but I think cli programs should avoid camel cased parameters

johnnyman727: Makes sense! I'll update the PR.

Frijol:
@johnnyman727 are you looking for review on this?

johnnyman727:
@Frijol I believe @rwaldron already reviewed but I wasn't sure if I should merge b/c this change might be controversial. I haven't heard any arguments in favor or against being more opinionated on which connection to use.

johnnyman727:
This branch has been rebased on top of #437 and should be merged as soon as #437 is.

lib/controller.js
```diff
@@ -239,17 +239,28 @@ Tessel.get = function(opts) {
...
       }
 
+
+      function finishSearchEarly(tessel) {
+        // Remove this listener because we don't need to search for the Tessel
+        seeker.removeListener('end', searchComplete);
+        // Stop searching
+        seeker.stop();
+        // Send this Tessel back to the caller
+        logAndFinish(tessel);
```
rwaldron: I missed this before, but the seeker object should be passed as an argument. It's not urgent

johnnyman727: @rwaldron `seeker` should be passed as an argument to `logAndFinish`? The definition of `logAndFinish` looks like it just accepts one argument (a Tessel Object).

rwaldron: Looks like I put this comment on the wrong line—it was meant for this line: 

```diff
+      function finishSearchEarly(tessel) {
```

johnnyman727: @rwaldron does it need that even though `seeker` is within that function's scope already?

rwaldron: > does it need that even though seeker is within that function's scope already?

I actually hadn't looked at the entire file to see the context around the declaration and call site. Seeing the entire thing, it looks fine. Sorry for the noise there.

johnnyman727: :+1: No worries - better to be alert than lazy.

PR #425, opened 2015-10-31
[WIP] Adds ability to disable wifi from CLI

### LINKED ISSUE #394: turn wifi on and off
Per spec:

`>  --off:       disconnect from a wifi connection`

*moved from https://github.com/tessel/t2-cli/issues/82*

johnnyman727:
Fixes #394

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -184,8 +184,11 @@ makeCommand('wifi')
     // TODO: Refactor switch case into controller.wifi
     if (opts.list) {
       callControllerWith('printAvailableNetworks', opts);
-    } else if (opts.off) {
-      callControllerWith('disableWifi', opts);
+    } else if (opts.off || opts.on) {
+      if (opts.off) {
+        opts.on = false;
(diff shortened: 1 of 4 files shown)


bin/tessel-2.js
```diff
@@ -183,6 +185,10 @@ makeCommand('wifi')
     metavar: 'PASSWORD',
     help: 'Set the password of the network to connect to'
   })
+  .option('off', {
+    flag: true,
+    help: 'Disable the wireless network'
```
HipsterBrown: Would it be more accurate to say "Disable the wireless network connection" ? This is more picky, semantics than anything.

HipsterBrown:
Looks good to me. :+1:

bin/tessel-2.js
johnnyman727: That's a good question. I actually haven't tried this with an access point enabled. It may cancel both station and access point connections!

rwaldron:
:+1:

PR #427, opened 2015-11-02
Do not include empty lines or comment lines in the patterns extracted from .tesselignore (matchs fstream-ignore semantics).

rwaldron:


rwaldron:
cc @johnnyman727 this is pretty straight forward

johnnyman727:
yep, looks good to me!

PR #438, opened 2015-11-14
Use os.EOL where necessary.

rwaldron:


rwaldron:
This was just a test!

Issue #441, opened 2015-11-14
Slim build construction can fail silently if script has syntax error


johnnyman727:
My deploys were hanging after 'generating slim build'. I had a feeling that some rejection was not being handled properly so I put a few console.logs in the `tarBundle` function. It turns out [this error was being rejected](https://github.com/tessel/t2-cli/blob/master/lib/tessel/deploy.js#L439):
```
error on bundle... { [Error: Parsing file /Users/Jon/Work/technical/modules/ambient-attx4/examples/ambient.js: Unexpected token (16:5)]
  stream:
   Labeled {
     _readableState:
      ReadableState {
        highWaterMark: 16,
        buffer: [],
        length: 0,
        pipes: [Object],
        pipesCount: 1,
        flowing: true,
        ended: false,
        endEmitted: false,
        reading: true,
        sync: false,
        needReadable: true,
        emittedReadable: false,
        readableListening: false,
        objectMode: true,
        defaultEncoding: 'utf8',
        ranOut: false,
        awaitDrain: 0,
        readingMore: false,
        decoder: null,
        encoding: null,
        resumeScheduled: false },
     readable: true,
     domain: null,
     _events:
      { end: [Object],
        error: [Object],
        data: [Function: ondata],
        _mutate: [Object] },
     _eventsCount: 4,
     _maxListeners: undefined,
     _writableState:
      WritableState {
        highWaterMark: 16,
        objectMode: true,
        needDrain: false,
        ending: true,
        ended: true,
        finished: true,
        decodeStrings: true,
        defaultEncoding: 'utf8',
        length: 0,
        writing: false,
        corked: 0,
        sync: false,
        bufferProcessing: false,
        onwrite: [Function],
        writecb: null,
        writelen: 0,
        buffer: [],
        pendingcb: 0,
        prefinished: true,
        errorEmitted: false },
     writable: true,
     allowHalfOpen: true,
     _options: { objectMode: true },
     _wrapOptions: { objectMode: true },
     _streams: [ [Object] ],
     length: 1,
     label: 'deps' } }
```

This was caused because my script had a syntax error.

PR #444, opened 2015-11-15
Ensure that bundling errors make are surfaced. Fixes gh-441

rwaldron:


lib/tessel/deploy.js
```diff
@@ -436,7 +436,10 @@ actions.tarBundle = function(opts) {
         })
         .bundle(function(error, results) {
           if (error) {
-            return reject(error);
+            return reject({
```
johnnyman727: A bunch of other places in the code return a string for a warning and an actual Error for a more stern error output. Why not return an Error here?

rwaldron: Because those are all taking the easy path that automatically logs with `logs.warn`, not `logs.err`. I explicitly wanted `logs.err` and this was the only way to force it to happen, because: https://github.com/tessel/t2-cli/blob/e7299f0b85862c01a10158ef92ec000ede42d07f/bin/tessel-2.js#L290-L307

HipsterBrown:
This makes browserify even more helpful for warning developers about syntax errors before pushing to Tessel. Looks good to me. :+1:

Issue #449, opened 2015-11-16
Stop using new Error(...) in reject()


rwaldron:
When `reject(new Error(...))` is called, `closeFailedCommand` (in bin/tessel-2.js) does this: 

```
  if (err instanceof Error) {
    throw err;
  } else {
```

Which means this happens: 

```
Unhandled rejection Error: blah blah blah
```

...Followed by the entire stack trace. 

Passing a string will log the string as a warning.

rwaldron:
Or should it just use `logs.err` when the argument is actually a `new Error()` object?

rwaldron:
Here's an example, where instead of rethrowing and doing absolutely nothing with it, t2 will log an error instead. 

Before: 

```
$ t2 provision
Unhandled rejection Error: No key provided to set as default.
    at /Users/rwaldron/clonez/t2-cli/lib/tessel/provision.js:160:21
    at Object.actions.setDefaultKey (/Users/rwaldron/clonez/t2-cli/lib/tessel/provision.js:157:10)
    at Function.Tessel.get (/Users/rwaldron/clonez/t2-cli/lib/controller.js:121:20)
    at /Users/rwaldron/clonez/t2-cli/lib/controller.js:280:19
    at tryCatcher (/Users/rwaldron/clonez/t2-cli/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._resolveFromResolver (/Users/rwaldron/clonez/t2-cli/node_modules/bluebird/js/main/promise.js:480:31)
    at new Promise (/Users/rwaldron/clonez/t2-cli/node_modules/bluebird/js/main/promise.js:70:37)
    at Object.controller.standardTesselCommand (/Users/rwaldron/clonez/t2-cli/lib/controller.js:278:10)
    at executeProvision (/Users/rwaldron/clonez/t2-cli/lib/controller.js:517:25)
    at tryCatcher (/Users/rwaldron/clonez/t2-cli/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/rwaldron/clonez/t2-cli/node_modules/bluebird/js/main/promise.js:507:31)
    at Promise._settlePromiseAt (/Users/rwaldron/clonez/t2-cli/node_modules/bluebird/js/main/promise.js:581:18)
    at Promise._settlePromiseAtPostResolution (/Users/rwaldron/clonez/t2-cli/node_modules/bluebird/js/main/promise.js:245:10)
[...]

PR #450, opened 2015-11-16
Instead of rethrowing errors, log them. Fixes gh-449

rwaldron:


bin/tessel-2.js
```diff
@@ -287,18 +287,27 @@ module.exports.closeSuccessfulCommand = function() {
...
-module.exports.closeFailedCommand = function(opts, err) {
-  if (!err) {
-    err = opts;
-    opts = {};
-  }
-  if (err instanceof Error) {
-    throw err;
+module.exports.closeFailedCommand = function(opts, error) {
+
+  if (opts) {
```
HipsterBrown: Is there a case where this would ever get called without at least `opts` passed in?

HipsterBrown:
Besides that one comment, it looks good to me. It will be nice to have some better error feedback from every command.

bin/tessel-2.js
rwaldron: There were tests in `test/unit/bin-tessel-2.js` that broke when I assumed that couldn't happen. I'm going to investigate.

rwaldron: I found no occurrences where `closeFailedCommand` is called with nothing, so I'm re-working this. I was also able to clean up the syntax error handling from https://github.com/tessel/t2-cli/commit/0726002569bb3d043e0b22edb2a5ac51e86b304c

HipsterBrown: Awesome! Simpler code is happier code. :smile:

bin/tessel-2.js
```diff
@@ -40,12 +40,7 @@ function makeCommand(commandName) {
 
 function callControllerWith(methodName, opts) {
   return controller[methodName](opts)
-    .then(module.exports.closeSuccessfulCommand, function(opts, err) {
-      if (typeof opts === 'object' && opts !== null) {
-        err = opts.message || '';
-      }
-      module.exports.closeFailedCommand(opts, err);
-    });
+    .then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
```
HipsterBrown: This is so much more approachable to follow. :clap:

HipsterBrown:
Fantastic refactor. :+1:

rwaldron:
@HipsterBrown thanks for the review (x2)

PR #451, opened 2015-11-19
Tests for --key option

Student007:
Validates whether no key would cause setting default key as known
before. Also tests non existing key and alternate (generated) key.

### DIFF
--- a/test/unit/key.js
+++ b/test/unit/key.js
@@ -1,0 +1,138 @@
+var sinon = require('sinon');
+var logs = require('../../lib/logs');
+var testDir = __dirname + '/tmp/';
+var testFile = 'test_rsa';
+var path = require('path');
+var testPath = path.join(testDir, testFile);
+var provision = require('../../lib/tessel/provision');
+var osenv = require('osenv');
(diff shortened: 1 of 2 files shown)


test/unit/key.js
```diff
@@ -0,0 +1,138 @@
...
+        test.equal(fs.statSync.firstCall.exception.code, 'ENOENT');
+        test.done();
+      });
+  },
+
+  validAlternateKey: function(test) {
+    var self = this;
+    test.expect(4);
+    // Create folders for the folder that we'd like to
+    createKeyTestFolder(function(err) {
```
HipsterBrown: Instead of creating a test folder on the fly, could we add a test folder to the `fixtures` directory? It might speed up the testing process and reduce the complexity of this test setup.

rwaldron: @Student007 hold off on making any sweeping changes for this, I just pinged @HipsterBrown to ask him about doing that change in another issue—will post back asap.

rwaldron: https://github.com/tessel/t2-cli/issues/452

rwaldron:
Great work @Student007!

PR #453, opened 2015-11-24
Adds root command

johnnyman727:


lib/controller.js
```diff
@@ -705,6 +705,38 @@ controller.disableAccessPoint = function(opts) {
...
+        }
+        // Stream output to the console
+        stream.stdout.pipe(process.stdout);
+        // Stream errors to the console
+        stream.stderr.pipe(process.stderr);
+        // Pipe input to the remote process
+        process.stdin.pipe(stream.stdin);
+
+        // Once the session completes
+        stream.on('close', function() {
```
reconbot: it's just a nitpick but you can pass resolve directly to the event emitter `stream.on('close', resolve)`

rwaldron:
@johnnyman727 I had some time this morning, so I wrote some tests to help move this along faster :) https://github.com/tessel/t2-cli/pull/454

johnnyman727:
Thanks so much @rwaldron!

johnnyman727:
Also, rebasing this PR is a beast since the original commit is so old. I'd prefer to sacrifice history hygiene to get more work done today.

rwaldron:
@johnnyman727 just squash it into a single commit first

johnnyman727:
@rwaldron how would you recommend doing that? If I run `git rebase -i origin/master` and replace all but the first commit with `squash`, I still have to update four different files with work I already did in `8887dfe`.

rwaldron:
@johnnyman727 

> how would you recommend doing that?

So... I just tried and I don't think this can be done cleanly (as you've already learned).

johnnyman727:
:+1:

PR #454, opened 2015-11-24
root: tests

rwaldron:


test/unit/controller.js
```diff
@@ -838,3 +838,114 @@ exports['controller.closeTesselConnections'] = {
...
+      };
+      this.stderr = {
+        pipe: sandbox.spy()
+      };
+      this.stdin = {
+        pipe: sandbox.spy()
+      };
+    }
+
+    S.prototype = Object.create(Emitter.prototype);
```
johnnyman727: I've always wondered how to assign the Emitter properties to a single object instance!

test/unit/controller.js
```diff
@@ -838,3 +838,114 @@ exports['controller.closeTesselConnections'] = {
...
 };
+
+exports['controller.root'] = {
+  setUp: function(done) {
+    this.sandbox = sinon.sandbox.create();
+    this.logsWarn = this.sandbox.stub(logs, 'warn', function() {});
+    this.logsInfo = this.sandbox.stub(logs, 'info', function() {});
+    this.logsBasic = this.sandbox.stub(logs, 'basic', function() {});
+
+    this.standardTesselCommand = this.sandbox.stub(controller, 'standardTesselCommand').returns(Promise.resolve());
```
johnnyman727: @rwaldron doesn't this need to resolve with a Tessel simulator?

johnnyman727:
@rwaldron thanks for putting this together. I got started on tests but didn't get far so this was very useful.

test/unit/controller.js
```diff
@@ -838,3 +838,114 @@ exports['controller.closeTesselConnections'] = {
...
+    this.sandbox.restore();
+    done();
+  },
+
+  setsOptsLanTrue: function(test) {
+    test.expect(3);
+
+    var opts = {};
+
+    controller.root(opts)
```
johnnyman727: I'm a little surprised `root` doesn't `reject` because `standardTesselCommand` doesn't return a Tessel.

johnnyman727: Oh I see - because `standardTesselCommand` returns Promise.resolve (instead of an unfulfilled promise), the whole `root` procedure finishes up before it gets to the `shell` part.

johnnyman727:
Looks great!

test/unit/controller.js
rwaldron: :+1:

test/unit/controller.js
rwaldron: Use a :hammer: !

PR #456, opened 2015-11-24
Preemptive update for next node-usb release

johnnyman727:


rwaldron:
In the future, version bumping should be it's own commit that is not tied to a PR.

johnnyman727:
@rwaldron I'm fine with that but why should it not be in a PR?

rwaldron:
Sorry, I didn't mean "no PR at all", I meant "not tied to a PR for something else"—sorry that was definitely not clear in my first message.

johnnyman727:
Gotcha! Yeah, that seems totally reasonable.

PR #459, opened 2015-11-25
Fix t2 push default slimPath so app start finds it

### LINKED ISSUE #458: Simple t2 push _filename.js_ causes pushed file not to run
When using `t2 push _filename_.js`, the resulting pushed file never runs on the t2.  The file laid down in this case is /app/__tessel_program__.js but both /app/start and /etc/init.d/tessel-app just do a `cd app` and then `exec node .` which errors because there is no index.js in /app anymore.

    host_box:my_puck jonathan$ t2 push blinky.js
    INFO Connected to mypuck-1 over LAN
    INFO Generating slim build.
    INFO Writing project to Flash on mypuck-1 (2.048 kB)...
    INFO Deployed.
    INFO You may now disconnect from the Tessel. Your code will be run whenever Tessel boots up. To remove this code, use `tessel erase`.
    INFO Running __tessel_program__.js...
    host_box:my_puck jonathan$ !ssh
    ssh -i ~/.tessel/id_rsa root@10.10.10.1
    
    
    BusyBox v1.23.2 (2015-11-24 22:57:01 UTC) built-in shell (ash)
    
    Tessel 2  /  Built on OpenWrt
    root@mypuck-1:~# cd /app
    root@mypuck-1:/app# ls
    __tessel_program__.js  start
    root@mypuck-1:/app# cat start
    #!/bin/sh
    cd /app
    exec node .
    root@mypuck-1:/app# node .
    module.js:339
        throw err;
        ^
    
    Error: Cannot find module '/app'
        at Function.Module._resolveFilename (module.js:337:15)
        at Function.Module._load (module.js:287:25)
        at Function.Module.runMain (module.js:467:10)
        at startup (node.js:134:18)
        at node.js:961:3
    root@mypuck-1:/app# cat /etc/init.d/tessel-app
    #!/bin/sh /etc/rc.common
    
    START=99
    USE_PROCD=1
    start_service() {
    	procd_open_instance
    	procd_set_param command /app/start
    	procd_close_instance
    }
    root@mypuck-1:/app#

jonathana:
When doing a slim build, the resulting script should be laid down as index.js so that when /app/start does its `exec node .` there is an index.js in place.  This is the simplest solution to attempt to resolve #458

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -167,7 +167,7 @@ makeCommand('push')
     help: 'Bundle only the required modules'
   })
   .option('slimPath', {
-    default: 'build.js'
+    default: 'index.js'
   })
   .help('Pushes the file/dir to Flash memory to be run anytime the Tessel is powered, runs the file immediately once the file is copied over');
 


rwaldron:
Thanks for the patch! I've landed it already, which I realize I shouldn't have done yet, because I hadn't confirmed that you've signed the CLA! If you could do that for me, I'd appreciate it: http://dojofoundation.org/about/claForm

jonathana:
CLA done.

johnnyman727:
:clap: Thanks @jonathana!

LinusU:
@HipsterBrown @rwaldron Is there any reason for keeping the `slimPath` at all? What purpose does it server? Won't any other value than `index.js` make it undeployable. I think that we should drop this option.

johnnyman727:
+1

jonathana:
@LinusU in the latest version /app/start is templated and has the value written in.  See #461.

LinusU:
That's nice, but my questions are still valid. Why is it even needed in the first place?

rwaldron:
It might not be, I can explore after I'm done with binary modules

PR #463, opened 2015-11-25
Drivers: Gracefully exit install-drivers operation

rwaldron:


johnnyman727:
Confirmed that it fixes the issue. If Travis gives the :+1:, let's merge.

PR #466, opened 2015-11-25
t2 root remote

Student007:
This commit adds the ability to connect to a Tessel is accessible by
port forwarding.
Tests are covered by @rwaldron awesome using a hammer test :wink:

In general all common ssh options are implemented.

Controller.js got the following new or changed objects:

|  Object | return | Description |
|---|---|---|
| controller.clear() | promise | clears terminal on mac, windows and linux |
| controller.remote(tessel, opts) | promise | ssh connection with advanced options, error handling and port forwarding support. The `tessel` option is used to get `tessel.name` for creating event handlers of the child process |
| controller.root(opts) | promise | conditional switch on `opts.ip` for connecting Tessels outside the current netmask (mDNS limit) |

### DIFF
--- a/lib/usb/usb_process.js
+++ b/lib/usb/usb_process.js
@@ -10,10 +10,12 @@ var protocol = require('usb-daemon-parser');
 // ...
 
 
-// The largest packet we can send over USB
-var MAX_PACKET_SIZE = 255;
+// The largest packet allowed is 255 bytes
+// We make the data 251 bytes so we can concat
+// the 4 header bytes to it and use only 1 packet
+var MAX_DATA_PACKET_SIZE = 4092;
 // The size of the back pressure buffers
-var MAX_BUFFER_SIZE = 4096;
+var MAX_BUFFER_SIZE = 32768;
 
 // A process is running remotely on the Tessel and data is piped in over USB
 function USBProcess(id, daemon) {
@@ -195,16 +197,18 @@ RemoteWritableStream.prototype._drainBackPressure = function() {
     entry.buffer = entry.buffer.slice(chunk.length);
 
     // Split the chunk into smaller packets
-    var indices = self._packetize(MAX_PACKET_SIZE, chunk);
+    var indices = self._packetize(MAX_DATA_PACKET_SIZE, chunk);
 
     // For each packet
     indices.forEach(function(index) {
       // Slice out the data packet
       var buffer = chunk.slice(index.start, index.end);
-      // Send the header for our control write
-      self.daemon.write(self.writeHeaderFunc(self.process.id, buffer.length));
+      // Create a header for the write
+      var header = self.writeHeaderFunc(self.process.id, buffer.length);
+      // Compact our header and data into one packet
+      var compact_buffer = Buffer.concat([header, buffer]);
       // Send over the data to the control stream
-      self.daemon.write(buffer);
+      self.daemon.write(compact_buffer);
     });
 
     // If we sent all of the remaining bytes of this write
@@ -288,13 +292,15 @@ RemoteReadableStream.prototype.ack = function(numToAck) {
   // Add the number to the credit count
   this.credit += numToAck;
   // Allocate a buffer for the data
-  var b = new Buffer(4);
+  var ackAmountBuf = new Buffer(4);
   // Write the number as an unsigned 32 but int into the buffer
-  b.writeUInt32LE(numToAck, 0);
-  // Write the header
-  this.daemon.write(this.ackHeaderFunc(this.process.id, b));
-  // Write the data
-  this.daemon.write(b);
+  ackAmountBuf.writeUInt32LE(numToAck, 0);
+  // Create a header for this ack
+  var header = this.ackHeaderFunc(this.process.id, ackAmountBuf);
+  // Combine the header and the ack amount
+  var packet = Buffer.concat([header, ackAmountBuf]);
+  // Write the entire packet
+  this.daemon.write(packet);
 };
 
 // A function to close the stream
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "uglify-js": "^2.5.0",
     "url-join": "0.0.1",
     "usb": "1.1.1",
-    "usb-daemon-parser": "0.0.1"
+    "usb-daemon-parser": "0.0.2"
   },
   "preferGlobal": true,
   "engines": {


lib/lan_connection.js
```diff
@@ -93,7 +92,7 @@ LAN.Connection.prototype.open = function() {
       resolve(self);
     });
     self.ssh.on('error', function() {
-      debug('Device open error:', self.host);
+      debug('Device open error:', self);
```
Student007: this is because host is not given by remote ssh connections

bin/tessel-2.js
```diff
@@ -324,11 +324,40 @@ makeCommand('ap')
 
 makeCommand('root')
   .callback(function(opts) {
+    opts.usb = false;
```
johnnyman727: Can you move this to `controller`? I already marked [`opts.lan = true`](https://github.com/tessel/t2-cli/blob/0772e4558bf264fb48389ed85953b103dd6c7059/lib/controller.js#L713) and we can be more explicit about USB there as well.

bin/tessel-2.js
```diff
@@ -324,11 +324,40 @@ makeCommand('ap')
...
     callControllerWith('root', opts);
   })
+  .option('key', {
```
johnnyman727: You don't need this option. It's already added with the `makeCommand` function.

bin/tessel-2.js
```diff
@@ -324,11 +324,40 @@ makeCommand('ap')
...
+  .option('port', {
+    abbr: 'p',
+    full: 'port',
+    metavar: 'PORT',
+    default: 22,
+    help: 'Specific PORT to use instead of default port 22'
+  })
   .help('Gain SSH root access to one of your authorized tessels');
 
-
```
johnnyman727: Can you add this whitespace back in?

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
+      }
+    });
+  });
+};
+
 /*
   The T2 root command is used to login into the Tessel's root shell.
 */
 controller.root = function(opts) {
+
```
johnnyman727: Please remove this whitespace modification

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
   // Only give us LAN connections
   opts.lan = true;
+
```
johnnyman727: Same as above

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
+      .spawn('ssh', options, {
+        stdio: 'inherit'
+      });
+
+    // error handling
+    ch.once('error', function(e) {
+      if (e === 255) {
+        controller.clear()
+          .then(function() {
+            logs.warn('Login to Tessel at failed with KEY ' + opts.key + ' !');
```
johnnyman727: Missing the IP Address in this warning (I assume "Tessel at ... failed")

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
+    ];
+
+    var ch = require('child_process')
+      .spawn('ssh', options, {
+        stdio: 'inherit'
+      });
+
+    // error handling
+    ch.once('error', function(e) {
+      if (e === 255) {
```
johnnyman727: Can you make constants for each of these numbers?

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
   });
 };
 
+// clear terminal windows and linux
+controller.clear = function() {
+  return new Promise(function(resolve) {
+    // clears terminal
+    if (!global.IS_TEST_ENV) {
+      process.stdout.write('\u001b[2J\u001b[0;0H');
```
johnnyman727: What is this string?

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
+      tessel.connection.open()
+        .then(function() {
+          if (tessel.lanConnection.authorized) {
+            tessel.getName()
+              .then(function() {
+                controller.clear().then(function() {
+                  controller.remote(tessel, opts)
+                    .then(resolve).catch(function() {
+                      resolve();
+                    });
```
johnnyman727: You should be able to clean this up:
```.js
tessel.getName()
.then(controller.clear)
.then(controller.remote.bind(self, tessel, opts)
.then(resolve)
.catch(resolve);
```

Be sure to add `var self = this` at the top if you do that.

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
+          });
+      } else if (e === 255) {
+        controller.clear()
+          .then(function() {
+            logs.warn('You are not authorized with key ' + opts.key);
+          }).then(resolve);
+      } else {
+        controller.clear()
+          .then(function() {
+            logs.warn('Connection to ' + tessel.name + ' closed due to reason (error code): ', e);
```
johnnyman727: I recommend changing the end of this string to `"closed:", e).` because `due to reason` doesn't make perfect sense.

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
+            logs.warn('Login to Tessel at failed with KEY ' + opts.key + ' !');
+          }).then(resolve);
+      } else {
+        reject('Error while connected to ' + tessel.name + ':', e);
+      }
+    });
+
+    // exit handling
+    ch.once('exit', function(e) {
+      if (e === 127 || e === 0) { // exit by user
```
johnnyman727: Can you make these numbers into constants as well?

lib/controller.js
```diff
@@ -705,39 +710,152 @@ controller.disableAccessPoint = function(opts) {
...
+        // clear console
+        controller.clear()
+          .then(function() {
+            logs.warn('Connection to ' + tessel.name + ' closed!\n');
+            resolve();
+          });
+      } else if (e === 255) {
```
johnnyman727: Same as above.

lib/controller.js
Student007: http://stackoverflow.com/questions/22488850/how-can-i-clear-a-windows-7-console-screen-between-function-calls-in-node-js/22494170#22494170

lib/controller.js
Student007: @johnnyman727 good point :+1:

LinusU: This is probably not the correct place to discuss this at all, but still. Since we already using stuff that only works in Node >= 4, shouldn't we use fat-arrow for binding this instead of `self = this`?

```javascript
tessel.getName()
.then(controller.clear)
.then(() => controller.remote(tessel, opts))
.then(resolve)
.catch(reject); // Changed this to reject as well, correct?
```

lib/controller.js
Student007: could you give an example ? - I didn't got the point for what reason ..

lib/controller.js
Student007: @LinusU I would love to use more *ES6* style (for example using `let` instead of `var` all the time). The argument against this would be it is still a minority is familiar with this I assume. But I would prefer this as I said :+1:

Student007: @LinusU got this :stuck_out_tongue_winking_eye: 
```
Fatal error: /Users/{anonym}/github/t2-cli/lib/controller.js:822
              .then(() => controller.remote(tessel, opts))
                     ^
Unexpected token )
```

LinusU: Hmm, is it the linting? Probably need to update eslint or plug in babel-eslint...

Student007: maybe it is `node --version` => `v1.6.2` :laughing:

rwaldron:
@linusu JSHint supported ArrowFunction grammar parsing before babel-eslint even existed. That error wasn't from linting.

Student007:
@johnnyman727 what happend ? I didn't something ... ?!?!?

PR #469, opened 2015-11-26
Fixes #455

### LINKED ISSUE #455: CLI errors out on any command if no key in ~/.tessel
Reproduction:
* Remove any existing `~/.tessel` folder
* Run `t2 list` (or any CLI command)

Error:
```
➜  t2-cli git:(jon-root) ✗ t2 list
ERR! Error: ENOENT: no such file or directory, stat '/Users/Jon/.tessel/id_rsa'
```

johnnyman727:
This looks like a bigger deal than it really is in the diffs. I simply removed the SSH key path as a default option for commands and added a wrapper around `Tessel.get` and `Tessel.list` so it was easier to test the change.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -24,273 +24,276 @@ var responses = {
   auth: 'No Tessels Found.'
 };
 
-Tessel.list = function(opts) {
+// Wrapper function for Tessel.list to set SSH key path
+controller.list = function(opts) {
+  return controller.keyHelper(opts, Tessel.list);
+};
+
+// Wrapper function for Tessel.get to set SSH key path
+controller.get = function(opts) {
+  return controller.keyHelper(opts, Tessel.get);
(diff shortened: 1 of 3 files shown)


lib/controller.js
```diff
@@ -24,273 +24,276 @@ var responses = {
...
+
+controller.keyHelper = function(opts, next) {
   var keyPromise = Promise.resolve();
   if (opts.key) {
     // Set the default SSH key path before we search
     keyPromise = provision.setDefaultKey(opts.key);
   }
 
-  return keyPromise.then(function() {
+  return keyPromise.then(next.bind(this, opts));
```
rwaldron: How about...

```js
return keyPromise.then(() => next(opts));
```

LinusU: :+1:

PR #476, opened 2015-12-06
Add better help text for `key generate`

reconbot:
and add `required: true` on `ap --ssid`

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -247,11 +247,6 @@ makeCommand('wifi')
   .help('Configure the wireless connection');
 
 parser.command('key')
-  .option('method', {
-    position: 1,
-    required: true,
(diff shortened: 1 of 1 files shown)


HipsterBrown:
@reconbot That's correct. Everything else either has a default or not required. :)

bin/tessel-2.js
```diff
@@ -247,11 +247,6 @@ makeCommand('wifi')
   .help('Configure the wireless connection');
 
 parser.command('key')
-  .option('method', {
-    position: 1,
-    required: true,
-    choices: ['generate'],
```
johnnyman727: I originally had it this way because I was going to add other choices like `remove`, `reset`, etc. but lost interest. I think this change looks fine.

PR #481, opened 2015-12-08
[WIP] Close connections on SIGINT during Tessel.get or Tessel.list

johnnyman727:
Currently, if the user aborts a script during `t2 list` or at the beginning of any other command that fetches a Tessel (ie `.Tessel.get` is in process), the script will exit without terminating the USB Connection. On Linux, the OS automatically releasing any interfaces when the parent process dies (source: @kevinmehall) but that does not seem to be the case with OSX. That means we need to be extra sure that we close the USB connection regardless of whether the script ends 'naturally' or not.

If we do not end the connections, the USB Daemon on Tessel 2 will keep all of those processes spawned and the next time the CLI tries to open a process, it will likely fail because it will try to create a new process id of a process that's already running (from the last time).

This PR adds checks:
- within `Tessel.get`
- within `Tessel.list`
- for any Tessel that we've discovered but whose `.open` command has not yet completed
- for any USB connection that we've opened but is currently unresponsive (usually because there was an issue with the USB Daemon)

Currently working on tests.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -41,6 +41,15 @@ Tessel.list = function(opts) {
       // Keep a list of all the Tessels we discovered
       var foundTessels = [];
 
+      // The user has aborted, we need to close any open connections
+      process.once('SIGINT', function() {
+        // Stop searching for new Tessels
+        seeker.stop()
+          // Close any opened Tessel connections
+          .then(() => controller.closeTesselConnections(foundTessels))
+          .then(resolve);
+      });
+
       // Options for  Tessel discovery
       var seekerOpts = {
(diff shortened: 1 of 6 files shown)


johnnyman727:
Just another thought I had: we might be able to lessen the risk of the USB interface not being released by creating a control transfer endpoint that basically tells the microcontroller to 'disable' the port which will ultimately result in the USB Daemon wiping any pre-existing processes. @kevinmehall what do you think about that?

johnnyman727:
Superseded by #482.

This does make somethings cleaner (like being sure to free all connections on a Tessel when `.close` is called) but I don't have time to write tests right now.

Issue #345, opened 2015-09-27
Aborting the deployment of a large (> several kB) script, T2 drops off the bus


johnnyman727:
This is likely the USB Daemon crashing. 

Example below (I ran `t2 init` and then `npm install twitter` to get a sufficiently large bundle):
```
➜  t2-demo  t2 run index.js --name Frank
INFO Looking for your Tessel...
INFO Connected to Frank over USB
INFO Writing index.js to RAM on Frank (3482.624 kB)...
^C^C
➜  t2-demo  t2 list
INFO Searching for nearby Tessels...
	Frank	LAN
```

johnnyman727:
Confirmed that restarting the USB Daemon resolves the issue. I'll dive more into the root cause later.

johnnyman727:
The issue here is that control+c is not actually being caught and cancelling the transfer. When you hit control+c the second time, it closes the CLI process without closing the remote process.

johnnyman727:
Another note for myself: control+c is being caught, and the process attempts to close, but for some reason it refuses to close until it finishes writing all data backpressured in `stdin`to the remote process. I'll need to look into our usb processes in the CLI.

johnnyman727:
Indeed, this can be fixed if I could cancel all pending `stdin.write`s but Node.js doesn't seem to have that capability for Streams. All of the calls to `stdin.write` to transfer the project get executed immediately, but the calls to the `_write` function take a while to process depending on the size of the project. I'd like to be able to just cancel all the pending and then send the command to T2 to close that process...

johnnyman727:
I also inquired about the feature request for Node here: https://github.com/nodejs/node/issues/3169

PR #482, opened 2015-12-09
Tells coprocessor to disable USB channel before claiming interface

johnnyman727:
Dependent on https://github.com/tessel/t2-firmware/pull/136, https://github.com/tessel/t2-firmware/pull/134, and https://github.com/tessel/t2-firmware/pull/103.

It will improve USB robustness overall and fixes https://github.com/tessel/t2-cli/issues/345.

### DIFF
--- a/lib/usb_connection.js
+++ b/lib/usb_connection.js
@@ -119,39 +119,58 @@ USB.Connection.prototype.open = function() {
       return reject(e, self);
     }
     // Set interface settings
-    self.intf.setAltSetting(2, function(error) {
-      if (error) {
-        return reject(error, self);
-      }
-      self.epIn = self.intf.endpoints[0];
-      self.epOut = self.intf.endpoints[1];
-      if (!self.epIn || !self.epOut) {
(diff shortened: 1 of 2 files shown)


lib/usb_connection.js
```diff
@@ -14,7 +14,8 @@ var debug = require('debug')('discovery:usb');
 var DFU = require('./dfu');
 var logs = require('./logs');
 
-
+const outVendorcontrolTransferType = 0x40;
+const requestResetUSBD = 0x02;
```
rwaldron: What do you think about sticking to writing `const` identifiers in the "traditional" way, eg `ALL_CAPS_WITH_UNDERSCORES`? We could probably set up a jscs rule for it

LinusU: +1 from me, I think it very clearly shows the intention

johnnyman727: Yeah, I'm good with that. I'll update this PR before merging. Need to do some more testing first.

PR #486, opened 2015-12-12
Speeds up USB communication

johnnyman727:
By combining headers and data into a single packet, we can dramatically reduce the number of packets sent and roundtrips between the CLI->mcu->spid->usbd and back.

eg. Instead of sending a 4 byte header and then 255 bytes of data, concatenate the 4 header bytes with 251 data bytes.

This also increases the maximum write sizes to the coprocessor.

Initial benchmarks indicate a 2.83x speed increase.

Depends on https://github.com/tessel/t2-firmware/pull/130 and https://github.com/tessel/usb-daemon-parser/pull/2.

### DIFF
--- a/lib/usb/usb_process.js
+++ b/lib/usb/usb_process.js
@@ -10,10 +10,12 @@ var protocol = require('usb-daemon-parser');
 // ...
 
 
-// The largest packet we can send over USB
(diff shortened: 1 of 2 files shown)


johnnyman727:
Will re-open on master because something weird happened.

PR #488, opened 2015-12-13
toggle wifi-iface instead of wifi-device

### LINKED ISSUE #487: Turning off wifi disables wireless radio
At the moment, [`commands.turnOnWifi()`](https://github.com/tessel/t2-cli/blob/master/lib/tessel/commands.js#L55-L57) disables the `wifi-idevice` `radio0` instead of the 'wifi' `wifi-iface`. This should be changed so that the `lan` access point `wifi-iface` can continue running when wifi is turned off.

HipsterBrown:
Fixes #487

### DIFF
--- a/lib/tessel/commands.js
+++ b/lib/tessel/commands.js
@@ -53,7 +53,7 @@ module.exports.setNetworkEncryption = function(encryption) {
   return ['uci', 'set', 'wireless.@wifi-iface[0].encryption=' + encryption];
 };
 module.exports.turnOnWifi = function(enabled) {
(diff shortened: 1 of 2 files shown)


johnnyman727:
@hipsterbrown, this is great. Could you add one or more unit tests that confirms that this command is called correctly?

HipsterBrown:
@johnnyman727 Sure. Making sure it's called? Or that it's calling the right `uci` setting?

johnnyman727:
Mostly the latter

HipsterBrown:
Got it. I'll work on it.

HipsterBrown:
@johnnyman727 Tests added. :)

johnnyman727:
Perfect!

PR #489, opened 2015-12-14
[WIP] Mirrors ssh2 lib hadError arg in close event

### LINKED ISSUE #475: First use of `t2 ap`, if over USB, will fail
I updated a fresh T2 up to `0.0.6`. Note that the wireless config files (related to `t2 ap`) won't actually be overwritten by OpenWRT going from (sub `0.0.6`, when we add the functionality, to `0.0.6`) so you'd have to [manually reset them to their original default state](https://github.com/tessel/openwrt-tessel/tree/f64d0046313bcd6855e37a999213bfb90f75a03e/files/etc/config) if you've already used `t2 ap`).

Running `t2 ap -n SSID -p PASS` results in a failure if it's run over USB:
```
➜  ambient-attx4 git:(jon-t2-compat) ✗ t2 ap -n Frank_Net -p Frank_Pass --usb
INFO Connected to Tessel-02A3E843B7A8.
events.js:141
      ^

Error: Error upon remote process exit: 1
    at USBProcess.<anonymous> (/Users/Jon/Work/technical/t2-cli/lib/usb/usb_process.js:56:26)
    at USBProcess.g (events.js:260:16)
    at emitOne (events.js:77:13)
    at USBProcess.emit (events.js:169:7)
    at Parser.<anonymous> (/Users/Jon/Work/technical/t2-cli/lib/usb/usb_daemon.js:127:14)
    at emitOne (events.js:77:13)
    at Parser.emit (events.js:169:7)
    at Parser._write (/Users/Jon/Work/technical/t2-cli/node_modules/usb-daemon-parser/index.js:152:16)
    at doWrite (_stream_writable.js:292:12)
    at writeOrBuffer (_stream_writable.js:278:5)
    at Parser.Writable.write (_stream_writable.js:207:11)
    at ondata (_stream_readable.js:528:20)
    at emitOne (events.js:77:13)
    at emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
    at Readable.push (_stream_readable.js:110:10)
```

The reason is because the CLI attempts to fetch the [SSID of the access point network](https://github.com/tessel/t2-cli/blob/master/lib/tessel/access-point.js#L130) which, since this is a fresh install, doesn't exist yet. If you run the command on T2 itself, you'll see that it emits an error message and the error code is non-zero:
```
root@Tessel-02A3E843B7A8:~# uci get wireless.@wi
fi-iface[1].ssid
uci: Entry not found

root@Tessel-02A3E843B7A8:~# $?
-ash: 1: not found
```

The USB Process that's running the command [notices the exit code of 1, interprets it as an error](https://github.com/tessel/t2-cli/blob/master/lib/usb/usb_process.js#L54-L570) and because there is nothing there to catch it, the error is thrown and process is aborted.

The question here is how the USBProcess *should* handle an error. I'm not convinced just emitting an error is the right way to go.

Also, the Access Point logic will need to check for a non-existent SSID.

johnnyman727:
Fixes #475.

### DIFF
--- a/lib/usb/usb_daemon.js
+++ b/lib/usb/usb_daemon.js
@@ -133,7 +133,7 @@ function USBDaemon() {
       var proc = entry.processes[packet.pid];
       // If it exists
       if (proc) {
-        proc.emit('close');
+        proc.emit('close', proc.exitedWithError);
       }
     });
 
--- a/lib/usb/usb_process.js
+++ b/lib/usb/usb_process.js
@@ -34,6 +34,8 @@ function USBProcess(id, daemon) {
   this.closed = false;
   // Boolean of whether this process has been forcibly killed
   this.forceKill = false;
+  // Code to be set upon process death
+  this.exitedWithError = false;
 
   // 4 remote process streams
   this.control = new RemoteWritableStream(self, daemon, protocol.controlWrite, protocol.controlClose);
@@ -55,7 +57,7 @@ function USBProcess(id, daemon) {
     // Check for a non-zero exit code
     var exitCode = parseInt(data.arg);
     if (exitCode !== 0 && !self.forceKill) {
-      self.emit('error', new Error('Error upon remote process exit: ' + exitCode));
+      self.exitedWithError = true;
     }
   });
 


johnnyman727:
Going to close and merge onto master.

Issue #497, opened 2015-12-29
No default Key for root command (if multiple Tessels are plugged)


Student007:
When running `t2 root` the default key is not set.
> INFO Starting SSH Session on Tessel. Type 'exit' at the prompt to end.
Warning: Identity file undefined not accessible: No such file or directory.
Permission denied (publickey).

reproduce this by plugging at least two Tessels and run `t2 root`

[x]    default key is present at ~/.tessel/id_rsa
[x]    Both Tessels are provisioned !
[ ]    `t2 root` also **should default to the default key** if it is not set by --key (it doesn't when multiple Tessels are plugged)

johnnyman727:
@Student007 are you passing in a key option? [The default key is set](https://github.com/tessel/t2-cli/blob/55c51328e39915a846657dbb74214e45d45b89cf/lib/tessel/provision.js#L23) when `provision.js` loads.

Student007:
> @Student007 are you passing in a key option? The default key is set when provision.js loads.

it seams it is not used if the `opts.key` is *undefined*. Means there is no path also if a key is present at `~/.tessel/id_rsa` but this is necessary if you have multiple Tessels accessible by wifi and you do not use the `--key` option (because you like to use the default key).

**I am suggesting three different solutions:**
https://github.com/tessel/t2-cli/pull/502 => ([handle a possible opts.key===*undefined* within `controller.root`](https://github.com/tessel/t2-cli/pull/502/files))
https://github.com/tessel/t2-cli/pull/498 => ([Override the default `--key` option with a default key path](https://github.com/tessel/t2-cli/pull/498/files))
https://github.com/tessel/t2-cli/pull/501 => ([Add a default path to the default `--key` option](https://github.com/tessel/t2-cli/pull/501/files))

My preference would be #498 because it is a valid pattern to override default behaviors if necessary.
My second favorite would be #501 because other commands would ignore `opts.key` if they do not use it. And former this was discussed [here](https://github.com/tessel/t2-cli/pull/330)
If no alternative solution is provided I would suggest to fix the issue by #502 because it simple catches a not defined `opts.key`until a better solution will be given by deeper researches.

...please add your own ideas (by discussion).

Student007:
@rwaldron what is your preference - ask you because you did the lightweight implementation of `t2 root` ?

johnnyman727:
[...]

PR #498, opened 2015-12-29
Adding default key for `t2 root`

Student007:
fixes https://github.com/tessel/t2-cli/issues/497

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -336,6 +336,13 @@ makeCommand('ap')
   });
 
 makeCommand('root')
+  .option('key', {
+    required: false,
+    default: controller.LOCAL_AUTH_KEY,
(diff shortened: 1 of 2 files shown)


johnnyman727:
@Student007 I don't think this is necessary because the `key` option is added [in the `makeCommand` function](https://github.com/Student007/t2-cli/blob/root_default_key/bin/tessel-2.js#L24-L29).

Student007:
@johnnyman727 it is necessary, because the issue is given. Here https://github.com/tessel/t2-cli/issues/501 is an alternative implementation.

Student007:
@johnnyman727 and one more alternative implementation https://github.com/tessel/t2-cli/pull/502

(please read the note within slack review channel. It is about my idea of how where is the root cause of the problem)

PR #501, opened 2015-12-30
Alternative PR: Adding default key for makeCommand (sqashed & hard rebased))

Student007:
@johnnyman727 instead of https://github.com/tessel/t2-cli/pull/498/

you could merge this one.

I fixed the tests for the list command.

In every case one of the both PRs is necessary because the issue
https://github.com/tessel/t2-cli/issues/497 is given.

Your decision what you merge. Simple close the other one when you
merged :smile:
> replaces this https://github.com/tessel/t2-cli/pull/500

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -24,6 +24,7 @@ function makeCommand(commandName) {
     .option('key', {
       required: false,
       metavar: 'PRIVATEKEY',
+      default: controller.LOCAL_AUTH_KEY,
       abbr: 'i',
       help: 'SSH key for authorization with your Tessel'
     })
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -967,3 +967,4 @@ module.exports = controller;
 // Shared exports
 module.exports.listTessels = controller.list;
 module.exports.getTessel = controller.get;
+module.exports.LOCAL_AUTH_KEY = Tessel.LOCAL_AUTH_KEY;
--- a/test/unit/bin-tessel-2.js
+++ b/test/unit/bin-tessel-2.js
@@ -148,6 +148,7 @@ exports['Tessel (cli: update)'] = {
       version: 42,
       _: ['update'],
       timeout: 5,
+      key: controller.LOCAL_AUTH_KEY,
       lanPrefer: false
     });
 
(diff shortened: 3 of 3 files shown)


Student007:
:clap: :ghost: :octocat: :checkered_flag: :tea:

lib/controller.js
```diff
@@ -967,3 +967,4 @@ module.exports = controller;
 // Shared exports
 module.exports.listTessels = controller.list;
 module.exports.getTessel = controller.get;
+module.exports.LOCAL_AUTH_KEY = Tessel.LOCAL_AUTH_KEY;
```
rwaldron: - `Tessel.LOCAL_AUTH_KEY` isn't defined in this file, so should not be exporting it
- `module.exports.LOCAL_AUTH_KEY = Tessel.LOCAL_AUTH_KEY` breaks the accessor

bin/tessel-2.js
```diff
@@ -24,6 +24,7 @@ function makeCommand(commandName) {
     .option('key', {
       required: false,
       metavar: 'PRIVATEKEY',
+      default: controller.LOCAL_AUTH_KEY,
```
rwaldron: Can you add `var Tessel = require('../lib/tessel/tessel');` to this file and use `Tessel.LOCAL_AUTH_KEY` directly?

test/unit/bin-tessel-2.js
```diff
@@ -521,8 +518,7 @@ exports['Tessel (cli: list)'] = {
 
     test.expect(4);
 
-    var keyPath = './FAKE_KEY';
```
rwaldron: Why is this test changed?

Student007: @rwaldron because setDefaultKey was called:

`controller.list` => `controller.keyHelper` => because of opts.key !== undefined `provision.setDefaultKey(opts.key)`is called
This is necessary, because `opts.key` is `undefined` what is a bad thing for usability about `t2 root`... you would be required to use `--key` in every case.

bin/tessel-2.js
Student007: @rwaldron if @johnnyman727 allows me to do :wink: ... I tried to add as less changes and code as I can. But former `var Tessel = require('../lib/tessel/tessel');` was in the `tessel-2.js` and I used `default: Tessel.LOCAL_AUTH_KEY`(PR wasn't merged).
I also would prefer this. Would you agree @johnnyman727 ? My local tests pass and the solution works. Waiting for submitting PR until getting a feedback.

test/unit/bin-tessel-2.js
rwaldron: Got it, thank you for explaining

johnnyman727:
I think #512 is the best way to solve this.

PR #502, opened 2015-12-30
Alternative PR: Adding default key within controller.root

Student007:
**It looks like the wrapper for controller.get doesn’t work like it
should what wasn’t covered by tests!**

There is no time to work on it now - I am 3 hours behind my day plan -
so I stop it now. Please merge one of the three PRs and open another
issue about right implementation of the wrapper.
> https://github.com/tessel/t2-cli/pull/498
> https://github.com/tessel/t2-cli/pull/501

also fixes https://github.com/tessel/t2-cli/issues/497
:smile:

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -737,6 +737,12 @@ controller.root = function(opts) {
   opts.lan = true;
   // We must already be authorized
   opts.authorized = true;
+
+  // Set a default key if it is not setten by command line
+  if (!opts.key) {
+    opts.key = Tessel.LOCAL_AUTH_KEY;
+  }
+
   // Fetch a Tessel
   return controller.standardTesselCommand(opts, function(tessel) {
     logs.info('Starting SSH Session on Tessel. Type \'exit\' at the prompt to end.');


johnnyman727:
@Student007 I'm not sure how this does anything different then what we already have in place [here](https://github.com/Student007/t2-cli/blob/root_default_key_controller/lib/controller.js#L34)? `Tessel.get` already sets the default key for SSH.

Student007:
@johnnyman727 
In the case no `t2 root` is called without `--key ~/.tessel/id_rsa` the **opts.key** is **undefined** so the conditional blog at https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L39-L42 within the wrapper is not catching this.
I think this is also the case if the wrapper is used for other things. When adding an *else* blog this would cause tests will not work anymore (and it doesn't help anything by purpose).
The cleanest solution is to add `default: controller.LOCAL_AUTH_KEY` to the default parameters within tessel-2.js what is exported from controller.js `Tessel.LOCAL_AUTH_KEY` (my preference).

The wrapper works for the option `--key`. @rwaldron did you ever try to use `t2 root` without a key (and multiple Tessels plugged) before committing the lightweight version of my current idea of `t2 root`? In general I thought you did it the lightweight way for matching milestone planing and knowing about I am totally blocked at the moment. But I didn't notice this issue because I didn't got time to **play around** with the Tessels (think this underlines my current situation!)...

johnnyman727:
@Student007 if `opts.key` is undefined, it won't set the new key path again because it was already set to a default [on initialization](https://github.com/tessel/t2-cli/blob/master/lib/tessel/provision.js#L23).

If I understand the issue correctly, the default file path doesn't exist on your computer so you can't connect? But you *must* have an SSH key path if you are using `t2 root` because it *only* works over LAN right now.

The resolution for this issue should actually be a better error message that explains that you'll need an SSH key for `t2 root`.

Student007:
@johnnyman727 I think this is not a path problem. Please try the following:

Use two or more Tessels plugged by USB and use `t2 root`

instead of using one Tessel ... it will error also on your computer :smile:

johnnyman727:
@Student007 yes, the CLI isn't currently designed to use two Tessels for anything. We should actually have an overarching issue to fix that at some point. But as it is, we have never tested the code with two Tessels attached (because, historically, we just haven't had very many Tessels). 

I don't think this is an issue with `t2 root`.

Student007:
@johnnyman727 in that case the PR https://github.com/tessel/t2-cli/pull/501 is the best solution - because it will prevent further issues depending on this in general.
(And also this is what I was doing when I worked on `t2 root`)

johnnyman727:
@Student007 I don't understand how #501 is a solution. The problem is that you are trying to use a function only available over SSH without an SSH key. The solution is to print a warning that you need an SSH key in order to use `root` and then abort. #501 just sets the keypath to the default again (after it was initially set). This won't help in the case that someone doesn't have keys on that path.

Student007:
> This won't help in the case that someone doesn't have keys on that path.

good point, but in this case the wrapper is not hitting *undefined* and will pass the opts.key to `provision.setDefaultKey(opts.key)` what will be checked about valid path [here](https://github.com/tessel/t2-cli/blob/master/lib/tessel/provision.js#L176) what will cause in a rejection path+`does not contain valid public and private SSH keys`

johnnyman727:
> good point, but in this case the wrapper is not hitting undefined and will pass the opts.key to provision.setDefaultKey(opts.key)

It will not try to set the default key again if it's `undefined` because [we check first](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L39-L42) before calling `setDefaultKey`.

Student007:
@johnnyman727 YES ... what is the reason it NEVER should be undefined ! (means: the keyHelper is also wrong - but not my issue)

johnnyman727:
@Student007 it should be undefined if the user does not explicity set an alternate path. I can't remember why we made that change but there was somewhere within `controller.js` that had to know if the user had specifically wanted to set the path or if it was being used as default.

Student007:
@johnnyman727 no the change had been done because we had problems with asynchronous way to `setDefaultKey()` ...

johnnyman727:
Ah no, now I remember. The problem is if you make `~/.tessel/id_rsa` the default path (as this PR does), but the user has no provisioned yet, then you will *always* get an error when you try to set the path as default. So, in effect, every new user would receive get an error thrown at them. It was fixed [here](https://github.com/tessel/t2-cli/pull/469).

In any case, as I mentioned before, this PR does not solve the problem so I will close it.

Student007:
@johnnyman727 
> It was fixed here.
> [...] this PR does not solve the problem so I will close it.

**well, the PR does solve MY problem** as well as my other alternative suggested PR. You didn't give a suggestion HOW you would solve the existing problem - and **this isn't fair enough**.

(ok, you just remember the reason you hold against... about https://github.com/tessel/t2-cli/issues/455 - I didn't know about BUT the mentioned argument about doesn't match today: The wrapper fixes this .. NOT the removing of default: ... )

Your arguments are not enough to explain why no default should be in the global `--key` option (commands without key ignores a default) OR why the [root command](https://github.com/tessel/t2-cli/pull/498/files) shouldn't override the global `--key` option also with a default AND finally your arguments doesn't explain why it shouldn't be handled within `controller.root` ([closed PR](https://github.com/tessel/t2-cli/pull/502/files)) when all other solutions doesn't match. **What do you want from me ?**

The situation is very very easy: **The problem exist** - show me a better solution than one of **my three suggestions**. One from you would be fair enough...

**The issue exists !**

johnnyman727:
Hi @Student007, I suggested a solution [here](https://github.com/tessel/t2-cli/pull/502#issuecomment-168227409) in my comments above. To reiterate, I think we should check if someone is attempting to use `t2 root` and does not have valid SSH keys in the designated key path (either provided or the default) and present an error if that's the case.

Student007:
@johnnyman727 
> The resolution for this issue should actually be a better error message that explains that you'll need an SSH key for t2 root.

your "resolution" was **better error msg** but keys are there as you have seen by testing yourself:
> @Student007 yes, the CLI isn't currently designed to use two Tessels for anything. We should actually have an overarching issue to fix that at some point. But as it is, we have never tested the code with two Tessels attached (because, historically, we just haven't had very many Tessels).

> I don't think this is an issue with t2 root.

however all three provided PR solve the problem - if you think it is a deeper thing (I don't belief) then it is a new issue (what can override the current PRs after merged one). I like to work with `t2 root`and the problem is solved properly with all three PR. I will not dive into your additional problems more as I did (because there is no more time)

[x] default key is present at `~/.tessel/id_rsa`
[x] Both Tessels are provisioned !
[!] the **ONLY** valid solution is (as for all other commands) `t2 root` **also should default to the default key** if it is not set by `--key`

lets begin a cleaner discussion here https://github.com/tessel/t2-cli/issues/497 (think you got what I mean and I got you like I should do a much deeper dive into the problems of the cli when working with multiple tessels - but it is not the time for games like this ! Excuse me!).

johnnyman727:
> [!] the ONLY valid solution is (as for all other commands) t2 root also should default to the default key if it is not set by --key

@Student007 perhaps I don't understand the issue you are trying to tackle because the above case is true. The CLI *does* use a default path without having to explicitly call `setDefaultKey`. You can check this for yourself by printing out whether Tessel `.isProvisioned` and the key path [here](https://github.com/tessel/t2-cli/blob/master/lib/lan_connection.js#L38-L40).

PR #503, opened 2015-12-31
Ap no credential changed warning

Student007:


Student007:
closed because tests didn't pass

HipsterBrown:
@Student007 They were only JCS errors if you still want to try pushing this through.

PR #506, opened 2015-12-31
t2 ap warn about credentials are not changed

Student007:

fixes https://github.com/tessel/t2-cli/issues/507

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -717,6 +717,9 @@ controller.createAccessPoint = function(opts) {
 
 controller.enableAccessPoint = function(opts) {
   opts.authorized = true;
+  if (opts.ssid) {
+    logs.warn('Credentials are not changed when switching on/off the accesspoint!');
+  }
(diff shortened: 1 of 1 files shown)


rwaldron:
Needs tests

Student007:
@rwaldron you are right ;) ... as written here https://github.com/tessel/t2-cli/issues/507
> @HipsterBrown
> It seems like the best way to fix that is to check if credentials are passed - if they are, assign those credentials before enabling or disabling.

> this #506 would do this. Maybe the --ssid shouldn't be flagged as required: true with this solution.
I think the --ssid is not necessary when switching the AP on/off because there is only one AP. Or do you plan to add multiple configuration sets what could be toggled by this way ?

> In general I think you are planing to work with multiple Tessels for mash nets (what is possible) so we should work together (at least for discussion) because I am also interested in having in mind there are multiple Tessels working together properly...

> In that case maybe it could be better to add a new command because there is no other way to identify a new configuration.

close this because I expect @HipsterBrown will rework this in general.

PR #511, opened 2016-01-03
[WIP] Allow deploys of Rust scripts

johnnyman727:
This is my initial brute force method of working through what it will be like to deploy Rust. There is a lot of cleanup/refactoring to do because this currently destroys JavaScript deployment and will probably not pass Grunt's tests. I just wanted to see how far I could get in a day.

To Test:
* Start up [a Rust compilation server VM](https://github.com/tessel/rust-compilation-server#install) 
* Create a Rust project on your host machine (not your VM). `cargo new hello --bin` should work well for that.
* Plug in your Tessel and make sure it finishes booting
* Check out this branch of the CLI
* From within your Rust project (the `hello` directory), run `t2 run .`. You should get output like this:
```
➜  hello git:(master) ✗ t2 run .
INFO Connected to Hen.
INFO Compiling Rust code...
INFO Writing project to RAM on Hen (295.535 kB)...
INFO Deployed.
INFO Running Rust Executable...
Hello, world!
```

TODO:
- [ ] Tests
- [ ] Get the cross compilation tools and express server on a long-living public IP Address
- [ ] Add automatic language detection to the CLI
- [ ] Refactor the code so it's easy to swap in code bundling, optional compilation, deployment, and initialization for different languages in the CLI
- [ ] Fix grunt complaints

### DIFF
--- a/lib/tessel/commands.js
+++ b/lib/tessel/commands.js
@@ -28,6 +28,9 @@ module.exports.untarStdin = function(filepath) {
 module.exports.runScript = function(filepath, relpath) {
   return ['node', filepath + relpath];
 };
+module.exports.runRustExecutable = function(filepath) {
+    return [filepath];
+}
 module.exports.openStdinToFile = function(filepath) {
   return ['dd', 'of=' + filepath];
 };
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -1,5 +1,6 @@
 // System Objects
 var path = require('path');
+var http = require('http');
 
 // Third Party Dependencies
 var browserify = require('browserify');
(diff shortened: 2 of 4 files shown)


rwaldron:
Why is the JavaScript cli tool being used for Rust code deployment? Shouldn't a Rust port of the cli be made? This is going to get very messy—especially when Python support comes along. I don't thing the maintenance cost is worth bearing.

rwaldron:
The to-do list and general approach is going to be further at odds with and complicated by the pre-compiled binary module system. I don't think this is feasible.

johnnyman727:
@rwaldron I think that are a couple of issues that need to be worked out, including yours and @kevinmehall's security concerns. I'll go ahead and open a general discussionary issue about how we can handle language support when it comes to the CLI.

johnnyman727:
Discussion is here: https://github.com/tessel/t2-cli/issues/513

cflewis:
I've verified the steps here work (yay!) but I had to hack the IP address in `deploy.js` to my `docker-machine ip` on Mac OS X.

johnnyman727:
Closing in favor of another proof of concept branch...

PR #512, opened 2016-01-04
Uses correct key path for root command

### LINKED ISSUE #497: No default Key for root command (if multiple Tessels are plugged)

johnnyman727:
Fixes #497.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -744,7 +744,7 @@ controller.root = function(opts) {
       // Spawn a new SSH process
       var child = cp.spawn('ssh', ['-i',
         // Use the provided key path
-        opts.key,
+        Tessel.LOCAL_AUTH_KEY,
         // Connect to the Tessel's IP Address
         'root@' + tessel.lanConnection.ip
       ], {
--- a/test/unit/controller.js
+++ b/test/unit/controller.js
@@ -957,6 +957,10 @@ exports['controller.root'] = {
       return callback(tessel);
     }.bind(this));
 
+    // Return our own local auth key (it won't be set because we have stubbed
+    // that functionality out of standardTesselCommand)
+    this.sandbox.stub(Tessel, 'LOCAL_AUTH_KEY', testKey);
+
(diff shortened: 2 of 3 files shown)


lib/controller.js
```diff
@@ -744,7 +744,7 @@ controller.root = function(opts) {
       // Spawn a new SSH process
       var child = cp.spawn('ssh', ['-i',
         // Use the provided key path
-        opts.key,
+        Tessel.LOCAL_AUTH_KEY,
         // Connect to the Tessel's IP Address
```
Student007: @rwaldron from my point of view this doesn't make sense (also if it works).
@johnnyman727 thanks for that suggestion but could you explain why this is a better solution ?

johnnyman727: @Student007 in my opinion, it's the best solution because it's the least amount of changed code.

rwaldron: Doesn't this change prevent the end user from specifying a key? Or am I misunderstood? Should it be: `opts.key || Tessel.LOCAL_AUTH_KEY,`?

johnnyman727: No, it does not because the new key path gets set within Tessel.get.

rwaldron: Ah—Indeed! I overlooked the `controller.get` wrapper.

rwaldron:
After reading through all of the associated issue comments and PRs, I'm going to land this because it's the simplest and likely easiest to further iterate on. Thanks to everyone that participated!!

PR #514, opened 2016-01-04
Alternative PR: Adding default key by require Tessel

Student007:
As discussed
https://github.com/tessel/t2-cli/pull/501#discussion_r48755581 this PR
solves https://github.com/tessel/t2-cli/issues/497 and is alternative
to #498 #501 #502 and @johnnyman727 #512 what can be closed if this one
had been merges.
This PR contains @johnnyman727 tests for root from #512 !

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -8,6 +8,7 @@ var parser = require('nomnom').script('t2');
 
 // Internal
 var controller = require('../lib/controller');
+var Tessel = require('../lib/tessel/tessel');
 var init = require('../lib/init');
 var key = require('../lib/key');
 var logs = require('../lib/logs');
@@ -23,6 +24,7 @@ function makeCommand(commandName) {
     })
     .option('key', {
       required: false,
+      default: Tessel.LOCAL_AUTH_KEY,
       metavar: 'PRIVATEKEY',
       abbr: 'i',
       help: 'SSH key for authorization with your Tessel'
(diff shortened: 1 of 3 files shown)


johnnyman727:
I'm still not a fan of setting a default key path in the command parser for these reasons:
1. This reverses the fix of https://github.com/tessel/t2-cli/pull/469. In effect, it will make every new user experience a CLI error until they run `t2 provision` (what if they don't want to use Tessel over LAN?)
2. It adds the overhead of checking the file system for the existence of the key with every command.
3. We lose the ability to see if the user wanted to explicitly change the path to a new one. This is not used anywhere in the codebase yet but I can imagine it becoming handy.

Student007:
@johnnyman727 in that case #498 => (Override the default --key option with a default key path) would be the right one. Then you could create a new PR after merging #498 and add your tests from #512 

The tests in this PR are failing for special cases of Travis CI (local grunt works) ... so this PR could be closed.
@rwaldron  now I remember why I removed the tests lines from StandardList test ... they are causing the fail :cry: 
#498 isn't affected by this problem.

johnnyman727:
> @johnnyman727 in that case #498 => (Override the default --key option with a default key path) would be the right one.

No, you are still [setting the default option](https://github.com/tessel/t2-cli/pull/498/files#diff-f5ef2431cf991331734a97c4b931a0f6R341) in this PR which would lead to the exact same problems I mentioned above. We should not change the way the options are parsed.

PR #515, opened 2016-01-05
Allows LAN deploys to close the process properly

johnnyman727:
Adding the `exec` option of `pty=true` to a LAN connection `exec` makes it possible for it to kill the spawned processes if the connection closes.

I would have liked to just made this a default option on all execs but when I tried that, I started getting syntax errors reported in the JavaScript of my blinky test file on Tessel. I don't really understand what pseudo-tty does but it seems to have negative impacts in some cases.

### DIFF
--- a/lib/lan_connection.js
+++ b/lib/lan_connection.js
@@ -40,7 +40,7 @@ LAN.Connection = function(opts) {
   }
 };
 
-LAN.Connection.prototype.exec = function(command) {
+LAN.Connection.prototype.exec = function(command, options) {
   var self = this;
   return new Promise(function(resolve, reject) {
 
@@ -56,8 +56,11 @@ LAN.Connection.prototype.exec = function(command) {
     // Turn an array of args into a string to execute over SSH
     command = self._processArgsForTransport(command);
 
+    // Set default options if none provided
+    options = options || {};
+
     // Execute the bash command
-    self.ssh.exec(command, function(err, godStream) {
+    self.ssh.exec(command, options, function(err, godStream) {
       if (err) {
         return reject(err);
       }
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -560,7 +560,9 @@ actions.runScript = function(t, filepath, entryPoint, opts) {
 
   logs.info('Running %s...', opts.slim ? 'bundled project' : entryPoint);
   return new Promise(function(resolve) {
-    return t.connection.exec(commands.runScript(filepath, actualEntryPointName))
+    return t.connection.exec(commands.runScript(filepath, actualEntryPointName), {
+        pty: true
+      })
       .then(function(remoteProcess) {
 
         // When the stream closes
--- a/test/unit/deploy.js
+++ b/test/unit/deploy.js
@@ -182,7 +182,8 @@ exports['Tessel.prototype.deployScript'] = {
   },
 
   runScript: function(test) {
-    test.expect(10);
+    test.expect(11);
+    this.exec = sandbox.spy(this.tessel.connection, 'exec');
     deployTestCode(this.tessel, test, {
       push: false,
       single: false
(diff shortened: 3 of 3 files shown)


rwaldron:
From: https://github.com/mscdex/ssh2

> pty - mixed - Set to true to allocate a pseudo-tty with defaults, or an object containing specific pseudo-tty settings (see 'Pseudo-TTY settings'). Setting up a pseudo-tty can be useful when working with remote processes that expect input from an actual terminal (e.g. sudo's password prompt).

If you look at ssh2's source, `pty: true,` results in the creation of a "pseudo-terminal" that is expecting input to send to remote process, follow: https://github.com/mscdex/ssh2/blob/master/lib/client.js#L1126-L1174, which calls this: https://github.com/mscdex/ssh2-streams/blob/master/lib/ssh.js#L1140-L1196, which in turn calls this: https://github.com/mscdex/ssh2-streams/blob/master/lib/ssh.js#L4877-L4992 

To test the behavior change, I created the following program: 

```js
var tessel = require('tessel');
tessel.led[2].on();

var count = 0;

setInterval(function () {
  tessel.led[2].toggle();
  tessel.led[3].toggle();

  if (++count === 100) {
    tessel.close();
    process.nextTick(process.exit);
  }
}, 100);

console.log("I'm blinking! (Press CTRL + C to stop)");
```

- On branch `jon-fix-lan-end`:
  - Open two terminal windows: 
    - **A**: ssh to the T2
    - **B**: run `t2 run index.js --lan`
  - Once the deployment process in **B** prints out: `I'm blinking! (Press CTRL + C to stop)`:
    - In **A**, run `ps`
      - There will be a process for `node /tmp/remote-script/__tessel_program__.js`
  - In **B**, sent `SIGINT` (`^C`)
    - Alternating LEDs must stop immediately
    - In **A**, run `ps`
      - Must be no process `node /tmp/remote-script/__tessel_program__.js`
    

- On branch `master`:
  - Open two terminal windows: 
    - **A**: ssh to the T2
    - **B**: run `t2 run index.js --lan`
  - Once the deployment process in **B** prints out: `I'm blinking! (Press CTRL + C to stop)`:
    - In **A**, run `ps`
      - There will be a process for `node /tmp/remote-script/__tessel_program__.js`
  - In **B**, sent `SIGINT` (`^C`)
    - Alternating LEDs will continue 
    - In **A**, run `ps`
      - There will be a process for `node /tmp/remote-script/__tessel_program__.js`

-------


When I made this change: `options = options || { pty: true };`, the following error occurred: 

```
...
INFO Running bundled project...
module.js:339
    throw err;
    ^

Error: Cannot find module '/tmp/remote-script/__tessel_program__.js'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:134:18)
    at node.js:961:3
```

A quick look in `root@bishop:/tmp/remote-script#` reveals that no files were created. I'm also not sure what specifically caused this interference... So I commented out the promise resolution that occurs right after `INFO Deployed.` is displayed and check the contents of `/tmp/remote-script/`, which was actually empty. I did some brute-force manual `ps` logging during a deploy and observed that the following things occurred (as expected): 

- A process for `tar -x -C /tmp/remote-script/`
- A process for `node /tmp/remote-script/__tessel_program__.js`

Here's the difference I can see: 

- `pty: true` negatively interferes when **writing data to the remote process stream**  (apparently prevents?) 
- `pty: true` does not negatively interfere with **executing simple commands**

Finally, I think it's completely ok to parameterize the options for `ssh2.exec(...)` for use _when necessary_.

johnnyman727:
@rwaldron is this good to merge?

johnnyman727:
Also, thanks for looking into `pty` - I came to the same conclusions.

PR #517, opened 2016-01-05
Uses an alternative wifi scan to produce consistent results

### LINKED ISSUE #350: Listing wifi networks only works the first time until reboot
➜  t2-cli git:(jon-erase) ✗ t2 wifi -l
INFO Connected to Folly over LAN
INFO Scanning for available networks...
INFO Currently visible networks (26):
INFO 	 technicallyHome (57/70)
INFO 	 Fried Chicken Sandwich (53/70)
INFO 	 xfinitywifi (45/70)
INFO 	 XFINITY (45/70)
INFO 	 CableWiFi (45/70)
INFO 	 Oy Francisco (45/70)
INFO 	 NETGEAR61 (43/70)
INFO 	 NETGEAR11 (35/70)
INFO 	 ATT5Q5Y3R2 (35/70)
INFO 	 xfinitywifi (29/70)
INFO 	 CableWiFi (29/70)
INFO 	 XFINITY (27/70)
INFO 	 Ergo (27/70)
INFO 	 CableWiFi (25/70)
INFO 	 Friendship Pizza (25/70)
INFO 	 XFINITY (25/70)
INFO 	 xfinitywifi (25/70)
INFO 	 xfinitywifi (25/70)
INFO 	 ATT496 (25/70)
INFO 	 moms comp (23/70)
INFO 	 The Hittite Network (23/70)
INFO 	 xfinitywifi (23/70)
INFO 	 HOME-F652 (21/70)
INFO 	 2NORA968 (21/70)
INFO 	 freekucho (21/70)
INFO 	 HOME-D338 (21/70)
➜  t2-cli git:(jon-erase) ✗ t2 wifi -l
INFO Connected to Folly over LAN
INFO Scanning for available networks...
INFO Currently visible networks (0):
```

### PR DESCRIPTION
**Pull request by johnnyman727**
Fixes #350.

We were originally using the `ubus call iwinfo scan '{"device":"wlan0"}` to return a JSON result of all available networks. It was convenient and easy to parse but the downside is that is seems to refresh very slowly. If you call the function again within a few minutes, you get empty results.

Instead, I've found another command that always returns the networks in range but it's not in JSON. There may be a better way to do this within OpenWRT but I know not the way. It was faster for me to modify the CLI.

@HipsterBrown thanks for the regex help.

### DIFF
--- a/lib/tessel/wifi.js
+++ b/lib/tessel/wifi.js
@@ -13,27 +13,31 @@ var Tessel = require('./tessel');
 Tessel.prototype.findAvailableNetworks = function() {
   logs.info('Scanning for available networks...');
   return this.simpleExec(commands.scanWiFi())
-    .then(function wifiScanResults(resultsJSON) {
-      // Parse the response
-      var networks = '';
-      try {
-        networks = JSON.parse(resultsJSON).results;
-      } catch (err) {
-        // Throw any unfortunate errors
-        return Promise.reject(err);
-      }
+    .then(function wifiScanResults(resultsData) {
+      // For each string chunk
+      return resultsData.trim().split('\n\n').reduce((networks, entry) => {
+        try {
 
-      // Sort by signal strength
-      networks.sort(compareBySignal);
+          var ssidRegex = /ESSID: "(.*)"/;
+          var qualityRegex = /Quality: (.*)/;
+          var encryptionRegex = /Encryption: (.*)/;
 
-      // Remove networks without SSIDs
-      networks.forEach(function(network, index) {
-        if (!network.ssid) {
-          networks.splice(index, 1);
+          var networkInfo = {
+            // Parse out the SSID
+            ssid: ssidRegex.exec(entry)[1],
+            // Parse out the quality of the connection
+            quality: qualityRegex.exec(entry)[1],
+            // Parse the security type - unused at the moment
+            security: encryptionRegex.exec(entry)[1],
+          };
+          // Add this parsed network to our array
+          networks.push(networkInfo);
+        } catch (err) {
+          // Suppress errors created by entries that cannot be parsed.
         }
-      });
 
-      return networks;
+        return networks;
+      }, []).sort(compareBySignal);
     });
 };
 
@@ -60,9 +64,10 @@ Tessel.prototype.getWifiInfo = function() {
 };
 
 function compareBySignal(a, b) {
-  if ((a.quality / a.quality_max) > (b.quality / b.quality_max)) {
+  // eval(quality) turns a string fraction like 21/70 into a decimal
+  if (eval(a.quality) > eval(b.quality)) {
     return -1;
-  } else if ((a.quality / a.quality_max) < (b.quality / b.quality_max)) {
+  } else if (eval(a.quality) < eval(b.quality)) {
     return 1;
   } else {
     return 0;
--- a/.jshintrc
+++ b/.jshintrc
@@ -15,6 +15,7 @@
   "strict": false,
   "esnext": true,
   "unused": true,
+  "evil" : true,
   "globals": {
     "exports": true,
     "document": true,
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -607,7 +607,7 @@ controller.printAvailableNetworks = function(opts) {
 
         // Print out networks
         networks.forEach(function(network) {
-          logs.info('\t', network.ssid, '(' + network.quality + '/' + network.quality_max + ')');
+          logs.info('\t', network.ssid, '(' + network.quality + ')');
         });
       });
   });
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by HipsterBrown** [comment]
Looks like a JSC error in the CI. 

Have you tried this on the Tessel yet?

**Comment by johnnyman727** [comment]
@HipsterBrown thanks for catching that - I'll fix and rebase. I have tried this on a Tessel and it works great.

**Comment by HipsterBrown** [comment]
@johnnyman727 Awesome. The code looks good to me and the tests pass, so that's a good confirmation. 

One suggestion would be caching the Regex patterns to variables before the `forEach` to keep that process as performant as possible. Otherwise, aces on the build.

File: lib/tessel/wifi.js [hunk 1]
```diff
@@ -13,26 +13,43 @@ var Tessel = require('./tessel');
...
+    .then(function wifiScanResults(resultsData) {
+      // Array of network dicts
+      var networks = [];
+      // Split data into individual networks
+      var networkStrings = resultsData.split('\n\n');
+      if (networkStrings.length <= 1) {
+        return networks;
       }
+      // Removes a bogus final element that consists of `\n`
+      networkStrings.pop();
```
rwaldron: What happens if you do `resultsData.trim().split('\n\n');` above?

johnnyman727: That's a cleaner alternative and it seems to work just the same. :+1:

rwaldron: I just tried `trim()` and it fixes the issue as well, and I think more appropriately

lib/tessel/wifi.js
```diff
@@ -13,26 +13,43 @@ var Tessel = require('./tessel');
...
-      } catch (err) {
-        // Throw any unfortunate errors
-        return Promise.reject(err);
+    .then(function wifiScanResults(resultsData) {
+      // Array of network dicts
+      var networks = [];
+      // Split data into individual networks
+      var networkStrings = resultsData.split('\n\n');
+      if (networkStrings.length <= 1) {
+        return networks;
```
rwaldron: So, if there is only one entry then return the empty list? Is that intentional or is that part of the work around on the next line?

johnnyman727: Yes, that can be removed now. :confetti_ball:

lib/tessel/wifi.js
```diff
@@ -13,26 +13,38 @@ var Tessel = require('./tessel');
...
-        networks = JSON.parse(resultsJSON).results;
-      } catch (err) {
-        // Throw any unfortunate errors
-        return Promise.reject(err);
-      }
+    .then(function wifiScanResults(resultsData) {
+      // Array of network dicts
+      var networks = [];
+      // Split data into individual networks
+      var networkStrings = resultsData.trim().split('\n\n');
```
rwaldron: :+1:

lib/tessel/wifi.js
```diff
@@ -13,26 +13,38 @@ var Tessel = require('./tessel');
...
+          };
+          // Add this parsed network to our array
+          networks.push(networkInfo);
+        } catch (err) {
+          return Promise.reject(new Error('Invalid data returned. Unable to parse networks.'));
+        }
+      });
 
       // Sort by signal strength
       networks.sort(compareBySignal);
```
rwaldron: Pinged you on slack, re: some cleanups here.

lib/tessel/wifi.js
```diff
@@ -60,9 +64,10 @@ Tessel.prototype.getWifiInfo = function() {
 };
 
 function compareBySignal(a, b) {
-  if ((a.quality / a.quality_max) > (b.quality / b.quality_max)) {
+  // eval(quality) turns a string fraction like 21/70 into a decimal
+  if (eval(a.quality) > eval(b.quality)) {
     return -1;
-  } else if ((a.quality / a.quality_max) < (b.quality / b.quality_max)) {
+  } else if (eval(a.quality) < eval(b.quality)) {
```
rwaldron: WHOA. I didn't see this here earlier and it's a security hazard.

HipsterBrown: I saw the `evil` addition to the `.jshintrc` but wasn't sure why. Missed that as well.

johnnyman727: What is the case where it is a security hazard here? If someone downloads malicious code onto their T2 which overwrites the original `wifi scan` function to return root-seeking JS code as the `quality` parameter?

LinusU:
```javascript
function parseFraction (src) {
  var m = /([0-9]+)\/([0-9]+)/.exec(src)
  if (m === null) return NaN

  return Number(m[1]) / Number(m[2])
}
```

Maybe it's not that big of a security hazard in this particular instance, but I would consider it a code smell. It's easy to parse without `eval`, and it makes the intent clearer.

PR #529, opened 2016-01-15
Support "bare_returns" for CJS modules

rwaldron:
This is necessary for CommonJS modules that take advantage of CJS's implied IIFE wrapper that allows for "bare returns" (returns that appear out of a function, which are actually inside a wrapper function when the module is evaluated). The option is not forwarded when using the provided `minify` function.

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -819,9 +819,32 @@ actions.project = function(options) {
 };
 
 actions.compress = function(source) {
-  return uglify.minify(source.toString(), {
-    fromString: true
-  }).code;
+  source = typeof source === 'string' ? source : source.toString();
+
+  var content = uglify.parse(source, actions.compress.options);
+
+  content.figure_out_scope();
+  content = content.transform(uglify.Compressor({
+    warnings: false
+  }));
+
+  var code = uglify.OutputStream();
+  content.print(code);
+
+  return code.toString();
+};
+
+actions.compress.options = {
+  fromString: true,
+  bare_returns: true,
+  spidermonkey: false,
+  outSourceMap: null,
+  sourceRoot: null,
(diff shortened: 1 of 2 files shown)


HipsterBrown:
# Smoke Test

As @rwaldron told me, this unblocks the use of express.js on Tessel. So this test project runs the "Hello World" example from the express.js docs. 

## Test Project:

**Setup:**
`mkdir tessel-test`
`cd tessel-test`
`t2 init`
`npm install --save express`

**Files:**

`index.js`
From the [Express Hello World](http://expressjs.com/en/starter/hello-world.html).

```js
var express = require('express');
var app = express();

app.get('/', function (req, res) {
  res.send('Hello World!');
});

app.listen(3000, function () {
  console.log('Example app listening on port 3000!');
});
```

Run project with `t2 run index.js`

## Results

**Terminal:**

```bash
nickhehr@Nicks-MacBook-Pro-2 tessel-test $ t2 run index.js --lan
INFO Building project.
INFO Writing project to RAM on tessel-router (641.024 kB)...
INFO Deployed.
INFO Running index.js...
Example app listening on port 3000!
``` 

**Browser:**
I connected to the Tessel through a running access point, so I went to http://tessel-router.local:3000.
screen shot 2016-01-15 at 4 44 03 pm

Success!!!

rwaldron:
Awesome smoke test :)

rwaldron:
And thanks for the review!

johnnyman727:
Nice catch @rwaldron!

What did the error look like before this patch was in place?

rwaldron:
Basically the surfaced error regarding this: 

```
$ node
> return;
SyntaxError: Illegal return statement
```

johnnyman727:
:+1: Thanks for the clarification.

Issue #507, opened 2015-12-31
t2 ap: no warning about credentials are noch changed when switching on/of the AP


Student007:
when using the `trigger` about on/off ... you cannot push new credentials to the AP ... this only works when on or off is NOT used.

johnnyman727:
It seems like the best way to fix that is to check if credentials are passed - if they are, assign those credentials before enabling or disabling.

Student007:
@HipsterBrown 
> It seems like the best way to fix that is to check if credentials are passed - if they are, assign those credentials before enabling or disabling.

this https://github.com/tessel/t2-cli/pull/506 would do this. Maybe the `--ssid` shouldn't be flagged as `required: true` with this solution.
I think the `--ssid` is not necessary when switching the AP on/off because there is only one AP. Or do you plan to add multiple configuration sets what could be toggled by this way ?

In general I think you are planing to work with multiple Tessels for mash nets (what is possible) so we should work together (at least for discussion) because I am also interested in having in mind there are multiple Tessels working together properly... 

In that case maybe it could be better to add a new command because there is no other way to identify a new configuration.

johnnyman727:
> this #506 would do this.

Unfortunately, #506 just prints a warning and then enables/disables the AP. It does not modify the AP if a new `ssid`, `password`, or `security` setting is provided.

Student007:
@HipsterBrown think you are currently working on it and it isn't necessary I am further trying to provide the PR for something is covered by your work. Also you could use my provided solution or something and make it better than #506 (with tests of cause as @rwaldron mentioned at #506).
I am back in lecture and exam period from tomorrow on. For that I assigned you to this issue.

PR #530, opened 2016-01-15
ssid not required for ap --trigger

### LINKED ISSUE #528: Toggling the Tessel access point requires ssid
When trying to `--trigger` my Tessel access point, the cli would insist I needed to include an ssid:

```bash
nickhehr@Nicks-MacBook-Pro-2 t2-cli (mime-types-include-list) $ t2 ap --trigger off

ssid argument is required

Usage: t2 ap [trigger] [options]

trigger     Trigger, i.e. on OR off, the access point

Options:
   -t TIMEOUT, --timeout TIMEOUT     Set timeout in seconds for scanning for networked tessels  [5]
   -i PRIVATEKEY, --key PRIVATEKEY   SSH key for authorization with your Tessel
   --name NAME                       The name of the tessel on which the command will be executed
   --lan                             Use only a LAN connection
   --usb                             Use only a USB connection
   --lanPrefer                       Prefer a LAN connection when available, otherwise use USB.  [false]
   -p, --pass                        Password to access network.
   -s, --security                    Encryption to use on network (i.e. wep, psk, psk2, wpa, wpa2).

Configure the Tessel as an access point
```

Not sure an ssid should be necessary for toggling an access point on or off.

HipsterBrown:
Fixes #528

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -299,7 +299,6 @@ makeCommand('version')
 makeCommand('ap')
   .option('ssid', {
     abbr: 'n',
-    required: true,
     help: 'Name of the network.'
   })
   .option('pass', {


Student007:
@HipsterBrown looks good for me :+1: 

is https://github.com/tessel/t2-cli/issues/507#issuecomment-169264427 related

HipsterBrown:
@Student007 thanks. This doesn't cover that comment, but I do want to work on the feedback from `t2 ap` commands after finishing up the Network API for the `tessel` module.

Issue #490, opened 2015-12-15
Investigate why it takes ~5 seconds to load CLI `require`s


johnnyman727:
It takes a super long time before `"Looking for your Tessel"` gets printed after running any command (about 3-5 seconds for me on OSX10). We should look into why this takes so long and how to reduce that time.

flaki:
Oh yes I was wondering why running any `t2` command were this unnervingly slow.

johnnyman727:
@flaki I did a preliminary investigation and it looks like the 3-5 seconds are actually just processing all the `require` statements. We might want to implement a way to more lazily load those dependencies.

rwaldron:
I'm immediately suspicious of third party modules.

A quick survey turns up a few we could do without (whether they are causing issues, I can't say, but unnecessary none-the-less):

- bluebird 
  - just use built in Promise
- debug 
  - used inconsistently anyway?
- osenv 
  - getting "home" is done in deploy.js without a 3rd party module, we can refactor and reuse our own utility
- async 
  - as used here: https://github.com/tessel/t2-cli/blob/master/lib/tessel/provision.js#L109 can be replaced by two promises that use `Promise.all([a, b]).then` 
  - here: https://github.com/tessel/t2-cli/blob/master/lib/usb/usb_daemon.js#L56 I'm not sure why that's async.eachSeries? 
  - here: https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L350 again, not sure why?
- lib/tessel/deploy.js
  - can be lazy required
- lodash
  - can likely be optimized (needs further investigation)

rwaldron:
Addressing various fixes should be done in different PRs, vs one huge change set

LinusU:
> bluebird
> just use built in Promise

Yes please

> async
> here: https://github.com/tessel/t2-cli/blob/master/lib/usb/usb_daemon.js#L56 I'm not sure why that's async.eachSeries?

It seems like the `.closeProcess` is asynchronous and that all processes should be closed after each other. There is however a big problem with this code since it will never return if `(proc)` ever is falsely, since the then the callback will never fire.

> here: https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L350 again, not sure why?

This could be much easier written using `Promise.all`

> lodash
> can likely be optimized (needs further investigation)

[...]

PR #545, opened 2016-02-08
drop bluebird

LinusU:
*related to #490*

This drop the dependency on Bluebird and uses only the builtin `Promise` instead.


rwaldron:
I suspect the appveyor failure is bogus, so I've restarted the build

rwaldron:
I knew it!

LinusU:
Thanks :+1:

Issue #542, opened 2016-02-03
Investigate ways of making binary deps detectable even when the .node binary isn't present


rwaldron:
If the dep doesn't build on the local machine, even with `--force`, try to detect the dependency for inclusion in the deployment.

PR #551, opened 2016-02-09
Detect binary modules even if .node file is missing. Fixes gh-542

rwaldron:
~~I still need to write new tests for this~~

Tests included


lib/tessel/deploy.js
```diff
@@ -421,20 +440,85 @@ actions.resolveBinaryModules = function(opts) {
...
+        // files can be, take a look at this:
+        //
+        // https://github.com/nodejs/node-gyp/wiki/%22binding.gyp%22-files-out-in-the-wild
+        //
+        var bindingGypPath = path.join(globRoot, modulePath, binName);
+        var bindingGypData;
+        var bindingGypJson;
+
+        try {
+          // Sometimes people write really tidy binding.gyp files that are
```
johnnyman727: This comment makes me simultaneously really happy and really sad.

lib/tessel/deploy.js
```diff
@@ -528,6 +604,21 @@ actions.resolveBinaryModules = function(opts) {
   });
 };
 
+actions.resolveBinaryModules.readGypFileSync = function(gypfile) {
+  var python = process.env.PYTHON || 'python';
+  var program = `import ast, json; print json.dumps(ast.literal_eval(open("${gypfile}").read()));`;
```
johnnyman727: So we've got Python code inside a JavaScript library to pull in a binary likely written in C++/C. Nice.

lib/tessel/deploy.js
```diff
@@ -528,6 +604,21 @@ actions.resolveBinaryModules = function(opts) {
...
+  var decoder = new StringDecoder('utf8');
+  var result = cp.spawnSync(python, ['-c', program]);
+  var output = result.output;
+
+  return output.reduce((accum, buffer) => {
+    if (buffer) {
+      accum += decoder.write(buffer);
+    }
+    return accum;
```
johnnyman727: This function is actually way cleaner than I thought it would end up being.

lib/tessel/deploy.js
```diff
@@ -421,20 +440,85 @@ actions.resolveBinaryModules = function(opts) {
...
+
+            logMissingBinaryModuleWarning(packageJson.name);
+            return bins;
+          }
+        }
+
+        if (bindingGypJson && Array.isArray(bindingGypJson.targets)) {
+          // Anything that can't be covered by this will have to be
+          // dealt with as we encounter them and as they are reported.
+          binName = bindingGypJson.targets[0].target_name + '.node';
```
johnnyman727: Seems reasonable. I don't really understand why there can be multiple targets but that's not really our problem.

johnnyman727:
+100 points for creativity.

lib/tessel/deploy.js
rwaldron: Same here my friend

lib/tessel/deploy.js
rwaldron: Yep!

:D

lib/tessel/deploy.js
rwaldron: Two things really stand out for me: 

- The addition of spawnSync
- Template strings really make a difference

lib/tessel/deploy.js
rwaldron: I have no idea either, but I had to draw a line somewhere and this seemed reasonable

rwaldron:
FWIW, I actually made use of my Windows 10 machine to work out the issues in the tests. That was interesting...

rwaldron:
@johnnyman727 @HipsterBrown good to go here?

johnnyman727:
Code looks good to me but I didn't get a chance to test it in action. Do you have a smoke test you use?

rwaldron:
Smoke testing this is pretty convoluted. As I began writing a smoke test for this, I discovered that nested binaries were still getting dropped—this has been fixed.

rwaldron:
## Smoke Test

With a fully booted Tessel 2 running, paste the following into the terminal: 

```sh
mkdir missing-binary-smoke-test && cd missing-binary-smoke-test && t2 init;
npm install bufferutil;
rm node_modules/bufferutil/build/Release/bufferutil.node;
rm node_modules/bufferutil/fallback.js;
echo "var bu = require('bufferutil').BufferUtil; console.log(typeof bu);" > index.js;
t2 run index.js;
```


1. Creates a directory, enters the directory and initializes a tessel project
2. Installs a module the contains a binary
3. Delete the the binary file
4. Delete the "fallback", this ensures that the running script doesn't produce a false positive
5. Populate the index.js file with a simple program
6. Deploy to the tessel


A complete and successful run will _end_ with the following: 

```sh
INFO Connected to bishop.
INFO Building project.
INFO Writing project to RAM on bishop (38.912 kB)...
INFO Deployed.
INFO Running index.js...
function
```

rwaldron:
Note that there remains a known issue where modules that contain a binary which is directly "exported" via package.json "main" property are still problematic. T2's deployment has no problem finding the binary and deploying, it but the node process running on the T2 cannot find it; this is out of scope for this project to attempt to fix.

johnnyman727:
:+1: Your smoke tests worked for me. I also did one additional test just for fun which also passed:

```
npm install utf-8-validate
rm node_modules/utf-8-validate/build/Release/validation.node
rm node_modules/utf-8-validate/fallback.js
echo "var u8v = require('utf-8-validate').Validation.isValidUTF8; console.log(typeof u8v);" > index.js
t2 run index.js
```
Output:
```
function
```

Feel free to merge!

rwaldron:
Thanks @johnnyman727 !

johnnyman727:
:pray:

PR #555, opened 2016-02-10
[WIP] Misc improvements

johnnyman727:
I'll eventually break this into smaller PRs which target a distinct issue.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -52,6 +52,10 @@ Tessel.list = function(opts) {
 
   return new Promise(function(resolve, reject) {
 
+    if (opts.verbose === false) {
+      logs.disable();
+    }
+
     // Grab all attached Tessels
     logs.info('Searching for nearby Tessels...');
 
@@ -98,14 +102,14 @@ Tessel.list = function(opts) {
       } else if (foundTessels.length === 1) {
         // Close all opened connections and resolve
         controller.closeTesselConnections(foundTessels)
-          .then(resolve);
+          .then(() => resolve(foundTessels));
       }
       // If we have only one Tessel or two Tessels with the same name (just USB and LAN)
       else if (foundTessels.length === 1 ||
         (foundTessels.length === 2 && foundTessels[0].name === foundTessels[1].name)) {
         // Close all opened connections and resolve
         controller.closeTesselConnections(foundTessels)
-          .then(resolve);
+          .then(() => resolve(foundTessels));
       }
       // Otherwise
       else {
@@ -127,7 +131,7 @@ Tessel.list = function(opts) {
             logs.info('Set default Tessel with environment variable (e.g. "export TESSEL=bulbasaur") or use the --name flag.');
             // Close all opened connections and resolve
             controller.closeTesselConnections(foundTessels)
-              .then(resolve);
+              .then(() => resolve(foundTessels));
           });
       }
     });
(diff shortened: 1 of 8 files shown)


lib/controller.js
```diff
@@ -52,6 +52,10 @@ Tessel.list = function(opts) {
 
   return new Promise(function(resolve, reject) {
 
+    if (opts.verbose === false) {
```
johnnyman727: It's useful to disable all the logging from the test rig. Other API consumer may also appreciate this.

lib/controller.js
```diff
@@ -98,14 +102,14 @@ Tessel.list = function(opts) {
       } else if (foundTessels.length === 1) {
         // Close all opened connections and resolve
         controller.closeTesselConnections(foundTessels)
-          .then(resolve);
+          .then(() => resolve(foundTessels));
```
johnnyman727: For the sake of folks who want to do something like this (including the test rig):
```.js
t2.list().then((tessels) => {
...
});
```
It's much preferred to actually return the listed Tessels rather than nothing.

lib/tessel/wifi.js
```diff
@@ -158,7 +158,7 @@ Tessel.prototype.setWiFiState = function(enable) {
             // End the connection
             remoteProcess.close();
           }, Tessel.__wifiConnectionTimeout);
-          remoteProcess.stdout.once('data', function(data) {
+          remoteProcess.stdout.on('data', function(data) {
```
johnnyman727: I found an issue where sometimes the first data back from this command doesn't contain the exact substring we're searching for. In that case, we should keep listening and not reject immediately.

lib/tessel/wifi.js
```diff
@@ -185,11 +178,9 @@ Tessel.prototype.setWiFiState = function(enable) {
             }
           });
         } else {
-          this.connection.end();
```
johnnyman727: All connection ending should happen by consumers of this API. Otherwise things like the test rig can't run multiple commands on a device.

lib/usb_connection.js
```diff
@@ -110,6 +110,7 @@ USB.Connection.prototype.open = function() {
     // Try to open connection
     try {
       self.device.open();
+      self.closed = false;
```
johnnyman727: We should explicitly mark a Tessel as not closed when it's opened so you can open and close multiple times from a single API consumer.

lib/usb_connection.js
```diff
@@ -189,12 +190,10 @@ USB.Connection.prototype.end = function() {
...
     // on account of this connection
     Daemon.deregister(self, function(err) {
-      self._close();
       if (err) {
-        reject(err);
-      } else {
-        resolve();
+        return reject(err);
       }
+      self._close(resolve);
```
johnnyman727: This was very bad. We need to wait for the close to complete before continuing...

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "t2-cli",
   "version": "0.0.9",
   "description": "The starting point of Tessel 2's command line interface",
-  "main": "./bin/tessel-2.js",
+  "main": "./index.js",
```
johnnyman727: The file exported to API consumers should be one that actually exports `controller` functionality.

lib/usb_connection.js
LinusU: Nice :+1:

johnnyman727:
Closing in favor of smaller PRs (#567, #569, #570, and #571).

PR #558, opened 2016-02-10
remove self and bind

LinusU:
This removes the use of `self` and `bind`, and replaces it with the awesome fat arrow (`=>`).

I have also updated some of the functions where there was a lot of unnecessary wrapping with `new Promise`, but not all of the places. I wanted this pull request to focus on only `self` and `bind`.

I have also marked three `FIXME`s in the code, again it didn't seem right to fix it in this pull request; and in some cases I wanted the original author to maybe clarify the intent. I think we should merge with them in, and then fix them later.

I would appreciate a fast pull here since this will probably need manual rebasing as soon as anything else lands :)

ping @johnnyman727, I saw that you where online, would you like to review? Thanks :+1: 

edit: also, I need to go to bed right away, sorry about that, I'll check this first thing in the morning!

edit2: note to self to update this comment when landing, https://github.com/tessel/t2-cli/issues/470#issuecomment-160171309

PR #559, opened 2016-02-10
WIP Some questions

reconbot:
- I broke `enterBootloader` and no tests failed so I want to write more tests, this shouldn’t be broken in it’s current state
- is Glob an actual dev dependency?
- When I run grunt it dies during deploy.js but when I run nodeunit alone it’s fine, huh?

### DIFF
--- a/lib/usb_connection.js
+++ b/lib/usb_connection.js
@@ -221,42 +221,46 @@ USB.Connection.prototype._close = function(callback) {
   });
 };
 
-// Returns a device in DFU mode
-USB.Connection.prototype.enterBootloader = function() {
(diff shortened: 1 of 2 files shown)


LinusU:
I did something very similar to your current changeset in #558, it looks nice :+1:

LinusU:
`glob` is not a `devDependency`, but a normal `dependency`, and that is correct since it's only used under `lib/`.

Someone else will have to answer about the tests though :)

johnnyman727:
> I broke enterBootloader and no tests failed

What do you mean by this? How did you break it and how did you know it was broken?

Issue #560, opened 2016-02-11
Make it easier for newcomers to understand & start hacking on the CLI


flaki:
So when I got back into in the saddle with the Tessel 2 (especially with the CLI), I could practically be considered as a newcomer, I had to re-learn and re-discover the whole of the CLI & the source.

What I realized was that it was suprisingly hard to "uncover" the "magic" the CLI was hiding/abstracting away — but what it was doing most of the time is that it was channeling commands to the T2's OS. Those commands were:
 - new to me (as I've never worked with OpenWRT before, at least not in this depth), and
 - suprisingly hard to track (as in, in what situation what commands get executed)

I think, as new folks come onboard shortly we should be focusing not just on making it easy for them to work _with_ the Tessel 2 (as in, push code and operate it via the CLI), but also work *on* the Tessel 2, so they can jump in and improve the tooling itself (especially the CLI).

First step of that for me was to create some new debugging properties to better see what's going on:
https://github.com/flaki/t2-cli/tree/command-debug

Also, I was documenting my fumbling around T2 & OpenWRT-world, and this is what came out of it:
https://gist.github.com/flaki/a1efdb438292dd8c56c3
@Frijol - do you think this should live somewhere better-discoverable? If yes, where?
@HipsterBrown - please add anything in here you yourself have came across while developing the AP mode. ;)

Some other commands complementing the CLI would be nice as well, I'll be adding those to their own respective issues.

johnnyman727:
I think #561 is probably one of the most helpful contributions towards making things more transparent. Other things would be allowing folks to see `stdout`, and `stderr` data, when a particular process has closed, what it's exit code was, and whether any data was written to `stdin` for a process.

The `DEBUG=discovery:*` tag that exists is also pretty helpful for diagnosing connection issues.

Frijol:
@flaki thanks for making this issue! That gist is extremely useful - perhaps it can live in the Docs repo?

I have hopes that we can turn Docs into not just APIs but also docs for contributors. I tried separating them out on the Onboarding repo at first, but think this would be a better approach.

mulderp:
This is a nice list indeed! I like the `handy commands` or command cheatsheet.
Something similar would be interesting to have in the "Embedded Internet" chapter if the Embedded Node.js book, @flaki if interested I can pass you a link of the draft via Twitter DM and you could check whether/how your getting started things are relevant for the book.

johnnyman727:
[...]

PR #561, opened 2016-02-11
Debugging commands sent via USB/LAN

flaki:
For motivations for this, see #560 - for how this should work, see [this gist/doc](https://gist.github.com/flaki/a1efdb438292dd8c56c3).


johnnyman727:
This is super useful - I wish we added it months ago. :+1:

PR #562, opened 2016-02-11
Harmonize push/run exec dir + tests, fixes #544

### LINKED ISSUE #544: Inconsistent working directory for t2 push/run
The default working directory is inconsistent between `push`-ing and `run`-ning a project.
`process.cwd()` is `/root` instead of the project directory -- but only when using `t2 run`, `t2 push` sets it to the app directory (currently `/app/remote-script`).

flaki:
I fixed the tests, this seems to work properly.

Do you think we should need to add a test that specifically reviews #544's case and makes sure both functions remain consistent?


rwaldron:
Nice! I will land it as soon as it goes green :)

PR #567, opened 2016-02-12
Adds output silencing option

johnnyman727:
This is useful for consumers of the API exposed by the CLI (in an upcoming PR) because if you're using it programmatically, you likely don't want to clutter your output with all the things the CLI reports.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -24,22 +24,40 @@ var responses = {
 
 // Wrapper function for Tessel.list to set SSH key path
 controller.list = function(opts) {
-  return controller.keyHelper(opts, Tessel.list);
+  return controller.defaultHelpers(opts, Tessel.list);
 };
 
 // Wrapper function for Tessel.get to set SSH key path
 controller.get = function(opts) {
-  return controller.keyHelper(opts, Tessel.get);
+  return controller.defaultHelpers(opts, Tessel.get);
 };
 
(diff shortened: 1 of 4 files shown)


lib/controller.js
```diff
@@ -24,22 +24,40 @@ var responses = {
...
+controller.defaultHelpers = function(opts, next) {
+  // First disable output if necessary
+  return controller.outputHelper(opts)
+    // Then make sure our SSH Key path is updated
+    .then(() => controller.keyHelper(opts))
+    // before continuing to the next function
+    .then(() => next(opts));
+};
+
+controller.keyHelper = function(opts) {
```
rwaldron: Seems unrelated.

johnnyman727: Since there are two functions every command should call (`keyHelper` and `outputHelper`), I decided it made sense to just refactor into a single function that would be easier to build upon.

rwaldron: Sure, but I mean that it's unrelated to the PR's title :P "Adds output silencing option"?

johnnyman727: It was necessary because I wanted to add the function below (`outputHelper`) without making the code look too hairy. `keyHelper` originally set the default SSH key and then was responsible for calling the subsequent command (either `list` or `get`) but I figured it makes more sense for one function to call each of the helpers and then pass it on to the command.

PR #569, opened 2016-02-12
Jon fix wifi

johnnyman727:
I ran into an issue where sometimes the output from the function that would check the state of the wifi connection would not come all at once. In that case, the CLI would get the first packet which wouldn't contain the keyword that indicates success so it would immediately fail rather than wait for more data. This fixes that.

Also, this had an old `tessel.connection.end` lying around that messes up using the CLI API when you want to run multiple execs on a Tessel.


johnnyman727:
:clap:

PR #570, opened 2016-02-12
Exposes the CLI as an API

johnnyman727:

Allows consumers to use (for example)
```.js
var cli = require('t2-cli');

cli.list({timeout: 1, output: false, usb:true)
.then((tessels) => {
    ...
});
```


rwaldron:
Sorry, I thought I landed this hours ago :\

PR #571, opened 2016-02-12
wait for close, allow multiple opens

johnnyman727:
This PR fixes to minor things:
* The `_close` function was being called but no callback was provided to ensure it completes
* You could not open and close a USB connection more than once

This is part of the disintegration of #555 and is needed for the test rig.

### DIFF
--- a/lib/usb_connection.js
+++ b/lib/usb_connection.js
@@ -192,15 +192,15 @@ USB.Connection.prototype.setAltSetting = function() {
 };
 
 USB.Connection.prototype.end = function() {
+  var self = this;
   return new Promise((resolve, reject) => {
     // Tell the USB daemon to end all processes active
     // on account of this connection
-    Daemon.deregister(this, (err) => {
-      this._close();
+    Daemon.deregister(self, function(err) {
       if (err) {
         reject(err);
       } else {
-        resolve();
+        self._close(resolve());
       }
     });
   });
--- a/lib/usb_connection.js
+++ b/lib/usb_connection.js
@@ -111,6 +111,7 @@ USB.Connection.prototype.open = function() {
   // Try to open connection
   try {
     this.device.open();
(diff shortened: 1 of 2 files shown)


lib/usb_connection.js
```diff
@@ -189,11 +190,10 @@ USB.Connection.prototype.end = function() {
     // Tell the USB daemon to end all processes active
     // on account of this connection
     Daemon.deregister(self, function(err) {
-      self._close();
       if (err) {
         reject(err);
       } else {
-        resolve();
+        self._close(resolve());
```
LinusU: This won't work...

self._close(function (err) {
  if (err) reject(err)

  resolve()
})

johnnyman727: Thanks, I think I fixed it with https://github.com/tessel/t2-cli/commit/885f385804e69533b6c86fa623adb412acbeb474. It was actually concealing a bigger problem with the tests!

rwaldron:
LGTM

lib/usb_connection.js
```diff
@@ -196,33 +197,36 @@ USB.Connection.prototype.end = function() {
...
+    return;
   }
 
   this.closed = true;
+
   this.epIn.stopPoll(() => {
     var attempt = () => {
       try {
         this.device.close();
+        callback();
```
LinusU: @johnnyman727 I missed this the first time and now this has already landed, but this is a big no-no. Since the `callback` function is wrapped in the `try`, *any* error that occurs within the `callback` function will now silently be dropped and cause `attempt` to be called again.

What do you think of something like this?

```javascript
try {
  this.device.close()
} catch (_) {
  return setTimeout(attempt, 100)
}

cb(null)
```

johnnyman727: @LinusU great catch! I'll fix that immediately and submit a PR.

PR #578, opened 2016-02-17
Fixes update listing and upgrading

### LINKED ISSUE #564: `t2 update` will try to update to the earliest build
I recently wrote a script to automate releases. It updates `builds.json` on our build server which is how the CLI figures out which builds are available. The script updates the array of optional builds in reverse order than how I used to update the builds manually.

Rather than change the build script, the CLI should be able to handle the builds in any order. The best way to do this is to sort the builds as are returned to the CLI, before any consumers of this function call receive the data. In particular, it should happen [here](https://github.com/tessel/t2-cli/blob/17765e26dc4b3980c39061e8faee5323789b1146/lib/update-fetch.js#L40) before the function resolves. The rest of the CLI expects the most recent updates to be at the end of the array.

johnnyman727:
Fixes #564. Builds were being parsed out of order leading to `t2 update` trying to place something other than the latest build on the device. It also fixes an issue where `connection.exec` was trying to be used as a Promise.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -7,6 +7,7 @@ var async = require('async');
 var colors = require('colors');
 var inquirer = require('inquirer');
 var sprintf = require('sprintf-js').sprintf;
+var semver = require('semver');
 
 // Internal
 var discover = require('./discover');
@@ -846,7 +847,6 @@ controller.update = function(opts) {
           return versionFromSHA.then(function(currentVersionInfo) {
             var build = updates.findBuild(builds, 'version', version);
             var verifiedVersion;
-
             // If the update is forced or this version was requested,
             // and a valid build exists for the version provided.
             if (version && build) {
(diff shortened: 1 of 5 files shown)


flaki:
In `controller.js` this is still a string compare, so it fails on this line without force:
```
              if (!opts.force && currentVersionInfo.version >= verifiedVersion) {
```
(line 875 on my machine).

More debug info:
```
Flakis-Air:t2-cli flaki$ t2 update
INFO Connected to ft2.
build:  { released: '2016-02-17T02:15:26.937Z',
  sha: 'fa6e1c42c2d31848787ac56f12e1f4556ced22df',
  version: '0.0.11' }
currentVersionInfo:  { released: '2015-12-16T00:07:20.450Z',
  sha: 'd34f30a766081f11b46011c802ce64fe6559755a',
  version: '0.0.7' }
verifiedVersion:  0.0.11
WARN ft2 is already on the latest firmware version (0.0.7). You can force an update with "t2 update --force".
Flakis-Air:t2-cli flaki$ 
```

flaki:
Now the build is downloaded all right, but update fails with a mystery `TypeError` right after _"Updating OpenWRT (1/2)"_
```
Flakis-Air:t2-cli flaki$ t2 update
INFO Connected to ft2.
INFO New firmware version found...0.0.11
INFO Updating ft2 to latest version (0.0.11)...
INFO Download complete!
INFO Updating OpenWRT (1/2)
ERR! TypeError: Cannot read property 'allowHalfOpen' of undefined
```

PR #580, opened 2016-02-17
Eliminates misuse of connection.exec API

### LINKED ISSUE #579: connection.exec used improperly
There are two placed where `tessel.connection.exec` are being used improperly as a result of the switch from a Promise-based API to callbacks: in [name setting](https://github.com/tessel/t2-cli/blob/2fa6bb9ad074bdc199b0b723d8e25e0592c420ea/lib/tessel/name.js#L44-L50) and [provisioning](https://github.com/tessel/t2-cli/blob/2fa6bb9ad074bdc199b0b723d8e25e0592c420ea/lib/tessel/provision.js#L198). These calls should not return the result of `exec` because `exec` no longer returns a Promise.

johnnyman727:
Fixes #579.

### DIFF
--- a/lib/tessel/name.js
+++ b/lib/tessel/name.js
@@ -41,7 +41,7 @@ Tessel.prototype.setName = function(name) {
       })
       .then(() => {
         // Write the new name to the kernel hostname as well (so it reports the proper name in the terminal)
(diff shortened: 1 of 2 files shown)


johnnyman727:
Bonus: I believe this issue was causing `nodeunit` failures every once in awhile where the tests just never complete.

Issue #583, opened 2016-02-19
Syntax Error detection is wrong


rwaldron:
Patch to follow.

PR #584, opened 2016-02-19
Corrections to syntax error handling in compress operation. Fixes gh-583

rwaldron:
The correct algorithm:

1. Let ast be a.parse(source)
2. If ast is a syntax error (e), let m be e.message
    1. Set the value of ast to u.parse(source)
        1. If ast is a syntax error (e), append e.message to m.
            1. throw new SyntaxError(m)

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -973,6 +973,7 @@ actions.compress = function(source) {
   };
 
   var ast;
+  var errors = [];
 
   try {
     // Attempt to use acorn, this will provide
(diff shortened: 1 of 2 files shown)


johnnyman727:
Travis is failing because it can't download `nvm`:
```
$ nvm install 4.2.1
######################################################################## 100.0%
Checksums empty
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
Binary download failed, trying source.
$ node --version
v0.10.36
```
It's trying to install everything with Node `0.10.36` which just won't work...

johnnyman727:
I believe this is actually an issue with `nvm`. On an ubuntu 14.04 box, I get the same failure when I run `nvm install 4.2.1` but not with `nvm install stable`.

johnnyman727:
All tests pass on my machine so I'm going to merge this while we fix Travis.

Issue #587, opened 2016-02-20
FRE Tweet example doesn't work


johnnyman727:
When I walk through the [First Run Experience's Tweet example](http://tessel.github.io/t2-start/tweet.html), my Tessel fails to deploy the code. It gets stuck with:
```
➜  tessel-tweet t2 run tweet.js
INFO Looking for your Tessel...
INFO Connected to Hose.
INFO Building project.
```
It does not hang if I run it with the `--full` flag.

Workaround is `t2 run tweet.js --full`.

johnnyman727:
The error is being thrown [here](https://github.com/tessel/t2-cli/blob/e016516aed2ea76ed50605a79d4b75fd61ffc68f/lib/tessel/deploy.js#L797) but it's being swallowed for some reason. 

When I print out the stack trace for the error, it looks like:
```
TypeError: self.key.print is not a function
    at eval (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:4826:18)
    at doit (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:4067:13)
    at AST_Node.eval [as print] (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:4073:13)
    at eval (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:4793:22)
    at Array.forEach (native)
    at eval (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:4787:29)
    at eval (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:3946:19)
    at OutputStream.with_indent (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:3920:38)
    at Object.with_block (eval at <anonymous> (/Users/Jon/Work/technical/t2-cli/node_modules/uglify-js/tools/node.js:22:1), <anonymous>:3945:9)
[...]

PR #588, opened 2016-02-20
Deploy: further refine handling of "unparseable" sources. Fixes gh-587

rwaldron:
The algorithm defined in https://github.com/tessel/t2-cli/commit/e6639b4fd9508269dc8f4a38b0a6bc437accbd4f was an improvement, but it's clear that we'll play "whack-a-mole" forever if we attempt to hand-hold too much. After reading #587, the obvious step was to simply NOT compress code that could not be parsed by uglify or acorn. If the code truly contains a syntax error, let the runtime itself handle that.

Taking this approach, which relaxes the potential for failures also addresses the outstanding issue with ES6 syntax being unparseable by acorn and/or uglify. Now, if either fail to parse because of an arrow function (or anything), the deployment is no longer terminated.


HipsterBrown:
Pragmatic and perfect. 👍

johnnyman727:
@rwaldron agreed that this is a better route. Will this have the additional advantage of showing a more clear and recognizable stack trace if there is a syntax error?

In any case, if it passed Grunt on your machine, I would say merge it and put out a new release. I'll investigate why nvm is failing for Travis when I get back this weekend.

rwaldron:
>  Will this have the additional advantage of showing a more clear and recognizable stack trace if there is a syntax error

Actually, yes it will—I hadn't even thought of that. Because acorn and uglify will fail, the code will remain in its original form and the runtime failure will display it as the author wrote it.

Issue #470, opened 2015-11-26
Decide which version of Node.js to target


LinusU:
As noted in #467 the code currently doesn't work because we are using template strings which isn't available in Node.js < 4 (I don't know about iojs). On the other hand, there are a lot of features in Node 4.x that we aren't taking advantage of.

I think that we should take a decision on which version to target, and then use all the new features available on that version.

The largest benefit would be fat-arrow functions which would eliminate the need for `self = this` sprinkled all over the code. It would also eliminate most uses of `.bind(...)`.

Me and @Student007 had a small discussion about it here: https://github.com/tessel/t2-cli/pull/466#discussion_r45955570

*(note that this is the version of Node that the CLI should run on, it has nothing to do with code on the actual tessel)*

**edit:**

Expected outcome:

- [ ] Add `'use_strict'` at the top of files (this is needed for next step)
- [ ] Switch all `var` to `const` and `let`
- [ ] Stop using `self = this` and `.bind(...)` in favour of `=>`
- [ ] Use template literals in favour of `+`-ing together strings
- [ ] Use `class` and `extend` keyword for making classes (instead of the cumbersome `function`, `prototype = Object.create`, `prototype.x = _`)
- [x] Use the builtin `Promise` (~~maybe we already are?~~ indeed :+1:)
- [ ] Mark a minimum `engine` in the `package.json` (e.g. `>4.0.0`, which is the current LTS)

rwaldron:
Node.js 4.2.x should be the minimum, as I said in the other thread: managing multiple versions is trivial with nvm.

LinusU:
As said in #467, 0.12 is only in LTS mode until the end of the year, which is very soon. Me and @rwaldron therefor think that we should target Node 4.x and newer.

I think that we should go thru the codebase with the following fixes:
- [ ] Add `'use_strict'` at the top of files (this is needed for next step)
- [ ] Switch all `var` to `const` and `let`
- [ ] Stop using `self = this` and `.bind(...)` in favour of `=>`
- [ ] Use template literals in favour of `+`-ing together strings
- [ ] Use `class` and `extend` keyword for making classes (instead of the cumbersome `function`, `prototype = Object.create`, `prototype.x = _`)
- [ ] Use the builtin `Promise` (maybe we already are?)

I would love to help out with this!

Student007:
@johnnyman727 @LinusU @rwaldron 
[...]

PR #591, opened 2016-02-21
package: update engines

LinusU:
Update the required version of Node.js to 4.2 (the current LTS release, "Argon") as per discussion in #470.

I also removed the `engine-strict` because 1) it's called `engineStrict` so it wasn't used and 2) [it's deprecated](https://docs.npmjs.com/files/package.json#enginestrict) and isn't respected anymore.

*@rwaldron: In the other issue you wrote 4.2.1, but I couldn't see anything in the changelog between 4.2.0 and 4.2.1 that would indicate that we really don't work on 4.2.0. I think that we should either say "the latest LTS" (4.3.1) or "any version of the current LTS" (4.2.0). Did I miss anything?*

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,8 +79,6 @@
   },
   "preferGlobal": true,
   "engines": {
-    "node": ">0.12.0",
-    "iojs": ">= 1.0.0"
-  },
-  "engine-strict": true
+    "node": ">=4.2.0"
+  }
 }


rwaldron:
The Travis failure is bogus

LinusU:
Thanks for the quick pull :+1:

rwaldron:
😬

Issue #592, opened 2016-02-21
Possible problems with the tests


LinusU:
I've been reading thru the testing code and found quite a few things that we should probably look at.

1) The use of `test.fail`

`test.fail` isn't documented at all in `nodeunit`, rather, it's an internal method with the signature of `function fail(actual, expected, message, operator, stackStartFunction)`. However I've seen a number of places where it's being used like this: `test.fail('Something bad happend')`.

2) Throwing inside promises

Another thing that seems to happen quite a lot, is throwing inside promises. This is often hidden by a call to `test.*`. Problematic code:

```javascript
provision.setDefaultKey(testPath)
  .then(...)
  .catch((err) => {
    test.ok(false, 'Something bad happened')
  })
```

The problem with this code is that `test.ok` internally calls `test.fail`, which will throw an `AssertionError`. The idea with nodeunit is that it listens for this error and then records it as a failure. But in this example, the only thing that will happen is that the promise returned from the `setDefaultKey` call, will be recorded as "rejected".

I've also seen the following code in a few places which I think is a way to work around that you can't actually throw inside a promise handler...

```javascript
fs.remove(testDir, (err) => {
  if (err) {
    console.log('Error removing tmp directory: ' + err);
    process.exit();
  }

  // ...
});
```

Had this been outside any promise, it could have been written as `if (err) throw err;` or even `test.ifError(err)`. (the naming is `ifError` because it usually lives on an object called `assert`. `assert.ifError` makes sense while `test.ifError` sounds like it will test if the parameter is an error)

3) `.then(test.fail)`

This doesn't actually do anything, a product of both of the above problems. Example:

```javascript
    // Attempt to provision
    this.tessel.provisionTessel()
      // If an error is not thrown, this test failed
      .then(test.fail)
```

4) `if (err) test.fail()`

This hides the message about why it failed. It should be replaced with `test.ifError(err)`

5) Assigning unused stuff to `this`

Many tests seems to assign quite a lot of stuff to `this` (the test context) that never gets used. ~~The one thing I can think of is that it was used before the use of `sinon.sandbox` to restore the stubs.~~ It's probably done because some of these gets the `.callCount` checked. It would be nice to remove the assignment to make the code more readable.

6) The failing test isn't shown

To me this is a very unhelpful message:

```text
Running "nodeunit:tests" (nodeunit) task
Testing access-point.js......OK
Testing bin-tessel-2.js..........................OK
Testing constructor.js......OK
Testing controller.js..............................OK
[...]

PR #593, opened 2016-02-21
test: remove self and bind

LinusU:
Building on #558, this does the same thing but for the `test/` directory. Now every file in the project are updated :checkered_flag: 

I discovered a lot of problems with the test though, see #592 :sob: 

I built this as one pull request since it should be easy to review, and since there are just two pull requests open at the moment, so the chance of conflict should be low :clap: 

@rwaldron Are you still online, care to review? :heart:


rwaldron:
Nice!

PR #594, opened 2016-02-21
Eliminate use of test.fail; enforce test.done on all errors and failures. Related to gh-592

rwaldron:


rwaldron:
I'm going to back through this again tomorrow before any reviews

rwaldron:
I hadn't seen https://github.com/tessel/t2-cli/pull/593 so I'll figure out how to resolve the two tomorrow

test/unit/controller.js
```diff
@@ -512,13 +514,13 @@ exports['Tessel.list'] = {
 exports['Tessel.get'] = {
 
   setUp: function(done) {
-    var testContext = this;
+    var self = this;
```
LinusU: Any specific reason why you are changing this back?

test/unit/controller.js
```diff
@@ -894,7 +905,9 @@ exports['controller.root'] = {
 
     this.child = new FakeChild();
 
-    this.spawn = this.sandbox.stub(cp, 'spawn', () => this.child);
+    this.spawn = this.sandbox.stub(cp, 'spawn', () => {
+      return this.child;
+    });
```
LinusU: Any specific reason for changing this back?

test/unit/controller.js
```diff
@@ -992,17 +1007,20 @@ exports['controller.root'] = {
...
       .then(() => {
         // Only spawn one process
-        test.fail('Should have rejected the root promise.');
+        test.ok(false, 'Should have rejected the root promise.');
+        test.done();
       })
-      .catch((err) => {
-        test.ok(err.indexOf(errMessage) !== -1);
+      .catch(error => {
+        test.ok(error.indexOf(errMessage) !== -1);
```
LinusU: `error.includes(errMessage)`?

test/unit/deploy.js
```diff
@@ -157,36 +157,42 @@ exports['Tessel.prototype.deployScript'] = {
...
-          INFO Writing app.js to RAM on arnold (2.048 kB)...
-          INFO Deployed.
-          INFO Running app.js...
-          testing deploy
-          INFO Stopping script...
-      */
-        test.equal(bundle.length, 2048);
-        deleteTemporaryDeployCode().then(function() {
+          target: deployFolder
+        }).then(bundle => {
```
LinusU: Instead of nesting here, you should probably just `return deploy.tarBundle(...)` and then run `then` on the main promise.

LinusU:
Instead of using `test.ok(false, ...)` which will trigger an AssertionError saying that `false` should equal `true`, how about using `test.ifError(...)` which will trigger the error in question instead.

Also, there is no need for `test.done` after `test.ok(false, ...)` since it will stop the flow of execution (by using throw).

There is also still the problem that most of these errors will just silently be dropped since they are inside a promise handler...

test/unit/deploy.js
```diff
@@ -283,11 +298,14 @@ exports['Tessel.prototype.deployScript'] = {
...
 
-    this.exec = sandbox.stub(this.tessel.connection, 'exec', (command, cb) => cb(null, this.tessel._rps));
+    this.exec = sandbox.stub(this.tessel.connection, 'exec', (command, callback) => {
+      return callback(null, this.tessel._rps);
+    });
 
     deploy.writeToFile(this.tessel, 'index.js');
   },
 
+
```
LinusU: Why two empty lines here?

test/unit/controller.js
rwaldron: No, but it's possible that `jsbeautify` did that. You should've run it anyway :P

rwaldron: No, I had to do some delicate rebasing with the first version of this branch and your self/bind branch. No big deal though, I will fix this

test/unit/controller.js
rwaldron: Same reason as below

rwaldron: (fixed locally, will push all together)

test/unit/controller.js
rwaldron: (fixed locally, will push all together)

test/unit/controller.js
rwaldron: Sure!

(fixed locally, will push all together)

test/unit/deploy.js
rwaldron: I don't follow this, can you do a fixup after this patch is done?

rwaldron:
> There is also still the problem that most of these errors will just silently be dropped since they are inside a promise handler...

How does this affect that actual functionality being tested, as it is currently?


> Also, there is no need for test.done after test.ok(false, ...) since it will stop the flow of execution (by using throw).

That's not really true. I created this test file: 

```js
exports['ok-false'] = {
  setUp: function(done) {
    done();
  },

  tearDown: function(done) {
    done();
  },

  okTrue: function(test) {
    test.expect(1);
    test.ok(true, 'this is true');
    test.done();
  },

  okFalse: function(test) {
    test.expect(1);
    test.ok(false, 'this is false');
    test.done();
  },

  okTrueAgain: function(test) {
    test.expect(1);
    test.ok(true, 'this is true again');
    test.done();
  }
};
```

Which produces this output: 

```
$ grunt nodeunit:only:test-ok-false-and-done
Running "nodeunit:only:test-ok-false-and-done" (nodeunit:only) task

Testing test-ok-false-and-done.js.F.
>> ok-false - okFalse
>> Message: this is false
>> Error: false == true
>> at Object.exports.ok-false.okFalse (test/unit/test-ok-false-and-done.js:18:10)
>> at Object.exports.ok-false.setUp (test/unit/test-ok-false-and-done.js:3:5)

Warning: 1/3 assertions failed (807ms) Use --force to continue.

Aborted due to warnings.
```


Here's the verbose output: 

```
Testing test-ok-false-and-done.js
ok-false - okTrue...OK
ok-false - okFalse...ERROR
>> Message: this is false
>> Error: false == true
>> at Object.ok (node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/lib/types.js:83:39)
>> at Object.exports.ok-false.okFalse (test/unit/test-ok-false-and-done.js:18:10)
>> at /Users/rwaldron/clonez/t2-cli/node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/lib/core.js:232:20
>> at /Users/rwaldron/clonez/t2-cli/node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/deps/async.js:168:13
>> at /Users/rwaldron/clonez/t2-cli/node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/deps/async.js:131:25
>> at /Users/rwaldron/clonez/t2-cli/node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/deps/async.js:165:17
>> at /Users/rwaldron/clonez/t2-cli/node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/deps/async.js:463:34
>> at Object.exports.ok-false.setUp (test/unit/test-ok-false-and-done.js:3:5)
>> at /Users/rwaldron/clonez/t2-cli/node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/lib/core.js:260:35
>> at /Users/rwaldron/clonez/t2-cli/node_modules/grunt-contrib-nodeunit/node_modules/nodeunit/deps/async.js:458:21

ok-false - okTrueAgain...OK
Warning: 1/3 assertions failed (787ms) Use --force to continue.

Aborted due to warnings.
```

Notice that in both cases, all three tests complete and the final result of all three is displayed? If I remove a `test.done()`: 

```js
exports['ok-false'] = {
  setUp: function(done) {
    done();
  },

  tearDown: function(done) {
    done();
  },

  okTrue: function(test) {
    test.expect(1);
    test.ok(true, 'this is true');
    test.done();
  },

  okFalse: function(test) {
    test.expect(1);
    test.ok(false, 'this is false');
    // test.done();
  },

  okTrueAgain: function(test) {
    test.expect(1);
    test.ok(true, 'this is true again');
    test.done();
  }
};
```

Here's what happens: 

```
$ grunt nodeunit:only:test-ok-false-and-done
Running "nodeunit:only:test-ok-false-and-done" (nodeunit:only) task

Testing test-ok-false-and-done.js.F
>> Incomplete tests/setups/teardowns:
>> ok-false - okFalse 0 [ 'ok-false - okFalse' ]
Fatal error: A test was missing test.done(), so nodeunit exploded. Sorry!
```


And execution continues as expected, for example: 

```js
exports['no-execution-stop'] = {
  setUp: function(done) {
    done();
  },

  tearDown: function(done) {
    done();
  },

  okFalse: function(test) {
    test.expect(1);
    test.ok(false, 'this is false');
    console.log('AFTER OK(false)');
    test.done();
  },
};
```


Results in...

```
$ grunt nodeunit:only:test-ok-false-and-done
Running "nodeunit:only:test-ok-false-and-done" (nodeunit:only) task

Testing test-ok-false-and-done.jsAFTER OK(false)
F
>> no-execution-stop - okFalse
>> Message: this is false
>> Error: false == true
>> at Object.exports.no-execution-stop.okFalse (test/unit/test-ok-false-and-done.js:12:10)
>> at Object.exports.no-execution-stop.setUp (test/unit/test-ok-false-and-done.js:3:5)

Warning: 1/1 assertions failed (1031ms) Use --force to continue.

Aborted due to warnings.
```

test/unit/deploy.js
rwaldron: Garbage from merge conflict resolution

rwaldron:
https://github.com/caolan/nodeunit/blob/d8b4f70c868d7c840f39f9e8ef7a848276169138/lib/core.js#L61-L75

test/unit/deploy.js
rwaldron: I'll make the change because you made the effort to review and to mention it, but I don't think it's really worth breaking any backs over.

johnnyman727:
@LinusU I'm not sure that running tests inside promises is actually an issue. For example, if you have the code below where a test fails inside the promise, it will actually enter the resolved branch rather than the rejected branch:
```.js
exports["Some Test"] = {
  setUp: function(done) {
    done();
  },
  tearDown: function(done) {
    done();
  },
  testOne: function(test) {
    var p = new Promise((resolve) => {
      test.ok(false);
      resolve();
    });

    return p.then(() => {
      console.log('resolved...');
      test.done();
    })
    .catch(() => {
      console.log('rejected');
      test.done();
    });
  }
};
```
`test.ok(false)` isn't actually throwing an Assertion Error.

johnnyman727:
@rwaldron looks good to me

LinusU:
It seems that I stand corrected. `test.ok(false)` is throwing an AssertionError, but it's also caught and stored in a list for later (`wrapAssert`), and the execution continues.

This is a bit strange for me since it can cause some very strange bugs if you aren't keeping this fact in mind. The problem I see with it is that it completely defeats the purpose of `ifError`. Consider this example.

```javascript
function asyncThing (cb) {
  setImmediate(cb, new Error('test'))
}

exports.testSomething = function (test) {
  asyncThing(function (err, res) {
    test.ifError(err)
    test.equal(res.thing, 'Hello')
    test.done()
  })
}
```

I'm expecting `asyncThing` to return `{ thing: 'Hello' }` but instead it errors. The problem is that this is the output:

```text
a

FAILURES: Undone tests (or their setups/teardowns): 
- testSomething

To fix this, make sure all tests call test.done()
```

The problem is that I've triggered a `TypeError: Cannot read property 'thing' of undefined`, but this get's hidden. Also, in the following example, I've misspelled `message` and thus it will trigger an error when I'm checking if the word `Hello` is in the error message.

```javascript
exports.testSomething = function (test) {
  new Promise(function (resolve, reject) {
    reject(new Error('Expect me: Hello'))
  }).then(function () {
    test.ok(false, 'Expected to throw error')
    test.done()
  }).catch(function (err) {
    test.ok(err.mossage.match(/Hello/))
    test.done()
  })
}
```

This error however only causes the promise to reject (again) with a `TypeError`, which can be seen if I add another `.catch` handler, on top of the other one.

```javascript
.catch(function (err) {
  console.error(err.stack)
})
```

```text
TypeError: Cannot read property 'match' of undefined
    at /private/tmp/nodeunittest/test/a.js:23:24
    at runMicrotasksCallback (node.js:360:7)
    at nextTickCallbackWith0Args (node.js:453:9)
    at process._tickCallback (node.js:382:13)
```

To be honest I'm not entirely sure where I'm heading with this since I'm very tired, it was just unexpected how it worked to me. As long as we are aware how this works, it shouldn't be too much of a problem. I don't see how we can use `ifError` though, unfortunately since that is quite nice.

Sorry for debating so much, I just want good code that's all :)

johnnyman727:
@LinusU I was surprised to figure that out as well and I totally see your point about how it would cause more errors in some cases. I suppose the benefit is that it *does* continue to run all your tests so that any any given time you get a maximum number of errors with your code. The alternative is that you fix that bug and then you've got several more to fix that were never run in the first place. :shrugs: they both seem reasonable to me.

PR #595, opened 2016-02-22
Allow updating with just a USB connection

johnnyman727:
Dependent on: https://github.com/tessel/openwrt-tessel/pull/47

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -309,6 +309,17 @@ makeCommand('update')
     flag: true,
     help: 'Update to the latest version regardless of current version.'
   })
+  .option('openwrt-path', {
+    abbr: 'op',
+    required: false,
+    flag: false,
+    help: 'Update with the OpenWRT image at the indicated local path.'
+  })
+  .option('firmware-path', {
+    abbr: 'fp',
+    required: false,
+    help: 'Update with the firmware image at the indicated local path.'
+  })
   .callback(function(opts) {
     if (opts.list) {
       callControllerWith('printAvailableUpdates');
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -808,117 +808,124 @@ controller.printAvailableUpdates = function() {
 
 controller.update = function(opts) {
   opts.authorized = true;
-  opts.lanPrefer = true;
   return controller.standardTesselCommand(opts, function(tessel) {
-    return new Promise(function updateProcess(resolve, reject) {
-      // If it's not connected via USB, we can't update it
-      if (!tessel.usbConnection) {
-        return reject('Must have Tessel connected over USB to complete update. Aborting update.');
-      }
-
-      // // If this Tessel isn't connected to the LAN
-      if (!tessel.lanConnection || !tessel.lanConnection.authorized) {
-        // Reject because USB updates are broken...
-        return reject('No LAN connection found. USB-only updates do not work yet. Please ensure Tessel is connected to wifi and try again');
-      }
-
-      return updates.requestBuildList().then(function(builds) {
-
-          var version = opts.version || 'latest';
-          var versionFromSHA = Promise.resolve(version);
-
-          // If we aren't forcing, we'll want to get the current SHA on Tessel
-          if (!opts.force) {
-            // Once we have the Tessel
-            // Over-ride the resolved Promise
-            versionFromSHA = new Promise(function(resolve, reject) {
-              // Figure out what commit SHA is running on it
-              return tessel.fetchCurrentBuildInfo()
-                // Once we have the current SHA, provide the version
-                .then(function(currentSHA) {
-                  return resolve(updates.findBuild(builds, 'sha', currentSHA));
-                })
-                .catch(function(err) {
-                  // If there was an error because the version file doesn't exist
-                  if (err.message.search('No such file or directory') !== -1) {
-                    // Warn the user
-                    logs.warn('Could not find firmware version on', tessel.name);
-
-                    if (opts.force !== false) {
-                      // Force the update
-                      opts.force = true;
-                      // Notify the user
-                      logs.warn('Forcefully updating...');
-                      // Resolve instead of reject (the string isn't used anywhere)
-                      return resolve('unknown version');
-                    } else {
-                      // Reject because the user specifically did not want to force
-                      return reject(err);
-                    }
-                  } else {
-                    // Reject because an unknown error occurred
-                    return reject(err);
-                  }
-                });
-            });
-          }
-
-          return versionFromSHA.then(function(currentVersionInfo) {
-            var build = updates.findBuild(builds, 'version', version);
-            var verifiedVersion;
-            // If the update is forced or this version was requested,
-            // and a valid build exists for the version provided.
-            if (version && build) {
-              // Fetch and Update with the requested version
-              return controller.updateTesselWithVersion(opts.force, tessel, currentVersionInfo.version, build);
-            } else {
-              // If they have requested the latest firmware
-              if (version === 'latest') {
-                build = builds[builds.length - 1];
-                verifiedVersion = build.version;
-              } else {
-                // They provided a valid version that matches a known build.
-                if (build) {
-                  verifiedVersion = build.version;
-                }
-              }
-
-              // If we've reached this point and no verified version has not
-              // been identified, then we need to abord the operation and
-              // notify the user.
-              if (!verifiedVersion) {
-                return reject('The requested build was not found. Please see the available builds with `t2 update -l`.');
-              }
-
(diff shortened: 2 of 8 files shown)


johnnyman727:
On weird issue I'm facing is that running `sysupgrade /tmp/openwrt.bin` does not seem to work. The only thing it seems to do is kill the SPI Daemon permanently so perhaps it is beginning to restart? @kevinmehall what do you think could be happening to keep `sysupgrade` from completing when I `exec` it from the USB Daemon?

kevinmehall:
Try adding `spid` and `usbexecd` to the whitelist of processes that sysupgrade won't kill when starting the update: https://github.com/tessel/openwrt/blob/2015-10-26/package/base-files/files/lib/upgrade/common.sh#L128.

kevinmehall:
You can just modify `/lib/upgrade/common.sh` on a Tessel, and if it works then patch the linked file in the openwrt repo.

johnnyman727:
Thanks, I'll give that a try.

johnnyman727:
Unfortunately, modifying `/lib/upgrade/common.sh` didn't seem to work. 

Strangely, `ubus call system upgrade` is a `ubus` option but it doesn't seem to do anything either.

But I *could* run either `require('child_process').exec('sysupgrade /tmp/openwrt.bin')` and `require('child_process').spawn('sysupgrade, ['/tmp/openwrt.bin'])` with positive results so I guess we could do that worst case...

bin/tessel-2.js
```diff
@@ -289,6 +289,17 @@ makeCommand('update')
     flag: true,
     help: 'Update to the latest version regardless of current version.'
   })
+  .option('openwrtPath', {
+    abbr: 'op',
+    required: false,
+    flag: false,
+    help: 'Update with the OpenWRT image at the indicated local path.'
+  })
+  .option('firmwarePath', {
```
rwaldron: We should avoid camel casing command line options :P

lib/controller.js
```diff
@@ -789,117 +789,126 @@ controller.printAvailableUpdates = function() {
...
+    }
+  });
+};
 
-      return updates.requestBuildList().then(function(builds) {
+controller.updateWithLocalBuilds = function(opts, tessel) {
+  return updates.loadLocalBinaries(opts)
+    .then((images) => {
+      return tessel.update(images);
+    })
```
rwaldron: Can just be `(images) => tessel.update(images)`

lib/tessel/update.js
```diff
@@ -53,33 +64,36 @@ Tessel.prototype.updateOpenWRT = function(image) {
...
         });
       });
     });
 };
 
 Tessel.prototype.updateFirmware = function(image) {
-  logs.info('Updating firwmare (2/2)');
-  // This must be USB connection
+  return new Promise((resolve, reject) => {
+    logs.info('Updating firmware (2/2)');
```
rwaldron: Why does this happen in the promise?

johnnyman727: I was just having issues with Promises resolving when I thought they should. I can revert this change if need be.

johnnyman727:
I think something else must be going on because I was able to successfully sysupgrade from the CLI with [this patch](https://github.com/tessel/t2-cli/commit/be9338d942467e84b315e43316b9602521658d65) and the `t2-cli usb-exec sysupgrade /tmp/openwrt.bin`... That should be essentially exactly the same as what happens in the `t2 update` command so there is something else causing the issue.

johnnyman727:
I've confirmed that two changes are necessary:
1. The patch as outlined by @kevinmehall above
2. Not closing the child process that `exec`ed the `sysupgrade` for about two minutes (or until it reports that the upgrade completed). 

Because I did not implement 2, it seemed as though 1 made no difference but with both implementations, the update succeeds.

johnnyman727:
Working on the OpenWRT patch [here](https://github.com/tessel/openwrt/commit/1c1896abbed88b7d10e0a3dc5466c288bb05f093) but the documentation on patches is pretty lacking. @kevinmehall can you look it over for me when you get a chance?

kevinmehall:
The file is in the openwrt tree, so just modify it in a normal commit. Patches are used to make changes to 3rd-party code that is downloaded as an upstream tarball and doesn't live in the repository. (The directory where you added that patch is supposed to be for patches against the uClibc source)

johnnyman727:
Ah okay, that makes much more sense. Thanks for the clarification.

johnnyman727:
So this feature introduces a fun complication which is that because we have to update our OpenWRT build to make updating over USB work, all previous builds will not have this capability. That means we have to check the version running on the Tessel and confirm that it's > 0.0.11 before moving forward with a *USB-only* update.

@rwaldron any thoughts on that? It's kind of shitty but I'm not sure we have a good way to get around it.

kevinmehall:
If you want a hack, the CLI could modify `/lib/upgrade/common.sh` on old versions before performing the update...

johnnyman727:
@kevinmehall I'm ashamed at you for even suggesting such a dirty hack.

But yeah, does `sed '/*procd*\|*ash*\|*init*/c \ \*procd*|*ash*|*init*|*watchdog*|*ssh*|*dropbear*|*telnet*|*login*|*hostapd*|*wpa_supplicant*|*nas*|*spid*|*usbexecd*) : ;;'\
 /lib/upgrade/common.sh `

kevinmehall:
<a href=http://xkcd.com/1197></a>

Don't you need `sed -i` for in-place edit? Or write to a temp file and `mv` over the original if `busybox` sed doesn't have that option.

If you were to include the whole original line, it would be a no-op in the case that the update had already been applied. Also, watch the escaping of the `*` characters in addition to the `|`.

johnnyman727:
> Don't you need sed -i for in-place edit? Or write to a temp file and mv over the original if busybox sed doesn't have that option.

Yes, I do and busybox sed does support it. 

> If you were to include the whole original line, it would be a no-op in the case that the update had already been applied. Also, watch the escaping of the * characters in addition to the |.

Ugh I don't think my fingers can type that many escapes. It's not physically possible.

johnnyman727:
For anyone who's checking in on this PR, it's currently blocked by my effort to add a hacky to fix an issue [described here](https://github.com/tessel/t2-cli/pull/595#issuecomment-193591352) so that folks running older versions of the T2 firmware can do updates over USB.

[The command to overwrite a specific line](https://github.com/tessel/t2-cli/blob/jon-update-usb/lib/tessel/commands.js#L160-L165) of a file which determines how the update process works doesn't seem to be effective over the USB Daemon but it does work on the Tessel shell console. I'm guessing the escapes aren't being handled properly...

rwaldron:
There must be an easier way to do that :\

johnnyman727:
I've tried all kinds of escaping with no luck. I may go the kinda hackier route and just ship the file with the CLI, then write it to a temporary file on T2, then rename the temporary file to the real file name (thus replacing it).

lib/tessel/update.js
```diff
@@ -27,59 +28,108 @@ Tessel.prototype.fetchCurrentBuildInfo = function() {
...
+          if (err) {
+            return reject(err);
+          }
 
-        logs.info('Transferring image of size', (image.length / 1e6).toFixed(2), 'MB');
-        // When we finish writing the image
-        remoteProc.once('close', resolve);
-        // Write the image
-        remoteProc.stdin.end(image);
+          logs.info('Transferring image of size', (image.length / 1e6).toFixed(2), 'MB. This will take 2-4 minutes...');
```
rwaldron: Use template string

rwaldron:
Damn dude, that was smoooooooth


```
$ t2 update
INFO Connected to Tessel-02A347EF35F2.
INFO New firmware version found...0.0.12
INFO Updating Tessel-02A347EF35F2 to latest version (0.0.12)...
INFO Download complete!
INFO Updating OpenWRT (1/2)
INFO Transferring image of size 19.14 MB. This will take 2-4 minutes...
INFO Transfer complete.
INFO Starting OpenWRT update.
INFO Please do not remove power from Tessel.
INFO This process will take at least two minutes...
INFO Updating firmware (2/2)
INFO Firmware update complete!
INFO Updated Tessel-02A347EF35F2 from  0.0.11  to  0.0.12
```


First time, with new Tessel 2 out of the box.

rwaldron:
Unplugged, re-plugged, ran t2 wifi setup and we're in business.

Please address the nit-picks and I will land this tomorrow

bin/tessel-2.js
johnnyman727: :+1:

lib/controller.js
johnnyman727: beautiful

johnnyman727:
Nit-picks addressed!

PR #596, opened 2016-02-23
Access Point Refactor && Update

### LINKED ISSUE #536: Remove --trigger flag from ap command
In order to have usage parity with `t2 wifi`, `t2 ap` should have the `--trigger` option removed (used to turn the access point on and off) so that the usage is (`t2 ap --on`) rather than (`t2 ap --trigger on`). See https://github.com/tessel/t2-cli/issues/528 for more discussion.

HipsterBrown:
Fixes #536 

So the new method of toggling an access point is using `t2 ap --on/--off`, just like `t2 wifi --on/--off`.

This PR also add functionality similar to `t2 wifi`, i.e.,

```
Nicks-MBP:tessel-test hipsterbrown$ t2 ap
INFO SSID: RouterName
INFO Password: APPassword
INFO Security: psk2
INFO IP Address: 192.168.1.101
INFO State: Enabled
```

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -716,28 +716,47 @@ controller.createAccessPoint = function(opts) {
       }
 
       if (security && securityOptions.indexOf(security) < 0) {
-        reject(security + ' is not a valid security option. Please choose on of the following: ' + securityOptions.join(', '));
+        reject(`${security} is not a valid security option. Please choose on of the following: ${securityOptions.join(', ')}`);
       }
       resolve();
     })
     .then(function() {
-      return controller.standardTesselCommand(opts, function(tessel) {
-        return tessel.createAccessPoint(opts);
-      });
+      return controller.standardTesselCommand(opts, (tessel) => tessel.createAccessPoint(opts));
     });
 };
 
 controller.enableAccessPoint = function(opts) {
   opts.authorized = true;
-  return controller.standardTesselCommand(opts, function(tessel) {
-    return tessel.enableAccessPoint();
-  });
+  return controller.standardTesselCommand(opts, (tessel) => tessel.enableAccessPoint());
 };
 
 controller.disableAccessPoint = function(opts) {
   opts.authorized = true;
+  return controller.standardTesselCommand(opts, (tessel) => tessel.disableAccessPoint());
+};
+
+controller.getAccessPointInfo = function(opts) {
+  opts.authorized = true;
   return controller.standardTesselCommand(opts, function(tessel) {
-    return tessel.disableAccessPoint();
+    return tessel.getAccessPointInfo()
+      .then((ap) => {
+        if (ap.ssid) {
+          logs.info(`SSID: ${ap.ssid}`);
+
+          if (ap.key || ap.encryption !== 'none') {
+            logs.info(`Password: ${ap.key}`);
+          }
+
+          logs.info(`Security: ${ap.encryption}`);
+          logs.info(`IP Address: ${ap.ip}`);
+          logs.info(`State: ${(!Number(ap.disabled) ? 'Enabled' : 'Disabled')}`);
+        } else {
+          logs.info(`${tessel.name} is not configured as an access point (run "t2 ap --help" to learn more)`);
+        }
+      })
+      .catch((error) => {
+        throw error;
+      });
   });
 };
 
--- a/lib/tessel/access-point.js
+++ b/lib/tessel/access-point.js
@@ -10,47 +10,70 @@ var logs = require('../logs');
 var Tessel = require('./tessel');
 
 
-function commitAndClose(tessel, status, resolve) {
+function commitAndClose(tessel, status, resolve, reject) {
 
-  var reconnectWifi = function() {
-    return tessel.simpleExec(commands.reconnectWifi());
-  };
+  var reconnectWifi = () => tessel.simpleExec(commands.reconnectWifi());
 
-  var reconnectDnsmasq = function() {
-    return tessel.simpleExec(commands.reconnectDnsmasq());
-  };
+  var reconnectDnsmasq = () => tessel.simpleExec(commands.reconnectDnsmasq());
 
-  var reconnectDhcp = function() {
-    return tessel.simpleExec(commands.reconnectDhcp());
-  };
+  var reconnectDhcp = () => tessel.simpleExec(commands.reconnectDhcp());
 
   return tessel.simpleExec(commands.commitWirelessCredentials())
     .then(reconnectWifi)
     .then(reconnectDnsmasq)
     .then(reconnectDhcp)
     .then((str) => logs.info(status, str))
-    .then(resolve);
+    .then(resolve)
+    .catch(reject);
 }
 
 Tessel.prototype.enableAccessPoint = function() {
   var status = 'Access Point successfully enabled.';
 
-  return new Promise((resolve) => {
+  return new Promise((resolve, reject) => {
     return this.simpleExec(commands.turnAccessPointOn())
-      .then(() => {
-        return commitAndClose(this, status, resolve);
-      });
+      .then(() => commitAndClose(this, status, resolve, reject));
   });
 };
 
 Tessel.prototype.disableAccessPoint = function() {
   var status = 'Access Point successfully disabled.';
 
-  return new Promise((resolve) => {
+  return new Promise((resolve, reject) => {
     return this.simpleExec(commands.turnAccessPointOff())
-      .then(() => {
-        return commitAndClose(this, status, resolve);
-      });
+      .then(() => commitAndClose(this, status, resolve, reject));
+  });
+};
+
+Tessel.prototype.getAccessPointInfo = function() {
+  var info;
+
+  return new Promise((resolve, reject) => {
+    return this.simpleExec(commands.getAccessPointConfig())
+      .then((config) => {
+        const props = ['ssid', 'key', 'encryption', 'disabled'];
+        const values = props.map((prop) => {
+          const regex = new RegExp(`${prop}='(.+)'`);
+          const value = regex.exec(config);
+
+          return value ? value[1] : null;
+        });
+
+        info = {
+          ssid: values[0],
+          key: values[1],
+          encryption: values[2],
+          disabled: values[3]
+        };
+
+        return this.simpleExec(commands.getAccessPointIP());
+      })
+      .then((ip) => {
+        info.ip = ip.replace('\n', '').trim();
+
+        resolve(info);
+      })
+      .catch(reject);
   });
 };
 
@@ -58,7 +81,7 @@ Tessel.prototype.createAccessPoint = function(opts) {
   var ssid = opts.ssid;
   var password = opts.pass;
   var security = opts.security;
-  var status = 'Created Access Point successfully.';
+  var status = 'Created Access Point successfully. ';
 
   var setupAccessPoint = () => {
     if (password && !security) {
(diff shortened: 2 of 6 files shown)


lib/controller.js
```diff
@@ -741,6 +741,35 @@ controller.disableAccessPoint = function(opts) {
...
+          }
+
+          logs.info('Security: ' + ap.security);
+          logs.info('IP Address: ' + ap.ip);
+          logs.info(!Number(ap.disabled) ? 'Enabled' : 'Disabled');
+        } else {
+          logs.info(tessel.name + ' is not configured as an access point (run "t2 ap --help" to learn more)');
+        }
+
+        return controller.closeTesselConnections([tessel]);
```
johnnyman727: @HipsterBrown You actually don't need to call `closeTesselConnections` because when you called `standardTesselCommand` on line 746 above, it wraps your callback such that it automatically closes Tessel once your Promise resolves.

johnnyman727: It does so here: https://github.com/HipsterBrown/t2-cli/blob/ap-refactor/lib/controller.js#L339-L348

lib/controller.js
```diff
@@ -741,6 +741,35 @@ controller.disableAccessPoint = function(opts) {
...
+      })
+      .catch((e) => {
+        logs.err(e);
+
+        return controller.closeTesselConnections([tessel]);
```
johnnyman727: Same comment as above. 

Also, if this command failed, it's best to reject the entire promise so that `tessel-2.js` will print out the proper ERR message.

johnnyman727:
If you rebase on master, the Travis errors should be fixed!

HipsterBrown:
Awesome, I'll make those fixes tonight.

johnnyman727:
Fantastic! The code looks good. Some smoke tests:
```
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap
INFO Connected to Tessel-02A3844F1188.
INFO Tessel-02A3844F1188 is not configured as an access point (run "t2 ap --help" to learn more)
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap --help

Usage: t2 ap [options]

Options:
   -t TIMEOUT, --timeout TIMEOUT     Set timeout in seconds for scanning for networked tessels  [5]
   -i PRIVATEKEY, --key PRIVATEKEY   SSH key for authorization with your Tessel
   --name NAME                       The name of the tessel on which the command will be executed
   --lan                             Use only a LAN connection
   --usb                             Use only a USB connection
   --lanPrefer                       Prefer a LAN connection when available, otherwise use USB.  [false]
   -o, --output                      Enable or disable writing command output to stdout/stderr. Useful for CLI API consumers.  [true]
   -p, --pass                        Password to access network.
   -s, --security                    Encryption to use on network (i.e. wep, psk, psk2, wpa, wpa2).
   --off                             Disable the access point
   --on                              Enable the access point

Configure the Tessel as an access point
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap --on
INFO Connected to Tessel-02A3844F1188.
INFO Access Point successfully enabled.
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap
INFO Connected to Tessel-02A3844F1188.
INFO Tessel-02A3844F1188 is not configured as an access point (run "t2 ap --help" to learn more)
```

So I think when we get a `--on` flag, we should actually check and see if an SSID and password have been set for the ap. In the above example, I ran `t2 ap --on` and it reported success but then `t2 ap` reported that it hasn't been configured to run as an access point. My `/etc/config/wireless` looks like:
```
 
config wifi-device 'radio0'
        option type 'mac80211'
        option channel '11'
        option hwmode '11g'
        option path '10180000.wmac'
        option htmode 'HT20'
        option disabled '0'
 
config wifi-iface
        option device 'radio0'
        option network 'wifi'
        option mode 'sta'
        option disabled '0'
        option ssid 'technicallyHome'
        option key 'Rusty007'
        option encryption 'psk2'
 
config wifi-iface
        option device 'radio0'
        option network 'lan'
        option mode 'ap'
        option encryption 'psk2'
        option disabled '0'
```
So you can see there is no SSID or password.

Continuing the smoke tests:
```
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap -n MyNetwork -p Tesssel11
INFO Connected to Tessel-02A3844F1188.
INFO Created Access Point successfully. SSID: MyNetwork, password Tesssel11, security mode: psk2
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap
INFO Connected to Tessel-02A3844F1188.
INFO SSID: MyNetwork
INFO Password: Tesssel11
INFO Security: psk2
INFO IP Address: 192.168.1.101
INFO Enabled
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap --off
INFO Connected to Tessel-02A3844F1188.
^C%   # This appears to be some other unrelated bug where it didn't run the process or something...                                                              
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap --off
INFO Connected to Tessel-02A3844F1188.
INFO Access Point successfully disabled.
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap
INFO Connected to Tessel-02A3844F1188.
INFO SSID: MyNetwork
INFO Password: Tesssel11
INFO Security: psk2
INFO IP Address: 192.168.1.101
INFO Disabled
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap --on
INFO Connected to Tessel-02A3844F1188.
INFO Access Point successfully enabled.
➜  t2-cli git:(HipsterBrown-ap-refactor) ✗ t2 ap
INFO Connected to Tessel-02A3844F1188.
INFO SSID: MyNetwork
INFO Password: Tesssel11
INFO Security: psk2
INFO IP Address: 192.168.1.101
INFO Enabled
```
Looks good! If I were really nitpicky, I might suggest you change "Enabled" to "State: Enabled" to be consistent with the rest of the logs.

So I think you could either fix the above issue regarding checking ssid before setting `--on` or you could do it in another PR if you want to move on.

HipsterBrown:
@johnnyman727 
Ah good catch on the `--on` predicament. I think I'll leave that for another immediate PR, just to break up the amount of changes in one build. Also, do you think I should return the access point info when it's enabled, so users don't have to go through an extra step to get that info?

I like the `State: Enabled/Disabled` idea. I wasn't sure how to phrase it before but that works perfectly. I'll push a fix for that asap and get this through.

johnnyman727:
> Also, do you think I should return the access point info when it's enabled, so users don't have to go through an extra step to get that info?

Yeah, I think that's actually a great suggestion. 

Also, I did some more testing to check if the weird hang I experienced would happen again and it turns out it happens every time I run `t2 ap --off` just after having run `t2 ap --on`. The reason is because the call to `tessel.simpleExec(commands.reconnectDhcp())` is actually returning an error that is reported by OpenWRT: `udhcpc: sendto: Network is down`. The `rejection` seems to be swallowed somewhere. I think we should fix this before merging to ensure a good user experience with the CLI.

HipsterBrown:
> The reason is because the call to tessel.simpleExec(commands.reconnectDhcp()) is actually returning an error that is reported by OpenWRT: udhcpc: sendto: Network is down. The rejection seems to be swallowed somewhere. I think we should fix this before merging to ensure a good user experience with the CLI.

Ok good catch. I'll work that into this build.

HipsterBrown:
@johnnyman727 I've updated the `t2 ap` output to match your suggestion. Also, errors are now rejected correctly and passed through to `tessel-2.js` to be logged.

lib/controller.js
```diff
@@ -741,6 +741,31 @@ controller.disableAccessPoint = function(opts) {
...
+
+          if (ap.password || ap.security !== 'none') {
+            logs.info('Password: ' + ap.password);
+          }
+
+          logs.info('Security: ' + ap.security);
+          logs.info('IP Address: ' + ap.ip);
+          logs.info('State: ' + (!Number(ap.disabled) ? 'Enabled' : 'Disabled'));
+        } else {
+          logs.info(tessel.name + ' is not configured as an access point (run "t2 ap --help" to learn more)');
```
rwaldron: Use template string

lib/tessel/access-point.js
```diff
@@ -46,14 +47,46 @@ Tessel.prototype.enableAccessPoint = function() {
...
 
+Tessel.prototype.getAccessPointInfo = function() {
+  var info;
+
+  return new Promise((resolve, reject) => {
+    return this.simpleExec(commands.getAccessPointConfig())
+      .then((config) => {
+        const props = ['ssid', 'key', 'encryption', 'disabled'];
+        const values = props.map((prop) => {
+          const regex = new RegExp(`${prop}='(.+)'`);
```
rwaldron: This pattern, but not this specific case, can potentially be a security hole. To be clear, the point im making is absolutely irrelevant to your changes here, I just wanted to use the opportunity as illustration.

lib/tessel/access-point.js
```diff
@@ -46,14 +47,46 @@ Tessel.prototype.enableAccessPoint = function() {
...
   });
 };
 
+Tessel.prototype.getAccessPointInfo = function() {
+  var info;
+
+  return new Promise((resolve, reject) => {
+    return this.simpleExec(commands.getAccessPointConfig())
+      .then((config) => {
+        const props = ['ssid', 'key', 'encryption', 'disabled'];
```
rwaldron: What is the benefit of using `const` here?

LinusU: I think that `const` should be preferred whenever you have a variable that won't change.

lib/tessel/access-point.js
```diff
@@ -29,16 +29,17 @@ function commitAndClose(tessel, status, resolve) {
...
 Tessel.prototype.enableAccessPoint = function() {
   var status = 'Access Point successfully enabled.';
 
-  return new Promise((resolve) => {
+  return new Promise((resolve, reject) => {
     return this.simpleExec(commands.turnAccessPointOn())
       .then(() => {
-        return commitAndClose(this, status, resolve);
+        return commitAndClose(this, status, resolve, reject);
       });
```
rwaldron: This could be cleaned up to:

```js   
.then(() => commitAndClose(this, status, resolve, reject));
```

lib/tessel/access-point.js
```diff
@@ -46,14 +47,46 @@ Tessel.prototype.enableAccessPoint = function() {
 Tessel.prototype.disableAccessPoint = function() {
   var status = 'Access Point successfully disabled.';
 
-  return new Promise((resolve) => {
+  return new Promise((resolve, reject) => {
     return this.simpleExec(commands.turnAccessPointOff())
       .then(() => {
-        return commitAndClose(this, status, resolve);
+        return commitAndClose(this, status, resolve, reject);
```
rwaldron: Same as above, this could be cleaned up to:

```js   
.then(() => commitAndClose(this, status, resolve, reject));
```

lib/tessel/access-point.js
```diff
@@ -46,14 +47,46 @@ Tessel.prototype.enableAccessPoint = function() {
...
+
+          return value ? value[1] : null;
+        });
+
+        info = {
+          ssid: values[0],
+          password: values[1],
+          security: values[2],
+          disabled: values[3]
+        };
```
rwaldron: What happens when `config` is not what we expected?

lib/tessel/wifi.js
```diff
@@ -48,7 +48,7 @@ Tessel.prototype.getWifiInfo = function() {
         var network = JSON.parse(resultsJSON);
 
         if (network.ssid === undefined) {
-          var msg = this.name + ' is not connected to Wi-Fi (run "tessel wifi -l" to see available networks)';
+          var msg = this.name + ' is not connected to Wi-Fi (run "t2 wifi -l" to see available networks)';
```
rwaldron: Make this a template string

lib/tessel/access-point.js
```diff
@@ -46,14 +47,46 @@ Tessel.prototype.enableAccessPoint = function() {
...
+
+        return this.simpleExec(commands.getAccessPointIP());
+      })
+      .then((ip) => {
+        info.ip = ip.replace('\n', '');
```
rwaldron: `ip.trim()` to ensure all extra white space is discarded.

lib/tessel/access-point.js
```diff
@@ -46,14 +47,46 @@ Tessel.prototype.enableAccessPoint = function() {
...
+          const value = regex.exec(config);
+
+          return value ? value[1] : null;
+        });
+
+        info = {
+          ssid: values[0],
+          password: values[1],
+          security: values[2],
+          disabled: values[3]
```
rwaldron: Name these as they are named in the props array

test/unit/access-point.js
```diff
@@ -294,3 +294,121 @@ exports['Tessel.prototype.disableAccessPoint'] = {
...
+
+    // Test is expecting two closes...;
+    this.tessel._rps.on('control', (command) => {
+      if (command.toString() === 'uci show wireless.@wifi-iface[1]') {
+        var info = new Buffer(
+          `wireless.cfg053579.ssid='${results.ssid}'
+          wireless.cfg053579.key='${results.password}'
+          wireless.cfg053579.encryption='${results.security}'
+          wireless.cfg053579.disabled='0'`
+        );
```
rwaldron: Just a heads up, this will preserve all the indentation white space:

```js   
var buffer = new Buffer(
  `a
   b
   c`
);

console.log(buffer);
console.log(buffer.length);
``` 

Result: 

```
a
   b
   c
11
``` 

Take a look in deploy.js, where I use the `tags.stripIndent` template tag from the "common-tags" module:

- https://github.com/tessel/t2-cli/blob/master/lib/tessel/deploy.js#L10
- https://github.com/tessel/t2-cli/blob/master/lib/tessel/deploy.js#L971-L974

lib/tessel/access-point.js
HipsterBrown: > What happens when config is not what we expected?

As line 69 shows, if the regex doesn't find the prop it expects from `config`, then `value` would be undefined and set to `null` for that prop. In `lib/controller.js` I check to see if `.ssid` is set before logging the info returned by `getAccessPointInfo`. Reference -> https://github.com/tessel/t2-cli/pull/596/files#diff-f68567477d803b49930337bf7fe1556bR749

lib/tessel/access-point.js
HipsterBrown: Thanks for that heads up. I will keep that in mind when using them. :+1:

lib/tessel/access-point.js
HipsterBrown: > I think that const should be preferred whenever you have a variable that won't change.

Agreed. If I'm not reassigning the variable at any point, (when I remember) I usually default to `const`.

johnnyman727:
:clap:

PR #598, opened 2016-02-23
Removes calls to process.close

johnnyman727:
The `process.close` can apparently be troublesome. I haven't investigated it fully, but it causes the next call to `.exec` to hang (and the test bench needs to make another call to ensure it can ping a network). We can just remove this function from use and when we need to close processes, just do it with `process.stdin.end('')` (or they will close naturally).

My guess is that in this case, the process was closing naturally and our request to close the process is confusing the usb daemon somehow. Fixing the usb daemon is definitely a better long term fix but I don't have the time to do that before this test rig is needed.

### DIFF
--- a/lib/tessel/wifi.js
+++ b/lib/tessel/wifi.js
@@ -149,15 +149,21 @@ Tessel.prototype.setWiFiState = function(enable) {
           return reject(err);
         }
         if (enable) {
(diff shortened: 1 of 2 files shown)


Frijol:
looks reasonable to me

PR #601, opened 2016-02-24
Use the proper stop method to use the scanner methods multiple times

johnnyman727:
Previously, it was impossible to call the `scanner.start` method multiple times because the scanner was never `stop`ped properly. All the Tessels that were originally found would still be considered found so no new Tessels would be returned. I also changed the default type of the scanners to `null` instead of `undefined` because that seemed more idiomatic.


Frijol:
looks good!

PR #605, opened 2016-02-25
[WIP] Adds progress bar animations for deploying code

johnnyman727:
This PR improves the UI of deploying code by adding a progress bar so you have some feedback of how long the process will take (and to ensure nothing is hanging). You can test it out with:
```
t2 init
t2 run index.js
```
You'll see a fancy progress bar quickly go from 0% to 100%.

The only thing I'm not sure about is how to monitor LAN `connection.write` progress well. As is, it just hits 100% immediately because the through stream I'm using just gets passed the entire bundle from `deploy.js` (`remoteProcess.stdin.end(bundle)`). I need some way to monitor progress in the stream's backpressure. @LinusU or @rwaldron, do you have any ideas of how I might be able to do that?

### DIFF
--- a/lib/lan/lan_progress_bar.js
+++ b/lib/lan/lan_progress_bar.js
@@ -1,0 +1,14 @@
+var ProgressBar = require('progress');
+var hookWritableStream = require('hook-writable-stream');
+
+// LAN Process specific wrapper around the `progress` module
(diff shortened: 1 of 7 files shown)


johnnyman727:
I'm closing this because I don't have a path forward for both types of connections (SSH and USB).

Issue #606, opened 2016-02-26
Wifi info timeout can be too short 


johnnyman727:
I ran into an issue with the test rig where the time between when we commit wireless credentials and when we check the result of the connection attempt is too short such that the process that prints connection info completes before the connection is taken up and it considers the connection a failure even though it succeeds a few seconds later.

Specifically, this code is [here](https://github.com/tessel/t2-cli/blob/12e4e212c9e540604f0ed27ffcf6bf3b86c2da78/lib/tessel/wifi.js#L145-L187). You can see that we call `reconnectWifi` which starts the process of connecting using credentials just saved in `/etc/config/wireless` then we set a timeout for 2 seconds to "give the connection time to settle". Then we create a process to print out the connection status which we scan for a keyword that indicates a successful connection. Unfortunately, this last process gets called before the connection has settled.

My guess is that a timeout is not the right way to go here if we want robust code. I'm surprised that we haven't been analyzing the output of `reconnectWifi` for a positive keyword.

johnnyman727:
For references, I ended up using an 8 second timeout in the test rig just to be safe. I had failures even at 4 seconds.

rwaldron:
Resolved https://github.com/tessel/t2-cli/pull/610

PR #610, opened 2016-03-02
Wifi: eliminate use of timeout in wifi connection verification

rwaldron:
Last night at nodebots.nyc, I was experiencing issues with the wifi connection commands reporting timeouts or producing `WARN Unable to verify connection. Please ensure you have entered the correct network credentials.`. After I landed #596, I took a closer look at `setWiFiState` and realized that if I eliminated the time out and replaced it with a recursively called "polling" mechanism, I could increase the likelihood of successfully verifying the wifi connection. I've arbitrarily selected 10 as the limit of tries that will be made; this could actually be increased to ~33, which is the approximate number of tries that could potentially be made in 12 seconds (the delay period (2) + timeout period (10)). I determined this by counting the number of complete tries made against a known invalid ssid, within a 12 second period. I ran the measurement code 10 times and the rounded average count was 33 tries.

### DIFF
--- a/lib/tessel/wifi.js
+++ b/lib/tessel/wifi.js
@@ -92,17 +92,18 @@ Tessel.prototype.connectToNetwork = function(opts) {
   var ssid = opts.ssid;
   var password = opts.password;
   var security = opts.security;
-  var status = 'Wifi connection successful.';
+  var status = 'Wifi Connected.';
 
   if (password && !security) {
     security = 'psk2';
   }
 
+  status += ` SSID: ${ssid}`;
+
   if (password && security) {
-    status += ' SSID: ' + ssid + ', password ' + password + ', security mode: ' + security;
+    status += `, password: ${password}, security: ${security}`;
   } else if (!password && (!security || security === 'none')) {
     security = 'none';
-    status += ' SSID: ' + ssid;
   }
 
   var setSSID = () => this.simpleExec(commands.setNetworkSSID(ssid));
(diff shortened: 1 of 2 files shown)


lib/tessel/wifi.js
```diff
@@ -137,61 +138,49 @@ Tessel.prototype.resetMDNS = function() {
...
+      .then(() => {
+        var settle = (rejection) => {
+          if (rejection) {
+            reject(rejection);
+          } else {
+            logs.info('Wifi', enable ? 'Enabled.' : 'Disabled.');
+            resolve();
+          }
+        };
+        var tries = 10;
```
rwaldron: See notes in PR message

test/unit/wifi.js
```diff
@@ -197,15 +197,13 @@ module.exports['Tessel.prototype.connectToNetwork'] = {
         // Write to stdout so it completes as expected
         // Wrap in setImmediate to make sure listener is set up before emitting
         setImmediate(() => {
-          this.tessel._rps.stdout.emit('data', 'signal');
-          setImmediate(() => this.tessel._rps.emit('close'));
+          this.tessel._rps.stdout.emit('data', new Buffer('signal'));
```
rwaldron: The string args will be changed to Buffers throughout, as `tessel.receive(...)` produces a Buffer.

HipsterBrown:
This is a similar solution to a similar wifi issue that came up while creating the Wifi API -> [Reference](https://github.com/tessel/t2-firmware/blob/master/node/tessel-export.js#L1229)

Looks good to me. I'll try a smoke test when I get home tonight.

johnnyman727:
@rwaldron I'm so glad that you worked on this. I was actually considering tackling it today because I was hitting this issue with the test rig (and it's why I created #606). You implemented the fix exactly how I was planning to do it as well. 

I ran it on my Tessel with correct and incorrect credentials and, with the exception of [this bug](https://github.com/tessel/t2-cli/issues/611) that's already on master, it worked exactly as expected. :+1:

PR #612, opened 2016-03-02
Ensure IDs are sorted prior to selecting next

johnnyman727:
When Tessel needs to create a new process on OpenWRT via a USB Connection (eg. `node /tmp/remote-script/index.js` to run a script), the CLI communicates with the USB Daemon running on OpenWRT to spawn a new process. As part of the protocol with the USB Daemon, the CLI must specify an integer to be used as the identifier for that process. That ensures the CLI and USB Daemon can route `stdin`, `stdout`, and `stderr` properly for multiple processes.

The `_nextID` function is responsible for selecting that next identifier for a new process (between 0-254) created over a USB Connection (LAN is managed by SSH). It does so by analyzing the processes identifiers already in use and selecting the next available number. For example, if the process IDs in use were `[0, 1, 3, 4, 6]`, the next available ID would be `2`.

The problem was that this function didn't take into account that the IDs-in-use list might be out of order so you could get something like `[1, 0]`. In this case, the function would see that the first element is not zero and return zero to be the next available ID, even though it was already in use.

This PR solves the problem by sorting the list of IDs-in-use before searching for the next available.

- [x] Add tests.

### DIFF
--- a/lib/usb/usb_daemon.js
+++ b/lib/usb/usb_daemon.js
@@ -225,7 +225,7 @@ function USBDaemon() {
   this._nextID = () => {
 
     // Get all currently used ids, what's the first available?
-    var usedIDs = this.getIDsInUse();
+    var usedIDs = this.getIDsInUse().sort();
     var prev;
     var cur;
 
--- a/test/unit/daemon.js
+++ b/test/unit/daemon.js
@@ -1,0 +1,62 @@
+module.exports['Daemon._nextID'] = {
+  setUp: function(done) {
+    this.sandbox = sinon.sandbox.create();
+    done();
+  },
+
+  tearDown: function(done) {
+    this.sandbox.restore();
+    done();
+  },
+
+  // Returns the next element in the list when ordered with no gaps
+  selectSubsequentId: function(test) {
+    test.expect(1);
+    this.getIDsInUse = this.sandbox.stub(Daemon, 'getIDsInUse').returns([0, 1, 2]);
+    var next = Daemon._nextID();
+    test.equal(next, 3);
+    test.done();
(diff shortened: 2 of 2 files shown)


HipsterBrown:
@johnnyman727 What's the best way to smoke test this?

johnnyman727:
Good question. There is no really easy way to smoke test. Perhaps adding a custom `controller` command that opens a bunch of processes in parallel over USB and checking that they all complete successfully? This code path is hit often but we rarely have multiple processes running in normal operation of the CLI.

HipsterBrown:
Would locally pulling this commit onto #595 and trying to update over USB be a good test?

johnnyman727:
@HipsterBrown the only instance where I've hit this problem is actually with the test rig. I think it' s not really worth the effort to build up a smoke test. Perhaps the simplest thing would be to see if you can still deploy code using this branch (or connect to wifi, etc)

HipsterBrown:
@johnnyman727 Alright. I'll give that a shot.

HipsterBrown:
Smoke Test:

**Setup**:
`mkdir tessel-test`
`cd tessel-test`
`t2 init`
`t2 run index.js --usb`

```bash
nickhehr@Nicks-MacBook-Pro-2 tessel-test $ t2 run index.js --usb
INFO Building project.
INFO Writing project to RAM on tessel-router (3.072 kB)...
INFO Deployed.
INFO Running index.js...
^Cnickhehr@Nicks-MacBook-Pro-2 tessel-test $
```

Success! :+1:

johnnyman727:
Thanks for the hand, @HipsterBrown!

PR #614, opened 2016-03-03
Check access point config before enabling through `ap --on`

HipsterBrown:
If someone tries to enable a Tessel as an access point before configuring it, it would confirm enabling without actually working. This PR checks the access point info before enabling and returns that info after enabling.


johnnyman727:
Works great for me! Merge whenever you're ready.

Issue #617, opened 2016-03-06
twitter example fails to run with slim build


johnnyman727:
Test Case:
```
mkdir twit-demo; cd twit-demo;
t2 init;
npm i twitter;
echo "require('twitter')" > index.js
t2 run index.js
```

Expected Results (which you can see by running `node index.js` or `t2 run index.js --full`):
```
INFO Looking for your Tessel...
INFO Connected to .
INFO Building project.
INFO Writing project to RAM on  (3816.448 kB)...
INFO Deployed.
INFO Running index.js...
# No output
```

Actual Results:
```
➜  twit-demo  t2 run index.js
INFO Looking for your Tessel...
INFO Connected to horse.
INFO Building project.
ERR! TypeError: self.key.print is not a function

**Comment by rwaldron** [comment]
I'll take a look at this immediately.

**Comment by johnnyman727** [comment]
I should have tested this more thoroughly after your recent PR @rwaldron but it looks like Slim builds are still failing for some reason. When I run [the twitter example](http://tessel.github.io/t2-start/tweet.html) I get the following error (and it doesn't occur if I run with `--full`):
```
➜  twit-demo  t2 run index.js
INFO Looking for your Tessel...
INFO Connected to Tessel-02A3BBAD6C82.
INFO Building project.
INFO Writing project to RAM on Tessel-02A3BBAD6C82 (974.336 kB)...
INFO Deployed.
INFO Running index.js...
/tmp/remote-script/node_modules/extend/index.js:1
[...]

PR #621, opened 2016-03-09
Allow all modules to have custom compression settings. Fixes gh-617 (2)

rwaldron:


rwaldron:
This may not be necessary. I discovered that the issue is very specifically with function expression names, and there was a patch that fixed it with an option.

PR #629, opened 2016-03-28
Add a crash reporter to diagnose CLI crashes.

tikurahul:
All crashes get submitted by default. The user can turn off crash reporting if he wants to.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -15,6 +15,7 @@ updateNotifier({
 
 // Internal
 var controller = require('../lib/controller');
+var CrashReporter = require('../lib/crash_reporter');
 var init = require('../lib/init');
 var logs = require('../lib/logs');
 var drivers = require('./tessel-install-drivers');
@@ -76,6 +77,42 @@ parser.command('install-drivers')
       .then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
   })
   .help('Install drivers');
+
+parser.command('crash-reporter')
+  .callback(opts => {
+    // t2 crash-reporter --on
+    if (opts.on) {
+      return CrashReporter.on().then(() => {
+        // t2 crash-reporter --on --test
+        if (opts.test) {
+          CrashReporter.test().then(module.exports.closeSuccessfulCommand);
+        }
+      }).then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
+    } else if (opts.off) {
+      // t2 crash-reporter --on
+      return CrashReporter.off()
+        .then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
+    }
+
+    // t2 crash-reporter --test
+    if (opts.test) {
+      // not handling failures, as we want to trigger a crash
+      CrashReporter.test()
+        .then(module.exports.closeSuccessfulCommand);
+    }
+  })
+  .option('off', {
+    flag: true,
+    help: 'Disable the Crash Reporter.'
+  })
+  .option('on', {
+    flag: true,
+    help: 'Enable the Crash Reporter.'
+  })
+  .option('test', {
+    flag: true,
+    help: 'Test the Crash Reporter.'
+  });
 
 parser.command('provision')
   .callback(callControllerCallback('provisionTessel'))
--- a/lib/crash_reporter.js
+++ b/lib/crash_reporter.js
@@ -1,0 +1,110 @@
+// System Objects
+var os = require('os');
+
+// Third Party Dependencies
+var request = require('request');
+var tags = require('common-tags');
+
+// Internal
+var logs = require('./logs');
+var packageJson = require('../package.json');
+var Preferences = require('./preferences');
+
+// the value of the crash reporter preference
+// the value has to be one of 'on' or 'off'
+var CRASH_REPORTER_PREFERENCE = 'crash.reporter.preference';
+var SUBMIT_CRASH_URL = 'http://crash-reporter.tessel.io/crashes/submit';
+
+// override for testing
+if (process.env.DEV_MODE === 'true') {
+  SUBMIT_CRASH_URL = 'http://localhost:8080/crashes/submit';
+}
+
+var CrashReporter = {};
+
+CrashReporter.off = function() {
+  return Preferences.write(CRASH_REPORTER_PREFERENCE, 'off')
+    .catch(error => {
+      // do nothing
+      // do not crash the crash reporter :)
+      logs.err('Error turning off crash reporter preferences', error);
+    });
+};
+
+CrashReporter.on = function() {
+  return Preferences.write(CRASH_REPORTER_PREFERENCE, 'on')
+    .catch(error => {
+      // do nothing
+      // do not crash the crash reporter :)
+      logs.err('Error turning on crash reporter preferences', error);
+    });
+};
+
+CrashReporter.submit = function(report) {
+  return Preferences.read(CRASH_REPORTER_PREFERENCE, 'on')
+    .then(value => {
+      if (value === 'on') {
+        var labels = tags.stripIndent `
+          ${packageJson.name},
+          CLI version: ${packageJson.version},
+          Node version: ${process.version},
+          OS platform: ${os.platform()},
+          OS release: ${os.release()}
+        `;
+
+        CrashReporter.post(labels, report)
+          .then(() => {
+            logs.info('Done !');
+          })
+          .catch(error => {
+            logs.err('Error submitting crash report', error);
+          });
+      }
+    })
+    .catch(error => {
+      // do nothing
+      // do not crash the crash reporter :)
+      logs.err('Error submitting crash report', error);
+    });
+};
+
+CrashReporter.post = function(labels, report) {
+  return new Promise((resolve, reject) => {
+    request.post({
+      url: SUBMIT_CRASH_URL,
+      form: {
+        crash: report,
+        labels: labels,
+        f: 'json'
+      }
+    }, (error, httpResponse, body) => {
+      try {
+        if (error) {
+          reject(error);
+        } else {
+          var json = JSON.parse(body);
+          if (json.error) {
+            reject(json.error);
+          } else {
+            resolve();
+          }
+        }
+      } catch (exception) {
+        reject(exception);
+      }
+    });
+  });
+};
+
+CrashReporter.test = () => {
+  return Promise.reject(new Error('Testing the crash reporter'));
+};
+
+var onError = error => {
+  return CrashReporter.submit(error.stack);
+};
+
+process.on('unhandledRejection', onError);
+process.on('uncaughtException', onError);
+
+module.exports = CrashReporter;
--- a/lib/preferences.js
+++ b/lib/preferences.js
@@ -1,0 +1,70 @@
+// System Objects
+var path = require('path');
+
+// Third Party Dependencies
+var fs = require('fs-extra');
+
+// Internal
+var logs = require('./logs');
+
+
+var home = (process.platform.startsWith('win')) ? process.env.HOMEPATH : process.env.HOME;
+var preferencesJson = path.join(home, '.tessel', 'preferences.json');
+var Preferences = {};
+
+Preferences.read = function(key, defaultValue) {
+  return Preferences.load().then(contents => {
+      if (contents) {
+        return contents[key] || defaultValue;
+      } else {
+        return defaultValue;
+      }
+    })
+    .catch(error => {
+      logs.err('Error reading preference', key, error);
+      return defaultValue;
+    });
+};
+
+Preferences.write = function(key, value) {
+  return new Promise((resolve, reject) => {
+    Preferences.load()
+      .then(contents => {
+        contents = contents || {};
+        contents[key] = value;
+        fs.writeFile(preferencesJson, JSON.stringify(contents), function(error) {
+          if (error) {
+            logs.err('Error writing preference', key, value);
+            reject(error);
+          } else {
+            resolve();
+          }
+        });
+      })
+      .catch(error => {
+        reject(error);
+      });
+  });
+};
+
+Preferences.load = function() {
+  return new Promise((resolve, reject) => {
+    fs.exists(preferencesJson, (exists) => {
+      if (exists) {
+        fs.readFile(preferencesJson, (error, data) => {
+          if (error) {
+            reject(error);
+          } else {
+            resolve(JSON.parse(data));
+          }
+        });
+      } else {
+        // we don't have any local preferences
+        // return falsy value
+        resolve();
+      }
+    });
+  });
+};
+
+module.exports = Preferences;
--- a/test/unit/crash-reporter.js
+++ b/test/unit/crash-reporter.js
@@ -1,0 +1,181 @@
+/*global CrashReporter */
+/*global Preferences */
+
+exports['CrashReporter'] = {
+  surface: function(test) {
+    test.expect(5);
+    test.equal(typeof CrashReporter.on, 'function');
+    test.equal(typeof CrashReporter.off, 'function');
+    test.equal(typeof CrashReporter.post, 'function');
+    test.equal(typeof CrashReporter.submit, 'function');
+    test.equal(typeof CrashReporter.test, 'function');
+    test.done();
+  }
+};
+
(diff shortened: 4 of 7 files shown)


rwaldron:
Sorry if this is a dumb question, but what crashes is this reporting?

tikurahul:
@rwaldron It's not a dumb question at all. :smile: 
Anytime, any of the CLI commands crash unexpectedly for any reason, this module captures those errors and reports it to the [service](http://tessel-error-reporter.appspot.com/).

This will help us understand proactively fix errors, rather than having users submit issues in Github manually. I wrote this, as I personally ran into this problem that @johnnyman727 helped me fix.

bin/tessel-2.js
```diff
@@ -70,6 +71,28 @@ parser.command('install-drivers')
...
+  .help('Turn off crash reporting completely');
+
+parser.command('crash-reporting-on')
+  .callback(function() {
+    crashReporter.turnOn()
+      .then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
+  })
+  .help('Turn on crash reporting');
+
+parser.command('crash-reporting-check')
```
johnnyman727: Can we take this out before merging? (along with `crashReporter.testSubmit`)

lib/crash_reporter.js
```diff
@@ -0,0 +1,191 @@
...
+var path = require('path');
+var request = require('request');
+
+// the value of the crash reporter preference
+// the value has to one of 'on' or 'off'
+var CRASH_REPORTER_PREFERENCE = 'crash.reporter.preference';
+var SUBMIT_CRASH_URL = 'http://tessel-error-reporter.appspot.com/crashes/submit';
+
+// override for testing
+// SUBMIT_CRASH_URL = 'http://localhost:8080/crashes/submit';
```
johnnyman727: Remove before merging

lib/crash_reporter.js
```diff
@@ -0,0 +1,191 @@
+var fs = require('fs');
+var logs = require('./logs');
+var packageJson = require('../package.json');
+var path = require('path');
+var request = require('request');
+
+// the value of the crash reporter preference
+// the value has to one of 'on' or 'off'
```
johnnyman727: Typo in this sentence

lib/crash_reporter.js
```diff
@@ -0,0 +1,191 @@
...
+// override for testing
+// SUBMIT_CRASH_URL = 'http://localhost:8080/crashes/submit';
+
+var preferenceFile = path.join(__dirname, '__preferences.json');
+
+var Preferences = function() {
+
+};
+
+Preferences.prototype.read = function(key, defaultValue) {
```
johnnyman727: Structurally, it seems like the `Preferences` class should be in its own file since it's not specific to crash reporting.

lib/crash_reporter.js
```diff
@@ -0,0 +1,191 @@
...
+      logs.err('Error turning on crash reporter preferences', error);
+    });
+};
+
+CrashReporter.prototype.submit = function(report) {
+  var that = this;
+  that.preferences.read(CRASH_REPORTER_PREFERENCE, 'on')
+    .then(function(value) {
+      if (value === 'on') {
+        var labels = [packageJson.name, packageJson.version].join(',');
```
johnnyman727: Other info that might be useful for a report:
* Operating System
* Node version
* npm version
* tessel version (although this might be harder to get from this point in the code)

lib/crash_reporter.js
```diff
@@ -0,0 +1,191 @@
...
+          } else {
+            resolve();
+          }
+        }
+      } catch (exception) {
+        reject(exception);
+      }
+    });
+  });
+  return promise;
```
johnnyman727: In the rest of the code, we usually follow the convention of `return new Promise(...` rather than creating a local `promise` variable and immediately returning it.

johnnyman727:
@tikurahul I've reviewed the code and overall it looks great.

My biggest concern is the `Preferences` implementation because @kevinmehall has warned me repeatedly of the potential dangers of having the CLI, which can be used programmatically, manipulating a file across calls. You can hit a race condition where multiple programmatically called commands try to read and write from the file at the same time and you have a sorry couple of debugging days ahead of you. Do you think we should be concerned about that? Are there better ways to implement it?

Additionally, I think we should be sending more info back with the report as I noted in the comment in the code.

I'm super excited for this feature!

rwaldron:
> Anytime, any of the CLI commands crash unexpectedly for any reason, this module captures those errors and reports it to the service.


ohhhh this is compelling

lib/crash_reporter.js
```diff
@@ -0,0 +1,191 @@
...
+            resolve();
+          }
+        });
+      })
+      .catch(function(error) {
+        reject(error);
+      });
+  });
+  return promise;
+};
```
rwaldron: Does `__preferences.json` ever change during the life of a single command? I'm not sure I understand why there's a whole class defined for this...?

bin/tessel-2.js
tikurahul: Yes, it was only meant for tests. I am not sure where integration tests go. If you have any suggestions, please let me know.

lib/crash_reporter.js
tikurahul: `__preferences.json` is just a way for us to serialize persistent state. So once the user decides that he no longer want to submit crash reports at all, it stays that way.

lib/crash_reporter.js
tikurahul: Moved `Preferences` to its own file.

lib/crash_reporter.js
tikurahul: Captured Operating System, and version & Node version.
Looks like I am having to resort to unpleasant hacks to determine `npm` and `tessel` versions. :disappointed: 

I could execute a shell command `npm ls`, but am not particularly fond of that approach. If you have other suggestions, please let me know.

lib/crash_reporter.js
tikurahul: Done

tikurahul:
To turn on `DEV_MODE`, we can use something like `export DEV_MODE='true'`, which I think is better than removing this completely. Helps run integration tests easily.

lib/preferences.js
```diff
@@ -0,0 +1,71 @@
...
+        // we don't have any local preferences
+        // return falsy value
+        resolve();
+      }
+    });
+  });
+  return promise;
+};
+
+module.exports.Preferences = Preferences;
```
rwaldron: Why not `module.exports = Preferences`?

.gitignore
```diff
@@ -2,3 +2,4 @@ config.env
 node_modules/
 !test/unit/fixtures/**/node_modules
 tmp
+**/__preferences.json
```
rwaldron: This file should be created in HOME/.tessel/ where the rest of the user-owned stuff is

bin/tessel-2.js
```diff
@@ -70,6 +71,30 @@ parser.command('install-drivers')
...
   .help('Install drivers');
 
+parser.command('crash-reporting-off')
+  .callback(function() {
+    crashReporter.turnOff()
+      .then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
+  })
+  .help('Turn off crash reporting completely');
+
+parser.command('crash-reporting-on')
```
rwaldron: This should be `t2 crash-reporter` with on and off flags, like all the others. Here's wifi:

```Js   
  .option('off', {
    flag: true,
    help: 'Disable the wireless network'
  })
  .option('on', {
    flag: true,
    help: 'Enable the wireless network'
  })
```

bin/tessel-2.js
```diff
@@ -70,6 +71,30 @@ parser.command('install-drivers')
...
+  .callback(function() {
+    crashReporter.turnOn()
+      .then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
+  })
+  .help('Turn on crash reporting');
+
+if (process.env.DEV_MODE === 'true') {
+  parser.command('crash-reporting-check')
```
rwaldron: check should be an option

lib/crash_reporter.js
```diff
@@ -0,0 +1,136 @@
...
+          // tessel cli labels
+          packageJson.name,
+          `CLI version: ${packageJson.version}`,
+          // node version
+          `Node version: ${process.version}`,
+          // os version
+          `OS platform: ${os.platform()}`,
+          `OS release: ${os.release()}`
+        ];
+
```
rwaldron: This can just be one multi line template string. Use `tags.stripIndent` (add "common-tags" to requires in this file)

lib/crash_reporter.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+CrashReporter.prototype.submitReport = function(labels, report) {
+  return new Promise(function(resolve, reject) {
+    request.post({
+      'url': SUBMIT_CRASH_URL,
+      'form': {
+        'crash': report,
+        'labels': labels,
+        'f': 'json'
```
rwaldron: Don't quote keys

lib/crash_reporter.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+    }, function(error, httpResponse, body) {
+      try {
+        if (error) {
+          reject(error);
+        } else {
+          var json = JSON.parse(body);
+          if (json['error']) {
+            reject(json['error']);
```
rwaldron: Use `json.error`

lib/crash_reporter.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+    })
+    .catch(function(error) {
+      // do nothing
+      // do not crash the crash reporter :)
+      logs.err('Error submitting crash report', error);
+    });
+};
+
+CrashReporter.prototype.submitReport = function(labels, report) {
```
rwaldron: Since we know it's a "report", just call this "submit"

lib/crash_reporter.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+CrashReportsHelper.testSubmit = function() {
+  return Promise.reject(new Error('Testing the crash reporter'));
+};
+
+process.on('unhandledRejection', onError);
+process.on('uncaughtException', onError);
+
+module.exports = CrashReportsHelper;
```
rwaldron: Why is `CrashReporter` a class if it never gets exposed?

lib/crash_reporter.js
```diff
@@ -0,0 +1,136 @@
...
+if (process.env.DEV_MODE === 'true') {
+  SUBMIT_CRASH_URL = 'http://localhost:9080/crashes/submit';
+}
+
+// the only instance of the crash reporter
+var crashReporter = null;
+
+var CrashReporter = function() {
+  this.preferences = new preferences.Preferences();
+};
```
rwaldron: CrashReporter doesn't need to be a class, since only one instance is ever created and used. Just make a plan object "bag of methods" and export that.

rwaldron:
Any function expressions that can be arrow functions should be changed.

lib/preferences.js
tikurahul: Done. The reason why i did this originally was to avoid mix-casing on requires. 

```javascript
var path = require('path');
var Preferences = require('./preferences');
```
I thought this was looking weird and inconsistent.

.gitignore
tikurahul: Great idea. Moving this to the `$HOME` folder.

lib/crash_reporter.js
tikurahul: Done

lib/crash_reporter.js
tikurahul: Make CrashReporter a regular `Object` with properties.

lib/crash_reporter.js
tikurahul: This one is actually making the HTTP POST request. Renamed the method to something more sensible.

lib/crash_reporter.js
tikurahul: Done

lib/crash_reporter.js
tikurahul: Done

lib/crash_reporter.js
tikurahul: Done

bin/tessel-2.js
tikurahul: Done

bin/tessel-2.js
tikurahul: Done

.gitignore
tikurahul: Done

lib/crash_reporter.js
```diff
@@ -0,0 +1,120 @@
...
+  preferences.read(CRASH_REPORTER_PREFERENCE, 'on')
+    .then(value => {
+      if (value === 'on') {
+        var labels = tags.stripIndent `
+          ${packageJson.name},
+          CLI version: ${packageJson.version},
+          Node version: ${process.version},
+          OS platform: ${os.platform()},
+          OS release: ${os.release()}
+        `;
```
rwaldron: Beautiful!

lib/crash_reporter.js
rwaldron: I think you can get rid of `CrashReportsHelper` entirely now. Just move `testSubmit` to `CrashReporter` and then change this line to `module.export = CrashReporter;`

lib/preferences.js
```diff
@@ -0,0 +1,70 @@
...
+          }
+        });
+      })
+      .catch(error => {
+        reject(error);
+      });
+  });
+};
+
+Preferences.prototype.readPreferences = function() {
```
rwaldron: nitpick: `readPreferences` => `load`


```and not `loadPreferences`, because we know this object is a "preferences" ;)

File: lib/crash_reporter.js (hunk 16 above)
Comment (tikurahul) [inline lib/crash_reporter.js]: You are right. Will fix.

File: test/unit/crash-reporter.js [hunk 20]
```diff
@@ -0,0 +1,181 @@
...
+
+    var report = 'Error: undefined is not a function';
+    CrashReporter.submit(report).then(() => {
+      var args = this.crPost.lastCall.args;
+      test.equal(this.crPost.callCount, 1);
+      test.equal(typeof args[0], 'string');
+      test.ok(args[0].indexOf('CLI version') > 0, 'Label CLI version should be present');
+      test.ok(args[0].indexOf('Node version') > 0, 'Label Node version should be present');
+      test.ok(args[0].indexOf('OS platform') > 0, 'Label OS platform should be present');
+      test.ok(args[0].indexOf('OS release') > 0, 'Label OS release should be present');
```
rwaldron: You can use `args[0].includes(...)` here if you'd like :)

test/unit/preferences.js
```diff
@@ -0,0 +1,125 @@
...
+    });
+  },
+
+  readSetValues: function(test) {
+    test.expect(1);
+
+    var defaultValue = null;
+    var value = 'value';
+    this.readFile = this.sandbox.stub(fs, 'readFile', (file, handler) => {
+      handler(null, `{"key": "${value}"}`);
```
rwaldron: 👍

test/unit/preferences.js
```diff
@@ -0,0 +1,125 @@
...
+  setUp: function(done) {
+    this.sandbox = sinon.sandbox.create();
+    this.state = {};
+    this.logsErr = this.sandbox.stub(logs, 'err');
+    this.logsInfo = this.sandbox.stub(logs, 'info');
+    this.exists = this.sandbox.stub(fs, 'exists', (file, handler) => {
+      handler(true);
+    });
+    this.readFile = this.sandbox.stub(fs, 'readFile', (file, handler) => {
+      handler(null, JSON.stringify(this.state));
```
rwaldron: This is smart—nice!

test/unit/crash-reporter.js
tikurahul: `args[0]` is actually `labels` and 
```javascript
typeof args[0] === 'string'
```
So, I cant use `args[0].includes(...)`.

---

Reference is in the source for [submit crash handler](https://github.com/tessel/t2-crash-reporter/blob/master/main.py#L73)

test/unit/preferences.js
tikurahul: I learn from the best :beers:

Issue #632, opened 2016-04-02
t2 update: remote updatePath should not be created with path.join(...)


rwaldron:
This:

```js
var updatePath = path.join('/tmp/', updates.OPENWRT_BINARY_FILE);
```

Is fine on *nix platforms, because the remote path will have the same structure, ie. forward slash delimiters. When updating from a windows machine, this will result in a windows normalized path (I will post an example directly from my windows machine as soon as I post this)

rwaldron:
```
Rick@Rick-PC MINGW64 ~
$ node
> path.join("/tmp/", "other");
'\\tmp\\other'
```

PR #633, opened 2016-04-02
Do not local path.join(...) for remote system paths. Fixes gh-632

rwaldron:
Fairly critical, as this will definitely be breaking `t2 update` on windows. I've tested on both OSX and Windows, so once this goes green I'm making an executive decision to land and release today.


LinusU:
LGTM :+1:

PR #637, opened 2016-04-07
warn about ignore folder

Student007:
Some people doing their first project on Tessel could be confused if
they didn’t read there has to be a `.tesselinclude` to deploy folders.
This PR adds the following warning if not `.tesselinclude` exists (for
each folder found)
```
INFO Connected to Third.
WARN Ignoring ./public (.tesselinclude does not exist)
INFO Building project.
INFO Writing project to RAM on Third (1508.864 kB)...
INFO Deployed.
INFO Running index.js...
```

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -731,6 +731,21 @@ actions.tarBundle = function(opts) {
 
   var includeRules = actions.glob.rules(target, '.tesselinclude');
 
+  // assuming a programmer could miss to add .tesselinclude
+  // if there are project folders
+  try {
+    fs.statSync('.tesselinclude');
+  } catch (e) {
+    var dirList = fs.readdirSync('./');
+    dirList.forEach(function(element) {
+      if (element !== 'node_modules' && fs.lstatSync(element).isDirectory()) {
+        // blame for directories because is seams the programmer
+        // forgot to create .tesselinclude and add it into it
+        logs.warn('Ignoring ./' + element + ' (.tesselinclude does not exist)');
+      }
+    });
+  }
+
   // Convert `deployLists.includes` into includeRules
   deployLists.includes.forEach(include => includeRules.push(`node_modules/**/${include}`));
 
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -734,16 +734,21 @@ actions.tarBundle = function(opts) {
   // assuming a programmer could miss to add .tesselinclude
   // if there are project folders
   try {
-    fs.statSync('.tesselinclude');
+    fs.statSync(globRoot+'/.tesselinclude');
   } catch (e) {
-    var dirList = fs.readdirSync('./');
+    var dirList = fs.readdirSync(globRoot);
+    var list = [];
     dirList.forEach(function(element) {
-      if (element !== 'node_modules' && fs.lstatSync(element).isDirectory()) {
+      if (element !== 'node_modules' && fs.lstatSync(globRoot+'/'+element).isDirectory()) {
         // blame for directories because is seams the programmer
         // forgot to create .tesselinclude and add it into it
-        logs.warn('Ignoring ./' + element + ' (.tesselinclude does not exist)');
+        list.push(element);
       }
     });
+    if (list.length > 0) {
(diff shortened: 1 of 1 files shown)


HipsterBrown:
This could certainly be quite helpful for folks getting started with Tessel. 

Could this also be potentially noisy? Wouldn't folders like `./node_modules` always have a warning, if a `.tesselinclude` does not exist?

I am in favor of this warning but want to recognize a potential negative effect as well.

johnnyman727:
I agree with you @Student007 that we don't have enough transparency/documentation about the `.tesselinclude` file but I don't think printing out more stuff each time you deploy is the right solution because, as @HipsterBrown pointed out, it would be really noisy.

Perhaps if we only print it once when there is any non-bundled folders/files in the top level directory?

Frijol:
Alternate proposal: we modify `t2 init` to create a `.tesselinclude` file that is empty except for a comment explaining how to use it/linking to docs on the subject.

That keeps CLI clean and gives people a clear pathway, as well as encouraging people to learn about it in the first run (we would mention it when we list what files `t2 init` makes).

johnnyman727:
I'm a favor of @Frijol's suggestion in conjunction with my previous suggestion. Any other thoughts?

Frijol:
yeah, I like @johnnyman727 's suggestion in conjunction

rwaldron:
Seconding @Frijol & @johnnyman727, here's a summary: 

1. Update `t2 init` to auto-create a `.tesselinclude` file, that contains the following: 
```
# This file should list any files or directories you want included with the 
# T2 bundling and deployment process. This is handy for including 
# non-JavaScript assets, like HTML, CSS, and images, for use within your project.
# .tesselinclude works the same as .npminclude 
# You DO NOT need to list node_modules or package.json
# For more information, visit: https://tessel.io/docs/cli#usage
```
1. Update `t2 run` and `t2 push` operations: 
  1. Read the **entry point file's directory** (which is `globRoot`, _not_ `'./'` \*) as `list` 
    1. If `list` does not include `.tesselinclude`, do the following:
      1. If `list.length` is 3 and `list` includes `node_modules`, `list` includes `[opts.resolvedEntryPoint]` and `list` includes `package.json`, **do nothing**
      1. Else, 
        1. If `list.length` is > 3,
          1. Let `assets` be an empty array
          1. Foreach `entry` in `list`, if `[opts.resolveEntryPoint, 'node_modules', 'package.json'].includes(entry) === false`, push `entry` into `assets`
          1. To the console, print: "The following items were found in your project:"
          2. `list.forEach(entry => console.log(\`${entry}\`)`
          3. To the console, print: "If you need to deploy them to your tessel, you must create a .tesselinclude file that contains the items you wish to deploy with your application." 


This spec should be sufficient to meet the requirements and only costs a single `readdirSync`. 






\* The current patch assumes that `t2 run ...` or `t2 push ...` is always called from the project directory, which is an invalid assumption.

Student007:
@rwaldron changed some things based on the discussion.

About your spec I think you should do this while fixing the behavior of 
```
Error: Cannot find module '/tmp/.../index.js'
```
when deploying out of a parent folder like `myholetesselprojects`...

This PR only is a quick fix about "is the .tesselinclude" present and if so, take a look whether the user has folders we don't expect and blame about this.
A much better solution is a thing of further development. At the moment this solution is a good hint at the right point for new Tessel-Owners .. as @HipsterBrown noted:
> This could certainly be quite helpful for folks getting started with Tessel. 

I think it also would be enough to warn about without folder names. But on the other hand a large list of folders only happens once. So the argument of much blaming is not valid. It has to be prominent.

rwaldron:
@Student007 Yep, I'm actually working on that right now.

lib/tessel/deploy.js
```diff
@@ -731,6 +731,26 @@ actions.tarBundle = function(opts) {
 
   var includeRules = actions.glob.rules(target, '.tesselinclude');
 
+  // assuming a programmer could miss to add .tesselinclude
+  // if there are project folders
+  try {
+    fs.statSync(globRoot+'/.tesselinclude');
+  } catch (e) {
```
rwaldron: `.tesselinclude` isn't required, so please remove this.

Issue #638, opened 2016-04-07
Implement complete tesselinclude detect spec


rwaldron:
1. Update `t2 init` to auto-create a `.tesselinclude` file, that contains the following: 
```
# This file should list any files or directories you want included with the 
# T2 bundling and deployment process. This is handy for including 
# non-JavaScript assets, like HTML, CSS, and images, for use within your project.
# .tesselinclude works the same as .npminclude 
# You DO NOT need to list node_modules or package.json
# For more information, visit: https://tessel.io/docs/cli#usage
```
1. Update `t2 run` and `t2 push` operations: 
  1. Read the **entry point file's directory** (which is `globRoot`, _not_ `'./'` \*) as `list` 
    1. If `list` does not include `.tesselinclude`, do the following:
      1. If `list.length` is 3 and `list` includes `node_modules`, `list` includes `[opts.resolvedEntryPoint]` and `list` includes `package.json`, **do nothing**
      1. Else, 
        1. If `list.length` is > 3,
          1. Let `assets` be an empty array
          1. Foreach `entry` in `list`, if `[opts.resolveEntryPoint, 'node_modules', 'package.json'].includes(entry) === false`, push `entry` into `assets`
          1. To the console, print: "The following items were found in your project:"
          2. `list.forEach(entry => console.log(\`${entry}\`)`
          3. To the console, print: "If you need to deploy them to your tessel, you must create a .tesselinclude file that contains the items you wish to deploy with your application." 

This spec should be sufficient to meet the requirements and only costs a single `readdirSync`. 

\* The current patch assumes that `t2 run ...` or `t2 push ...` is always called from the project directory, which is an invalid assumption.

PR #639, opened 2016-04-07
Implement complete tesselinclude detect spec. Closes gh-638

rwaldron:
@HipsterBrown review?

### DIFF
--- a/lib/init.js
+++ b/lib/init.js
@@ -1,6 +1,6 @@
 // System Objects
 var path = require('path');
-var fs = require('fs');
+var fs = require('fs-extra');
 
 // Third Party Dependencies
 var PZ = require('promzard').PromZard;
@@ -124,7 +124,23 @@ function generateSample() {
     fs.createReadStream(path.resolve(__dirname + './../resources/' + filename))
       .pipe(fs.createWriteStream(filename));
 
-    console.log('Wrote \'Hello World\' to index.');
+    console.log('Wrote "Hello World" to index.js');
+  });
+}
+
+function createTesselinclude() {
+  var tesselinclude = '.tesselinclude';
+
+  return new Promise((resolve) => {
+    fs.exists(tesselinclude, (exists) => {
+      if (exists) {
+        return resolve();
+      }
+
+      fs.copySync(path.resolve(__dirname + './../resources/' + tesselinclude), tesselinclude);
+      console.log('Created .tesselinclude.');
+      resolve();
+    });
   });
 }
 
@@ -160,7 +176,7 @@ module.exports = function(opts) {
   // Set interactive boolean off of CLI flags
   options = opts;
 
-  console.log('Initializing tessel repository...');
+  console.log('Initializing Tessel 2 Project...');
 
   readPackageJson()
     .then(function(data) {
@@ -197,6 +213,7 @@ module.exports = function(opts) {
     .then(JSON.parse)
     .then(getDependencies)
     .then(npmInstall)
+    .then(createTesselinclude)
     .then(generateSample)
     .catch(function(error) {
       console.error(error);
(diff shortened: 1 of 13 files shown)


HipsterBrown:
Awesome update! 👍 LGTM

lib/tessel/deploy.js
```diff
@@ -780,13 +760,32 @@ actions.tarBundle = function(opts) {
...
 
-  logs.info('Building project.');
+  if (assets.length) {
+    logs.warn(
+      tags.stripIndent `
+        The following items were found in your project directory, but not deployed.
+
+          ${assets.join('\n')}
+
+        If you need to deploy them to your Tessel 2, you must create a .tesselinclude
```
Frijol: should this be "create" a .tesselinclude or something else to reflect that they probably have one due to t2 init?

lib/tessel/deploy.js
```diff
@@ -780,13 +760,32 @@ actions.tarBundle = function(opts) {
...
+        file that lists the items you wish to deploy with your application.
+
+        You may also create a .tesselignore file to list items that should be ignored
+        by the deployment process.
```
Frijol: add a line that tells you where to go for more info?

Also, not sure whether the .tesselignore note is necessary here

lib/tessel/deploy.js
```diff
@@ -1004,7 +1004,7 @@ actions.writeToFile = function(t, entryPoint) {
           // When that process completes
           remoteProcess.once('close', function() {
             // Let the user know
-            logs.info('You may now disconnect from the Tessel. Your code will be run whenever Tessel boots up. To remove this code, use `tessel erase`.');
+            logs.info('You may now disconnect from the Tessel. Your code will be run whenever Tessel boots up. To remove this code, use `t2 erase`.');
```
Frijol: oof, good catch

lib/init.js
```diff
@@ -1,6 +1,6 @@
 // System Objects
 var path = require('path');
-var fs = require('fs');
+var fs = require('fs-extra');
```
Frijol: need to add to package.json?

Frijol:
Made a PR to the start page to reflect the changes: https://github.com/tessel/t2-start/pull/88

lib/tessel/deploy.js
rwaldron: If the project wasn't started with `t2 init`, they will have to create the file. We could also say "type `t2 init`", but what if they _did_ and purposely deleted the `.tesselinclude` file, thinking that they didn't need to use it? It might seem strange to say "run `t2 init`" to someone that already did that.

lib/tessel/deploy.js
rwaldron: > add a line that tells you where to go for more info?

Sure thing

> Also, not sure whether the .tesselignore note is necessary here
 
If I'm seeing this first time and not aware of these features, I think it's useful to be told about both... but the previous item will satisfy this, ie. including a link to the docs.

lib/init.js
rwaldron: Nope, it's in use throughout the codebase, this case was out of date

lib/tessel/deploy.js
```diff
@@ -780,13 +760,32 @@ actions.tarBundle = function(opts) {
...
+      tags.stripIndent `
+        The following items were found in your project directory, but not deployed.
+
+          ${assets.join('\n')}
+
+        If you need to deploy them to your Tessel 2, you must have a .tesselinclude
+        file that lists the file and directory assets for your application. This can be
+        created manually or by typing 't2 init'.
+
+        For more information, visit: https://tessel.io/docs/cli#starting-projects
```
rwaldron: @Frijol thanks for review—I think this is much better, but let me know!

lib/tessel/deploy.js
Frijol: Yeah, sounds fine to leave as is

Frijol:
looks good to me!

rwaldron:
@Frijol thanks for the review and follow up!

PR #640, opened 2016-04-07
uglify warn with filename

Student007:
Due to hard debugging problems I decided to publish this solution as a
PR. It creates WARN messages if the final uglify process lands in the
last catch and gives reasonable feedback like this:
```
WARN Unexpected token operator «*», expected punc «(» at line 55:18 in
file: /Users/Student007/tessel2/experiment1/index.js
```

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -803,6 +803,7 @@ actions.tarBundle = function(opts) {
           dependencies.forEach(dependency => {
             var isJS = dependency.file.endsWith('.js');
             var source = dependency.source;
+            var file = dependency.file;
             var target = path.normalize(dependency.file.replace(absRoot, tempBundleDir));
             var compressionOptions = deployLists.compressionOptions[dependency.packageName];
 
@@ -812,7 +813,7 @@ actions.tarBundle = function(opts) {
 
             if (isJS) {
               try {
-                source = actions.compress(source, compressionOptions);
+                source = actions.compress(source, compressionOptions, file);
               } catch (error) {
                 reject(error);
               }
@@ -1018,7 +1019,7 @@ actions.project = function(options) {
   return new Project(options);
 };
 
-actions.compress = function(source, options) {
+actions.compress = function(source, options, file) {
   source = typeof source === 'string' ? source : source.toString();
 
   options = options || {
@@ -1094,6 +1095,7 @@ actions.compress = function(source, options) {
       // the edge cases ourselves. This also means
       // that ES6 code won't trigger false failures
       // on minifier parsers that haven't been updated.
+      logs.warn(uglifyError.message + ' at line ' + uglifyError.line + ':' + uglifyError.col + ' in file: ' + file);
       return source;
     }
   }
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -1095,7 +1095,7 @@ actions.compress = function(source, options, file) {
       // the edge cases ourselves. This also means
       // that ES6 code won't trigger false failures
       // on minifier parsers that haven't been updated.
-      logs.warn(uglifyError.message + ' at line ' + uglifyError.line + ':' + uglifyError.col + ' in file: ' + file);
+      logs.warn(`${uglifyError.message} at line ${uglifyError.line}: ${uglifyError.col} in file: ${file}`);
       return source;
     }
   }


Student007:
an alternative solution would be to add a debug mode and only warn once like this:
```
WARN Problems detected while shrinking files. 
INFO Please run the command with `DEBUG=deploy` to get more information.
```
The debug mode would do what this PR is doing.
Maybe this feels much cleaner ?

lib/tessel/deploy.js
```diff
@@ -1094,6 +1095,7 @@ actions.compress = function(source, options) {
       // the edge cases ourselves. This also means
       // that ES6 code won't trigger false failures
       // on minifier parsers that haven't been updated.
+      logs.warn(uglifyError.message + ' at line ' + uglifyError.line + ':' + uglifyError.col + ' in file: ' + file);
```
rwaldron: Please use template strings here

johnnyman727:
@Student007 our organization's standard process (explained in more detail [here](https://github.com/tessel/project/blob/master/CONTRIBUTING.md#pull-requests)) for introducing changes is to:
1. Create an issue that describes the bug or missing feature. What did you expect and what actually happened? How do we reproduce?
2. Continue a discussion on the issue until a consensus is reached on the best solution.
3. An implementation is created and a PR opened.
4. The code implementation is reviewed by the team and merged once comments are addressed.

The process has several benefits:
1. Ensures discussion is cleanly disambiguated between solution discussion (on issue) and implementation discussion (on PR) so it's easier to find info later.
2. Ensures you don't waste time on an implementation that is subpar.
3. Ensures the community doesn't waste time on reviewing an implementation that will change.

I'm glad to see you submitting code again but it would benefit the community if you could follow this process. 

Once you create a Github issue for the problem you describe here, I'd be happy to follow up with my thoughts :+1:

Student007:
@johnnyman727 didn't see you to do so too. 
There are a lot of quick fixes from other (new) contributors are not required to be required by that ruling.

This PR is reasonable and a solution until a better way is found. Also it helps others to find their way. You can require people to do so, **if you pay them but not if they make a gift**. So take it, override it later or let it be.
And btw: We wasted a lot of time on a lot of issues by discussing with less use. This leads me to **decide my own how I do contributing**. You don't pay me - so you don't decide how I do things - you simple decide what to take and what you don't.
For me the job is done with a suggestion which can be **either a PR or an Issue. Remember: I do and did both !!!**
And if there are quality tasks I fixed it every time which a PR is the right place for (review).

johnnyman727:
> @johnnyman727 didn't see you to do so too. 
There are a lot of quick fixes from other (new) contributors are not required to be required by that ruling.

You make a good point that we haven't been doing a good job of holding all developers (myself included) to the same process. I do believe it's the best process to use in order to keep code and user experience quality high as I described above so I intend to be more strict about the process moving forward.

> You don't pay me - so you don't decide how I do things - you simple decide what to take and what you don't.

Our organization cannot force you to do anything, paid or not. However, we have decided as a community on the process that we want to use in order to collaborate and we won't accept code from members that do not abide by that process. If you think we should change our process, I encourage you to make an issue on our `project` repo describing the change you'd like to make.

I certainly hope that you will choose to cooperate with the rest of the team because you've provided some excellent contributions in the past and I'd be sorry to see your interest shift.

rwaldron:
After reviewing this further and looking at it in context of the operation, I realized this is not at all what we want to do. It's possible for uglify to fail to parse code that is valid, because uglify is way behind the JavaScript implementation provided by V8 in Node.js versions >=4.2.1. Here's a concrete example: Write a valid program that uses an Arrow Function. That should run on the Tessel 2 without any issue, but uglify can't parse it. The approach we took was to simply back off the compression in those cases and send the source unchanged. If there are _actual_ syntax errors, they will be caught by the runtime itself. Putting this error here contradicts that strategy.

Student007:
> make an issue on our project repo describing the change you'd like to make.

@johnnyman727 https://github.com/tessel/project/issues/146

>  If there are actual syntax errors, they will be caught by the runtime itself

@rwaldron it was no runtime error ! I also run this code on a x64 platform and it works well.
There are in fact reasons when to warn.
But I would agree to change things in the future by switching between `DEBUG=deploy` and a `WARN Not all files affected by --slim (Please use DEBUG=deploy for more information)` where the files, errors and positions are named.

rwaldron:
> There are in fact reasons when to warn.

Can you provide examples?

Student007:
> Can you provide examples?

@rwaldron not at the moment - I am unable to reproduce the error.
So you are right.

I think the solution 
> WARN Problems detected while shrinking files. 
INFO Please run the command with `DEBUG=slim` to get more information.

would be **reasonable**. Maybe I got confused on my way to here because I didn't got a hint at the right stage of experiments. But I am still sure there was a difference between Tessel and Server - the Tessel runtime blamed.
I agree with the argument it would be to much blame if it WARN for any reason and for a working program due to working fallback to non slim.

I think the PR simple should get the following additional code:
```
actions.slimWarnDone = false;
```
replacing [logs.warn](https://github.com/tessel/t2-cli/pull/640/files#diff-191f4ebe176e9e816abc0445a664be98R1098)
```
if(!actions.slimWarnDone){
    actions.slimWarnDone = true;
    logs.warn('Problems detected while shrinking files');
    logs.info('Please run the command with DEBUG=slim to get more information.');
}
var debug = require('debug')('slim');
debug(`${uglifyError.message} at line ${uglifyError.line}: ${uglifyError.col} in file: ${file}`); 

```
Maybe you have a better switching solution than `action.slimWarnDone` :smile:

PR #643, opened 2016-04-08
Update node-usb to 1.1.2

johnnyman727:


LinusU:
Is there a reson for not changing this to `^1.1.2`?

johnnyman727:
@LinusU just an unreasonable paranoia about not having complete control. It would probably be a better solution in the long term.

rwaldron:
I'm not opposed to that paranoia ;)

PR #646, opened 2016-04-09
Type of option 'password' should be string

hoegertn:
When your wifi password contains only numbers and is long the cli converts it to a number with exponential representation. This breaks the configuration.

```
$ t2 wifi -n MySSID -p 432565XXXXXXXXXXXXXXXXXXXXXXXXXXXXX8743
INFO Connected to Tessel.
INFO Wifi Enabled.
INFO Wifi Connected. SSID: MySSID, password: 4.32565XXXXXXXXXXXXX+39, security: psk2
```

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -228,6 +228,7 @@ makeCommand('wifi')
   .option('password', {
     abbr: 'p',
     metavar: 'PASSWORD',
+    type: 'string',
     help: 'Set the password of the network to connect to'
   })
   .option('security', {


rwaldron:
Great catch!

rwaldron:
I'm trying to write a test for this and the test revealed that even with the `type: "string"` set, the value is being coerced into a number with an exponent part.

hoegertn:
it worked on the cli. I was able to connect my wifi

rwaldron:
I assumed that was the case—I didn't think you'd make the patch otherwise ;)

rwaldron:
I'll worry about tests later, this stays in.

rwaldron:
@hoegertn I totally forgot to make sure that you had signed the CLA. If you haven't done so, please do that here: http://dojofoundation.org/about/claForm

hoegertn:
done

rwaldron:
Brilliant! Thank you kindly

PR #648, opened 2016-04-10
ensure WiFi is enabled before scanning; closes #620

### LINKED ISSUE #620: [wifi] --list fails if not connected to wifi
If a Tessel is not actively connected to a network, it will fail to list available networks. This is really a problem because many people like to ensure their Tessel can detect their network before attempting to connect. Originally reported in https://github.com/tessel/t2-cli/issues/619.

Steps to reproduce:
```
t2 wifi --off
t2 wifi --list
```

Expected Output:
```
➜  v2-firmware git:(master) ✗ t2 wifi --list
INFO Connected to Tessel-02A36F738EAE.
INFO Scanning for available networks...
INFO Currently visible networks (9):
INFO 	 Fried Chicken Sandwich (70/70)
INFO 	 technicallyHome (51/70)
INFO 	 NETGEAR61 (43/70)
INFO 	 xfinitywifi (39/70)
INFO 	 Friendship Pizza (35/70)
INFO 	 NETGEAR11 (31/70)
INFO 	 SETUP (27/70)
INFO 	 HOME-F652 (27/70)
INFO 	 HOME-D338 (23/70)
```

Actual Output:
```
➜  v2-firmware git:(master) ✗ t2 wifi --list
INFO Connected to Tessel-02A36F738EAE.
INFO Scanning for available networks...
ERR! Error: No such wireless device: wlan0
```

boneskull:
wasn't exactly sure how you wanted this to be tested.

at any rate, I just got my T2 in the mail today, and the first thing I did was scan for networks...whoops.  This should make the OOBE a little smoother.

### DIFF
--- a/lib/tessel/wifi.js
+++ b/lib/tessel/wifi.js
@@ -38,6 +38,12 @@ Tessel.prototype.findAvailableNetworks = function() {
 
         return networks;
       }, []).sort(compareBySignal);
+    })
+    .catch(function checkNoSuchDeviceError(err) {
+      if (/No such wireless device/.test(err.message)) {
+        return Promise.reject(new Error('Unable to find wireless networks while wifi is disabled. Run "t2 wifi --on" before trying again.'));
+      }
+      return Promise.reject(err);
     });
 };
 


johnnyman727:
@boneskull thanks for getting involved!

I'd like to better understand how you hit this issue. Did you connect your Tessel to wifi and then disable your wifi? Or were you intending to enable your computer's wifi before scanning for your Tessel?

lib/tessel/wifi.js
```diff
@@ -12,7 +12,9 @@ var Tessel = require('./tessel');
 
 Tessel.prototype.findAvailableNetworks = function() {
   logs.info('Scanning for available networks...');
-  return this.simpleExec(commands.scanWiFi())
+  var scanWiFi = () => this.simpleExec(commands.scanWiFi());
+  return this.simpleExec(commands.turnOnWifi(true))
+    .then(scanWiFi)
```
LinusU: I feel that it's unnecessary to store the function in a variable, how about just adding the following?

```javascript
.then(() => this.simpleExec(commands.scanWiFi()))
```

boneskull: sure.  Was just trying to pattern after other code in the file

boneskull:
> I'd like to better understand how you hit this issue. Did you connect your Tessel to wifi and then disable your wifi? Or were you intending to enable your computer's wifi before scanning for your Tessel?

1.  Unwrap Tessel
2.  Plug into USB
3.  `npm install -g t2-cli`
4.  `t2 list`
5.  `t2 rename squirtle`
6.  `t2 wifi --list`; error

lib/tessel/wifi.js
boneskull: done

johnnyman727:
@boneskull sorry to draw this out but I can't reproduce with those steps. What was the error that you saw?

HipsterBrown:
@boneskull Thanks for catching this. 

Does this fix work for you? I'll test it out tonight, but enabling the Wifi takes 3 commands, see [Tessel.prototype.setWiFiState](https://github.com/tessel/t2-cli/pull/648/files#diff-ad1c3af214118b233a97e4dd4f67d729R149) as a reference.

boneskull:
@johnnyman727 Yes, that's the one, "no such wireless device: wlan0"

boneskull:
@HipsterBrown Yes, the fix works for me.  We don't need to execute the other two commands, because they are telling wifi to actually *connect* to something, which is not necessary if we're just checking the available networks.

boneskull:
@HipsterBrown running `turnOnWiFi(true)` simply enables the adapter, afaik.

HipsterBrown:
@boneskull Thanks for the update. I'll work on getting a smoke test together tonight or Thursday at the latest. We should add some tests to make sure this doesn't break in the future. :) Let me know if you have any questions about that.

rwaldron:
@HipsterBrown what's the status on this?

HipsterBrown:
@rwaldron I will have a smoke test done today. We should probably still have tests to cover this.

HipsterBrown:
## Smoke Test

**Setup:**
- Both `wifi` and `ap` are turned off. (which should mimic a fresh T2)
- Run `t2 wifi --list`

**Result:**
```
Nicks-MacBook-Pro:tessel-test hipsterbrown$ t2 wifi --list
INFO Scanning for available networks...
ERR! Error: No such wireless device: wlan0

```

The `wifi` doesn't get enabled, so Tessel can't scan for available networks. Even when I have the following:

```
return this.simpleExec(commands.turnOnWifi(true))
    .then(() => this.simpleExec(commands.commitWirelessCredentials()))
    .then(() => this.simpleExec(commands.reconnectWifi()))
    .then(() => this.simpleExec(commands.scanWiFi()))
    .then(function wifiScanResults(result) {
```

The `this.simpleExec(commands.reconnectWifi())` Promise resolves before Tessel's wifi is done restarting, so I get the same result as above. 

I ran into a similar issue when adding WiFi capabilities to `t2-firmware`. [Reference](https://github.com/tessel/t2-firmware/pull/140/files#diff-84087272b70af9498872cf6f791003abR1090)

If you look at the `scanWifi` function in `node/tessel-export.js` in the `t2-firmware` repo ([Reference](https://github.com/tessel/t2-firmware/pull/140/files#diff-84087272b70af9498872cf6f791003abR1236)), in order to scan for networks as soon as the chip is ready, I had to recursively check the result of the scan. 

Options for moving forward:
- do the recursive check after immediately enabling `wifi`
- don't enable wifi within `Tessel.prototype.findAvailableNetworks`, instead return  a more helpful message if `wifi` is disabled, i.e. "Tessel can't scan for networks while the wifi is off. Run 't2 wifi --on` and try again."

While the 2nd option would be a quick fix, the 1st option will be a better user experience overall. 

@boneskull Please let me know if you have any questions or concerns about this.

rwaldron:
@HipsterBrown what if we do option 2 in this patch, then you and @boneskull can work on option 1 in a follow up?

HipsterBrown:
@rwaldron I think that's a great solution. :+1:

boneskull:
If I understand correctly, I need to turn wifi on, then basically poll until it's ready?

HipsterBrown:
@boneskull Yes, that sums it up fairly well. Like @rwaldron, if we want to get a patch in to tell folks to manually enable wifi when it's not enabled by running `t2 wifi --on` before we work on the polling, that may be the best course of action.

boneskull:
right.  I'll change this to message the user.

boneskull:
hope this is a good stop-gap for now

lib/tessel/wifi.js
```diff
@@ -38,6 +38,12 @@ Tessel.prototype.findAvailableNetworks = function() {
 
         return networks;
       }, []).sort(compareBySignal);
+    })
+    .catch(function checkNoSuchDeviceError(err) {
+      if (/No such wireless device/.test(err.message)) {
+        return Promise.reject(new Error('Cannot scan; enable WiFi first using "t2 wifi --on"'));
```
HipsterBrown: So "Cannot scan" might be odd feedback for someone who just entered `t2 list` and expected a list of wireless networks. Maybe something a bit more verbose will clear up why that command did not work. 

"Unable to find wireless networks while wifi is disabled. Run "t2 wifi --on" before trying again."

Pinging @Frijol for feedback on this copy as well.

Frijol: +1 I like @HipsterBrown's proposed copy

boneskull: k.  will implement later today. unless I forget.

johnnyman727: ping @boneskull because he forgot :)

boneskull: I'll do this tonight

boneskull:
@johnnyman727 done

johnnyman727:
Great, thanks! Could you sign [our CLA](dojofoundation.org/about/claForm) and ping me once that's done? I'll merge immediately after.

boneskull:
@johnnyman727 done

cc @rwaldron

rwaldron:
@boneskull 🎉 ❤️

johnnyman727:
Awesome, thanks @boneskull!

Issue #644, opened 2016-04-08
New warning about tesselinclude is annoying


rwaldron:
I'm over here testing different versions of a program, generating a bunch of files in a directory and every time I run code, this warning comes up. I don't know what we should do.

Student007:
@rwaldron  we could add a `--silence` flag. Or as I suggested [here](https://github.com/tessel/t2-cli/pull/640#issuecomment-207446672) we could change things into a debugging mode but blame in a single line with the hint to use debugging mode for more information.

> an alternative solution would be to add a debug mode and only warn once like this:
```
WARN Problems detected while shrinking files. 
INFO Please run the command with `DEBUG=deploy` to get more information.
```
The debug mode would do what this PR is doing.
Maybe this feels much cleaner ?

Student007:
@rwaldron sorry - mixed up things.

Student007:
The text should be part of `--help` and the the WARN should be like this
> WARN Ignored folders due to missing .tesselinclude (see: t2 run --help)

instead of
>WARN The following items were found in your project directory, but not deployed.

          .tesselinclude
    auth.js
    login.html
    public

        If you need to deploy them to your Tessel 2, you must have a .tesselinclude
        file that lists the file and directory assets for your application. This can be
        created manually or by typing 't2 init'.

        For more information, visit: https://tessel.io/docs/cli#starting-projects

boneskull:
I was going to address this, but due to #649 I couldn't make heads or tails of what I was doing.

PR #655, opened 2016-04-10
Revised "overlooked asset" detection; tesselinclude hinting. Fixes gh-644

rwaldron:
This revision fixes some annoyances and garbage that I overlooked in the first implementation: 

1. This revised warning is _never_ shown for --full, because that makes no sense, all files will be pushed every time. 
2. Takes into consideration that asset files and directories might indeed be directly required by the program. 
3. Doesn't display such an awful message, no more file list, etc. 
4. Puts the information in the --help text (both run and push)


rwaldron:
cc @johnnyman727 @HipsterBrown

johnnyman727:
r+

HipsterBrown:
Very smart system. Nice that we can scan the dependency graph like that. Great update. :+1:

rwaldron:
Thanks everyone

PR #660, opened 2016-04-11
Dependency: update to t2-project@0.2.0; Use explicit dirname for dependency graph.

rwaldron:
Closes tessel/ambient-attx4#53


rwaldron:
> Who knew module bundling could be so complicated? :P


A real goddamn nightmare.

Issue #661, opened 2016-04-12
t2 update: autocheck for out of date CLI


Frijol:
..and prompt update if out of date. T1 CLI did this as well as https://github.com/tessel/t2-cli/issues/257 (the same, but for firmware).

Important, because otherwise there's every chance we'll fix something, but users won't know to update for a fix.

LinusU:
I can recommend [update-notifier](https://npmjs.org/packages/update-notifier), it's widely used and it's awesome!

Frijol:
oh, perfect! Thanks!

PR #665, opened 2016-04-12
Adds update checker for T2 CLI

Frijol:
(fixes https://github.com/tessel/t2-cli/issues/661)

To try it, change the package.json version to something old, and replace line 12 in /bin/tessel-2.js with something like `updateNotifier({pkg, updateCheckInterval: 1000}).notify();` so that it doesn't wait the default length of a day in order to check.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -5,6 +5,13 @@ var path = require('path');
 
 // Third Party Dependencies
 var parser = require('nomnom').script('t2');
+const updateNotifier = require('update-notifier');
+
+// Check for updates
+const pkg = require('../package.json');
+updateNotifier({
+  pkg
+}).notify();
 
 // Internal
 var controller = require('../lib/controller');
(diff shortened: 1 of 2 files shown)


Frijol:
@rwaldron and/or @johnnyman727 r?

johnnyman727:
Looks good to me and smoke tests pass. If Appveyer and Travis give the :+1:, feel free to merge.

LinusU:
I'm a bit late here but `const` is used outside of strict mode here which means that it will use the very old definition of `const` that didn't make it into any standard (as far as I remember).

I think that we should add `'use strict'` to the top of this file...

bin/tessel-2.js
```diff
@@ -5,6 +5,13 @@ var path = require('path');
 
 // Third Party Dependencies
 var parser = require('nomnom').script('t2');
+const updateNotifier = require('update-notifier');
+
+// Check for updates
+const pkg = require('../package.json');
+updateNotifier({
+  pkg
```
rwaldron: @Frijol <3 property shorthands

johnnyman727:
@LinusU what are the repercussions of not having `use strict` at the top of a file?

PR #671, opened 2016-04-14
show tessel.connectionType before tessel.name

dudleyjosh:
moved tessel.connectionType in front of tessel.name for `t2 list` to improve readability of output.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -104,7 +104,7 @@ Tessel.list = function(opts) {
       }
 
       // Print out details...
-      logs.basic(sprintf('\t%s\t%s\t%s', tessel.name, tessel.connection.connectionType, note));
+      logs.basic(sprintf('\t%s\t%s\t%s', tessel.connection.connectionType, tessel.name, note));
     });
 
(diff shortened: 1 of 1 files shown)


dudleyjosh:
Looks like the [install](https://github.com/tessel/t2-start/blob/gh-pages/install.md) page of the tessel.io will need to be updated to reflect these changes as well.

rwaldron:
Great start!

You'll need to update this part as well: https://github.com/dudleyjosh/t2-cli/blob/jd-reformat-t2-list-output/lib/controller.js#L230-L235

dudleyjosh:
@rwaldron that section is updated... code below was copied from the [L230-L235](https://github.com/dudleyjosh/t2-cli/blob/jd-reformat-t2-list-output/lib/controller.js#L230-L235)?

```javascript
var display = sprintf(
                          '\t%s\t%s\t%s',
                          tessel.connection.connectionType,
                          tessel.name,
                          authorization
                        );
```

rwaldron:
Yep, I'm maybe a little bit crazy? Sorry for any confusion

Issue #619, opened 2016-03-07
tessel wifi --list fails and tessel wifi -n <name> -p <password> does not work.


tikurahul:
I am trying to use my new Tessel 2, and for whatever reason I cannot get the Tessel 2 to connect to my Wireless network.

If i use the command `tessel wifi --list` I just get the following exception.

```
INFO Looking for your Tessel...
INFO Connected to rahulrav over USB
INFO Scanning for available networks...
events.js:154
      throw er; // Unhandled 'error' event
      ^

Error: Error upon remote process exit: 4
    at USBProcess.<anonymous> (/usr/local/lib/node_modules/t2-cli/lib/usb/usb_process.js:49:26)
    at USBProcess.g (events.js:273:16)
    at emitOne (events.js:90:13)
    at USBProcess.emit (events.js:182:7)
    at Parser.<anonymous> (/usr/local/lib/node_modules/t2-cli/lib/usb/usb_daemon.js:121:14)
    at emitOne (events.js:90:13)
    at Parser.emit (events.js:182:7)
    at Parser._write (/usr/local/lib/node_modules/t2-cli/node_modules/usb-daemon-parser/index.js:152:16)
    at doWrite (_stream_writable.js:301:12)
    at writeOrBuffer (_stream_writable.js:287:5)
    at Parser.Writable.write (_stream_writable.js:215:11)
    at ondata (_stream_readable.js:536:20)
    at emitOne (events.js:90:13)
    at emit (events.js:182:7)
    at readableAddChunk (_stream_readable.js:153:18)
    at Readable.push (_stream_readable.js:111:10)
```

I modified `wifi.js` with the following patch:

```
diff --git a/lib/tessel/wifi.js b/lib/tessel/wifi.js
index d70fb1b..26367cc 100644
--- a/lib/tessel/wifi.js
+++ b/lib/tessel/wifi.js
@@ -38,6 +38,10 @@ Tessel.prototype.findAvailableNetworks = function() {

         return networks;
       }, []).sort(compareBySignal);
+    }).catch(function(error) {
+      logs.err('Error while scanning for available networks...', error);
+      // continue to throw the error
+      throw error;
     });
 };
```

With my changes, when step through the code using `node-inspector` I see the following:

```
➜  CLI git:(master) ✗ node-debug t2 wifi --list
Node Inspector v0.12.7
Visit http://127.0.0.1:8080/?port=5858 to start debugging.
Debugging `t2`

Debugger listening on port 5858
[...]

PR #678, opened 2016-04-15
add option n for update command to not save config

thojansen:
While trying to set up wifi manually via ssh and uci commands, my wifi settings were somehow broken. Even manually setting the wifi settings to another ad-hoc network was no longer working as only parts are overwritten by the cli tool and ssh was no longer working as there was no wifi possible anymore. On the other hand the update did not reset my configuration, so I essentially needed this fix to unbrick my tessel2. Any feedback is appreciated to get this merged. Workaround was originally posted in #619. @johnnyman727

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -353,6 +353,12 @@ makeCommand('update')
     flag: true,
     help: 'Update to the latest version regardless of current version.'
   })
+  .option('delete', {
+    abbr: 'd',
+    required: false,
+    flag: true,
+    help: 'Delete configuration files during update.'
+  })
   .option('openwrt-path', {
     abbr: 'op',
     required: false,
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -353,11 +353,11 @@ makeCommand('update')
     flag: true,
     help: 'Update to the latest version regardless of current version.'
   })
-  .option('delete', {
-    abbr: 'd',
+  .option('n', {
+    abbr: 'n',
     required: false,
     flag: true,
-    help: 'Delete configuration files during update.'
+    help: 'Do not save configuration during update.'
   })
   .option('openwrt-path', {
     abbr: 'op',
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -800,7 +800,7 @@ controller.printAvailableUpdates = function() {
   });
 };
 
-controller.update = function(opts) {
+controller.update = function(opts) {  
   opts.authorized = true;
   return controller.standardTesselCommand(opts, function(tessel) {
     if (opts.openwrtPath || opts.firmwarePath) {
@@ -813,7 +813,7 @@ controller.update = function(opts) {
 
 controller.updateWithLocalBuilds = function(opts, tessel) {
   return updates.loadLocalBinaries(opts)
-    .then((images) => tessel.update(images))
+    .then((images) => tessel.update(opts, images))
     .then(() => logs.info('Finished updating Tessel with local builds.'));
 };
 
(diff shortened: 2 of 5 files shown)


lib/controller.js
```diff
@@ -800,7 +800,7 @@ controller.printAvailableUpdates = function() {
   });
 };
 
-controller.update = function(opts) {
+controller.update = function(opts) {
```
rwaldron: This shouldn't be here ;)

You need to run `grunt` locally, which will clean up things like that for you and also ensures that all the linting and tests are run. Ping me after you've had a chance to do that.

bin/tessel-2.js
```diff
@@ -353,6 +353,12 @@ makeCommand('update')
     flag: true,
     help: 'Update to the latest version regardless of current version.'
   })
+  .option('delete', {
+    abbr: 'd',
+    required: false,
+    flag: true,
+    help: 'Delete configuration files during update.'
+  })
```
rwaldron: This is a strange name for a thing associated with `t2 update`

thojansen: Yepp, agreed. Here the docs.. maybe just re-use it?
https://wiki.openwrt.org/doc/techref/sysupgrade

` -n do not save configuration over reflash`

rwaldron:
Travis is failing because this changeset breaks the update tests: https://travis-ci.org/tessel/t2-cli/builds/123382880

Please have a look and update them appropriately. You'll also need to add new tests for the new capability. Ping me here or on Slack for any help getting started with testing.

thojansen:
@rwaldron now the option is renamed from `d` to `n` matching sysupgrade. Linting and unit tests are passing again and two additional tests were added (although i am not 100% happy with the tests).

test/unit/update.js
```diff
@@ -350,7 +350,55 @@ exports['controller.update'] = {
...
         test.equal(this.update.callCount, 1);
         // It was provided the binaries
-        test.equal(this.update.calledWith(binaries), true);
+        test.equal(this.update.calledWith(opts, binaries), true);
+        // Then Tessel was closed
+        test.equal(this.tessel.closed, true);
+        // We closed all open Tessel connections
+        test.equal(this.closeTesselConnections.callCount, 1);
+        // We called the close function with an array
+        test.equal(Array.isArray(this.closeTesselConnections.args[0]), true);
```
rwaldron: Great start here! For this test, use test.deepEqual to compare arg[0] with the array you expect it to be

test/unit/update.js
```diff
@@ -679,6 +727,34 @@ exports['Tessel.update'] = {
...
+    this.sysupgradeNoSaveConfig = this.sandbox.stub(commands, 'sysupgradeNoSaveConfig');
+
+    this.tessel.updateOpenWRT({
+      n: true
+    }, this.newImage.openwrt).then(() => {
+      test.equal(this.openStdinToFile.callCount, 1);
+      test.equal(this.sysupgradeNoSaveConfig.callCount, 1);
+      test.equal(this.openStdinToFile.lastCall.args[0], updatePath);
+      test.equal(this.sysupgradeNoSaveConfig.lastCall.args[0], updatePath);
+      test.done();
```
rwaldron: Similar to the test I described above, add a test that asserts the right args were used

rwaldron:
This is fantastic work! I left a couple test related comments, once those are addressed, I will do a smoke test with an actual T2

thojansen:
Added deepEqual for the first test. For the second one, the options are actually not passed to the commands, but a different command is used. I choose to implement it this way, as there was basically no logic in `commands.js`.

rwaldron:
If you haven't done so already, please sign the CLA: http://dojofoundation.org/about/claForm

test/unit/update.js
rwaldron: Derp, this is not what I thought it was... Ignore me here ;)

thojansen:
CLA signed

rwaldron:
Thank you kindly :)

Issue #604, opened 2016-02-25
Make a command for fetching the current version of the CLI


johnnyman727:
This is useful for debugging purposes.

The likely implementation will take the current `t2-cli` commit sha and either compare it to Github tags (where we save each version) or we'll save a copy of each version on our AWS server like we did with Tessel 1.

wyze:
Couldn't this be as simple as just reading in the version field from `package.json`? I added the output really quick and could look something like this:

```
❯ t2 version
INFO Looking for your Tessel...
INFO Connected to t2-testing.
INFO Tessel [t2-testing] CLI version: 0.0.18
INFO Tessel [t2-testing] Firmware version: 0.0.12
INFO Tessel [t2-testing] Node version: 4.2.1
```

johnnyman727:
@wyze It's been a while now but what I think I meant was that we want a way to compare the version of the CLI being used with the most recently published (which I referred to as "current") CLI. In any case, that was closed with #665 but outputting the CLI version in the `t2 version` command would definitely still be useful.

wyze:
Alright, that makes sense. I will send a PR over adding CLI version to output of `t2 version` and it can be discussed there further if the team would like it to be added or not. Thanks for the clarification.

johnnyman727:
:+1:

PR #679, opened 2016-04-15
Add Node version output to `t2 version` command

### LINKED ISSUE #667: Add a command to display Node version
[Reference](https://forums.tessel.io/t/what-version-of-node-does-tessel-2-use/2349)

wyze:
Fixes #667 

This adds additional information to the `t2 version` command. If for some reason it fails to fetch the Node version, but was able to get the firmware version, it will output just that along with unknown for the Node version.

If it fails to get the firmware version first then it skips trying to get the Node version and just outputs unknown for both versions.

All the tests pass. There weren't any currently testing `controller.tesselFirmwareVerion` function, so I added those with this as well since I touched that function. I would love some feedback on the tests as I am pretty green with Sinon. :)

Also, note that `Verion` above should be `Version`, but I opted to not correct the spelling and would tackle that in a separate PR. Trying not to do so much here, but if you would like, I could rename that in this PR as well.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -938,20 +938,32 @@ controller.updateTesselWithVersion = function(opts, tessel, currentVersion, buil
     });
 };
 
-controller.tesselFirmwareVerion = function(opts) {
+controller.tesselFirmwareVerion = opts => {
   opts.authorized = true;
-  return controller.standardTesselCommand(opts, function(tessel) {
+  return controller.standardTesselCommand(opts, tessel => {
+    var output = (thing, version) => `Tessel [${tessel.name}] ${thing} version: ${version}`;
+
     // Grab the version information
     return tessel.fetchCurrentBuildInfo()
-      .then(function(versionSha) {
-        return updates.requestBuildList().then(function(builds) {
+      .then(versionSha => {
+        return updates.requestBuildList().then(builds => {
           // Figure out what commit SHA is running on it
-          var version = updates.findBuild(builds, 'sha', versionSha).version;
-          logs.info('Tessel [' + tessel.name + '] version: ' + version);
+          var firmwareVersion = updates.findBuild(builds, 'sha', versionSha).version;
+
+          return tessel.fetchCurrentNodeVersion()
+            .then(nodeVersion => {
+              logs.info(output('Firmware', firmwareVersion));
+              logs.info(output('Node', nodeVersion));
+            })
+            .catch(() => {
+              logs.info(output('Firmware', firmwareVersion));
+              logs.info(output('Node', 'unknown'));
+            });
         });
       })
-      .catch(function() {
-        logs.info('Tessel [' + tessel.name + '] version: unknown');
+      .catch(() => {
+        logs.info(output('Firmware', 'unknown'));
+        logs.info(output('Node', 'unknown'));
       });
   });
 };
--- a/lib/tessel/version.js
+++ b/lib/tessel/version.js
@@ -1,0 +1,19 @@
+// System Objects
+// ...
+
+// Third Party Dependencies
+// ...
+
+// Internal
+var Tessel = require('./tessel');
+
+/*
+  Gathers node version.
+*/
+Tessel.prototype.fetchCurrentNodeVersion = function() {
+  return this.simpleExec(['node', '--version'])
+    .then(version => {
+      // strip the `v` preceding the version
+      return version.trim().substring(1);
+    });
+};
--- a/lib/tessel/tessel.js
+++ b/lib/tessel/tessel.js
@@ -126,4 +126,5 @@ require('./erase');
 require('./name');
 require('./provision');
 require('./update');
+require('./version');
 require('./wifi');
--- a/test/unit/controller.js
+++ b/test/unit/controller.js
@@ -350,6 +350,145 @@ exports['controller.runHeuristics'] = {
   }
 };
 
+exports['controller.tesselFirmwareVerion'] = {
+  setUp: function(done) {
+    this.sandbox = sinon.sandbox.create();
+    this.logsWarn = this.sandbox.stub(logs, 'warn', function() {});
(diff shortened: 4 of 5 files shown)


lib/tessel/version.js
```diff
@@ -0,0 +1,19 @@
...
+
+// Internal
+var Tessel = require('./tessel');
+
+/*
+  Gathers node version.
+*/
+Tessel.prototype.fetchCurrentNodeVersion = function() {
+  return this.simpleExec(['node', '--version'])
+    .then(function versionRead(version) {
```
tikurahul: Use arrow functions when you can.

lib/controller.js
```diff
@@ -946,11 +946,22 @@ controller.tesselFirmwareVerion = function(opts) {
         return updates.requestBuildList().then(function(builds) {
           // Figure out what commit SHA is running on it
           var version = updates.findBuild(builds, 'sha', versionSha).version;
-          logs.info('Tessel [' + tessel.name + '] version: ' + version);
+          var tesselVersion = 'Tessel [' + tessel.name + '] version: ' + version;
+
+          return tessel.fetchCurrentNodeVersion()
+            .then(function(nodeVersion) {
```
tikurahul: Arrow functions will make this more succinct.

lib/tessel/version.js
wyze: Okay. I was following what I saw, and it was a lot of mixed between `function` keyword and arrow functions.

tikurahul: I know.. :smile:. Going forward we want to avoid adding more, and cleaning up as we go.

lib/controller.js
```diff
@@ -946,11 +946,22 @@ controller.tesselFirmwareVerion = function(opts) {
         return updates.requestBuildList().then(function(builds) {
           // Figure out what commit SHA is running on it
           var version = updates.findBuild(builds, 'sha', versionSha).version;
-          logs.info('Tessel [' + tessel.name + '] version: ' + version);
+          var tesselVersion = 'Tessel [' + tessel.name + '] version: ' + version;
```
rwaldron: Update to use template strings

lib/controller.js
```diff
@@ -946,11 +946,22 @@ controller.tesselFirmwareVerion = function(opts) {
...
+            .catch(() => {
+              logs.info(tesselVersion);
+              logs.info('Tessel [' + tessel.name + '] Node version: unknown');
+            });
         });
       })
-      .catch(function() {
+      .catch(() => {
         logs.info('Tessel [' + tessel.name + '] version: unknown');
+        logs.info('Tessel [' + tessel.name + '] Node version: unknown');
```
rwaldron: Update all concatenated strings to use template strings.

lib/controller.js
rwaldron: Also, might be best if this was `Tessel [Foo] Firmware version: x.x.x`

lib/controller.js
```diff
@@ -946,11 +946,22 @@ controller.tesselFirmwareVerion = function(opts) {
...
       });
   });
 };
```
rwaldron: The more I think about it, the more I wish this was a serializable format, or at very least super easy to parse

lib/controller.js
wyze: I thought about adding the word firmware in there. I did add it. Then removed it before committing. Prefer it capitalized like you have it or lowercase?

Will get it updated with template strings when I get home.

lib/controller.js
wyze: What do you mean by that? Anything I can change to help?

lib/controller.js
```diff
@@ -937,20 +937,34 @@ controller.updateTesselWithVersion = function(opts, tessel, currentVersion, buil
...
 
-controller.tesselFirmwareVerion = function(opts) {
+controller.tesselFirmwareVerion = opts => {
   opts.authorized = true;
-  return controller.standardTesselCommand(opts, function(tessel) {
+  return controller.standardTesselCommand(opts, tessel => {
+    var tesselName = `Tessel [${tessel.name}]`;
+    var firmwareOutput = version => `${tesselName} Firmware version: ${version}`;
+    var nodeOutput = version => `${tesselName} Node version: ${version}`;
```
rwaldron: Very clean, I like this.

rwaldron: Let's iterate on it one more time ;)

```js   
var output = (thing, version) => `${tesselName} ${thing} version: ${version}`;
``` 

WDYT?

wyze: I like it. Then `tesselName` could just be in the one function. Then it also paves a way for when #604 gets added.

wyze:
Updated based on @rwaldron feedback and rebased onto latest master.

lib/controller.js
```diff
@@ -938,20 +938,32 @@ controller.updateTesselWithVersion = function(opts, tessel, currentVersion, buil
...
+          var firmwareVersion = updates.findBuild(builds, 'sha', versionSha).version;
+
+          return tessel.fetchCurrentNodeVersion()
+            .then(nodeVersion => {
+              logs.info(output('Firmware', firmwareVersion));
+              logs.info(output('Node', nodeVersion));
+            })
+            .catch(() => {
+              logs.info(output('Firmware', firmwareVersion));
+              logs.info(output('Node', 'unknown'));
```
rwaldron: 👍

rwaldron:
Great work!!

wyze:
Thanks! More to come.

lib/controller.js
rwaldron: This is part of a bigger issue that can be addressed once I've finished the migration to `npmlog`. When that's done I will call on you to help with this. Basically the idea is that other programs should be able to invoke these commands from spawn or exec and get useful and easily consumable information from them—but like I said, that needs to be on hold until after I finish the move to npmlog

PR #684, opened 2016-04-16
(Hopefully) Fixes mDNS Discovery issues

### LINKED ISSUE #650: [list] lan connection not found
I've run t2 wifi {credentials} and got a response of WIFI enabled. WIFI {connection info}, but when I run t2 update it can not find the lan connection.

HipsterBrown:
Fixes #650 (probably a few others too)

Found [`node_mdns`](https://github.com/agnat/node_mdns) to be more reliable and consistent for discovering the `_tessel._tcp` service. I'll need to test on more networks to be sure but it's looking good so far. 

Blockers:
- I'm having trouble authorizing the discovered Tessel, so I would appreciate some help from anyone familiar with that process.
- After trying `t2 provision`, I can't discover the Tessel until I restart it. 
- `node_mdns` has some installation complications on Linux & Windows ([Reference](https://github.com/agnat/node_mdns#installation))
  - the Linux issue could be solved with a `preinstall` script that check the system OS and fetches `libavahi-compat-libdnssd-dev` if needed
  - this definitely needs testing on Windows but the complicated setup could be a deal breaker. :(


johnnyman727:
@HipsterBrown I looked into using this library when first writing the discovery portion of the CLI but decided against it because it would have been too much work for me to make it cross-platform. 

I do think it's probably a more reliable library though so if we can get a functional setup for all platforms, that would be great. Maybe we can just use the original `mdns-js` library is a Windows OS is detected?

HipsterBrown:
@johnnyman727 Yea the cross-platform issues are a bummer. I'm going to do a bit more research around mDNS discovery on Windows, maybe we can avoid the Bonjour for Windows download complication. This is definitely a more reliable library, probably because it's using native bindings with `dns-sd`. 

> Maybe we can just use the original mdns-js library is a Windows OS is detected?

Is `t2 list` working for folks on Windows right now?

johnnyman727:
> Is t2 list working for folks on Windows right now?

For folks who were able to get `t2-cli` installed on Windows, I believe `t2 list` is working.

HipsterBrown:
@johnnyman727 

> For folks who were able to get t2-cli installed on Windows, I believe t2 list is working.

Good to know. I'll be looking into the [`bonjour`](https://github.com/watson/bonjour) package as well, which should have less installation complications since it's a pure JS alternative. If that isn't as reliable, I'll try the OS conditional option of using `mdns` for Unix and `mdns-js` for Windows. 

After doing a bit more research, it looks like we could eventually rely on just the `mdns` package, based on [this article](https://www.slightfuture.com/technote/windows-mdns-dnssd). That would make this process a bit easier.

rwaldron:
@HipsterBrown can you take a look at travis and appveyor? I didn't look myself, just noted that they're both red.

HipsterBrown:
@rwaldron Yea, it's because of the installation on Linux complication:

> the Linux issue could be solved with a preinstall script that check the system OS and fetches libavahi-compat-libdnssd-dev if needed

I haven't made that preinstall script yet, so the build is red. It is something I'll be working on this week.

rwaldron:
Oh, right. Duh. Like I said...


>  I didn't look myself, just noted

mattsoftware:
works for me

Might need router settings changed to allow IGMP Snooping and MLG snooping

bin/preinstall-check.js
```diff
@@ -0,0 +1,32 @@
...
+      throw error;
+    }
+
+    mdnsInstall();
+  });
+} else if (type === 'Darwin') {
+  mdnsInstall();
+} else {
+  process.exit(0);
+}
```
rwaldron: Instead of `os.type()`, you can use `process.platform`

rwaldron: Also, the function expressions can be arrow functions

lib/discover.js
```diff
@@ -48,7 +48,7 @@ TesselSeeker.prototype.start = function(opts) {
       .then(() => {
         debug('Connection opened:', tessel.connection.host);
         if (conn.connectionType === 'LAN' && !conn.authorized) {
-          tessel.name = conn.host.slice(0, -6);
+          tessel.name = conn.host.slice(0, -7);
```
rwaldron: Above this line, write an explanation comment that includes an example of the full `conn.host` string, as well as the sliced resulting string.

HipsterBrown: Good note. Will add that.

johnnyman727:
@HipsterBrown how are things going with this? It would be great to get it into the next release. 

With regards to the Linux deps, I recommend we just add those dependencies to the existing dependencies we list on the start experience for users to manually install. That will keep them from having to run any `t2` commands with `sudo`.

lib/discover.js
```diff
@@ -48,7 +49,8 @@ TesselSeeker.prototype.start = function(opts) {
       .then(() => {
         debug('Connection opened:', tessel.connection.host);
         if (conn.connectionType === 'LAN' && !conn.authorized) {
-          tessel.name = conn.host.slice(0, -6);
+          // depending on the platform the returned host name could be "tessel-name.local." or "tessel-name.local"
+          tessel.name = conn.host.slice(0, conn.host.search(LOCAL_REGEX));
```
LinusU: `String#search` works not only with regexes, but strings also. You could do `.search('.local')` instead of `.search(/\.local/)`

HipsterBrown:
While I didn't have much time to work on it while traveling the past 2 weeks, I had a great discussion with someone knowledgable of mDNS discovery with Node at JSConf Budapest. I'll be looking into some of their proposed solutions after I get back home. 

Concerns at the moment:
While `node-mdns` works very well on OS X, the complications with installing on Linux, running through Travis, and using a different package on Windows makes me doubtful of it as the correct solution. So I may return to investigating pure JS implementations. 

I will keep this updated as I continue to work on it.

HipsterBrown:
Closing this because I believe, after a lot of research and testing, `mdns-js` is still our best option for reliable, cross-platform device discovery. I'll be looking into other ways to resolve discovery issues, like telling devices to re-broadcast from a CLI command.

PR #685, opened 2016-04-17
Build the start script path correctly on Windows

jendib:
Linked to https://forums.tessel.io/t/tessel-2-does-not-start-my-application-with-push/2329/3

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -1035,7 +1035,7 @@ actions.pushScript = function(t, script, opts) {
 actions.writeToFile = function(t, entryPoint) {
   return new Promise(function(resolve, reject) {
     // Path of the script to run a Node project
-    var shellScriptPath = path.join(Tessel.REMOTE_PUSH_PATH, PUSH_START_SCRIPT_NAME);
+    var shellScriptPath = Tessel.REMOTE_PUSH_PATH + PUSH_START_SCRIPT_NAME;
     // Open a stdin pipe tp the file
     t.connection.exec(commands.openStdinToFile(shellScriptPath), (err, remoteProcess) => {
       if (err) {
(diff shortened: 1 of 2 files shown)


johnnyman727:
@jendib thanks for taking the initiative to fix this.

Could you explain why the `path.join` method doesn't work on Windows? What path does it create? I was under the impression that the purpose of `path.join` was to abstract different file system implementations  across platforms.

jendib:
Yes `path.join` indeed abstract the host FS, however in this case, we don't want this. We want to create the path to the init script used to launch the node app when Tessel is starting. So we need to build a unix path.
Before, this line does this: `\app\script`
And after: `/app/script`

rwaldron:
@johnnyman727 this is correct, for the same reason as: https://github.com/tessel/t2-cli/commit/50732a7162a2d3e9ddb35d196d530579249b0a45 I'll spend some time today checking for anymore mistakes like this 

@jendib two things:

- based on the commit I linked above, do you think you can write a test for this change? If you need help, ping me here or on slack—I'd be happy to help get you started
- If you haven't done so already, please sign the CLA: http://dojofoundation.org/about/claForm

rwaldron:
@jendib and I discussed a strategy for tests, all that remains now: 

- @jendib signs CLA (or just confirm that it's already signed)
- @jendib merges https://github.com/jendib/t2-cli/pull/1

Once those are complete, and this PR is green, then it's +1 to merge

jendib:
It's merged and I signed the CLA, thanks @rwaldron !

brunohunziker:
Could somebody make a new relaease containing the change?

lib/tessel/deploy.js
```diff
@@ -1035,7 +1035,7 @@ actions.pushScript = function(t, script, opts) {
 actions.writeToFile = function(t, entryPoint) {
   return new Promise(function(resolve, reject) {
     // Path of the script to run a Node project
-    var shellScriptPath = path.join(Tessel.REMOTE_PUSH_PATH, PUSH_START_SCRIPT_NAME);
+    var shellScriptPath = Tessel.REMOTE_PUSH_PATH + PUSH_START_SCRIPT_NAME;
```
LinusU: You could change this into `path.posix.join(...)` instead

jendib: Will do, thanks!

PR #689, opened 2016-04-18
Add CLI version to `t2 version` command output (#604)

### LINKED ISSUE #604: Make a command for fetching the current version of the CLI

wyze:
Fixes #604 

As discussed in #604, the initial issue was resolved with a separate PR, but this information might be handy as well.

If this is not wanted on the CLI, that's fine, we can just close the PR. But, it is up here for discussion.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -947,6 +947,7 @@ controller.updateTesselWithVersion = function(opts, tessel, currentVersion, buil
 controller.tesselEnvVersions = opts => {
   opts.authorized = true;
   return controller.standardTesselCommand(opts, tessel => {
+    var cliVersion = require('../package.json').version;
     var output = (thing, version) => `Tessel [${tessel.name}] ${thing} version: ${version}`;
 
     // Grab the version information
@@ -958,16 +959,19 @@ controller.tesselEnvVersions = opts => {
 
           return tessel.fetchCurrentNodeVersion()
             .then(nodeVersion => {
+              logs.info(output('CLI', cliVersion));
               logs.info(output('Firmware', firmwareVersion));
               logs.info(output('Node', nodeVersion));
             })
             .catch(() => {
+              logs.info(output('CLI', cliVersion));
               logs.info(output('Firmware', firmwareVersion));
               logs.info(output('Node', 'unknown'));
             });
         });
       })
       .catch(() => {
+        logs.info(output('CLI', cliVersion));
         logs.info(output('Firmware', 'unknown'));
         logs.info(output('Node', 'unknown'));
       });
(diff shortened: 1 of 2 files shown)


lib/controller.js
```diff
@@ -941,6 +943,8 @@ controller.updateTesselWithVersion = function(opts, tessel, currentVersion, buil
 controller.tesselEnvVersions = opts => {
   opts.authorized = true;
   return controller.standardTesselCommand(opts, tessel => {
+    var pkg = path.join(__dirname, '..', 'package.json');
+    var cliVersion = fs.readJsonSync(pkg).version;
```
rwaldron: Let's simplify that: 

```js
var cliVersion = require('../package.json').version;
```

lib/controller.js
```diff
@@ -952,16 +956,19 @@ controller.tesselEnvVersions = opts => {
...
+              logs.info(output('CLI', cliVersion));
               logs.info(output('Firmware', firmwareVersion));
               logs.info(output('Node', 'unknown'));
             });
         });
       })
       .catch(() => {
+        logs.info(output('CLI', cliVersion));
         logs.info(output('Firmware', 'unknown'));
         logs.info(output('Node', 'unknown'));
```
rwaldron: Sharp as hell

rwaldron:
Just one minor change and this is good to go

lib/controller.js
wyze: I did have this at first, then I was updating the tests, and I wasn't sure how to mock that. If that `require` isn't mocked then every time a new version of the cli was released, we would have to update the tests.

Thoughts?

rwaldron: The tests shouldn't be calling real `require`, you an stub it just for the test and the sandbox will restore it

rwaldron: *can

wyze: Private conversation in Slack with a solution:

```
rwaldron [7:26] 
Since your plan is to just include that value in the display, we don't really need to care about ​_what_​ the value is, just that it "is"

[7:30] 
So! You can get the ​_expected_​ value from the package.json in your test! Does that make sense?

wyze [7:31 AM] 
yeah just require in the package.json in the test? I was going to do that too, but it almost felt a little dirty. lol

rwaldron [7:31 AM] 
That means you can get rid of the stub all together, let it do the full operation, because it relies on invariants

[7:31] 
Naw, it's cool
```

wyze:
Updated as per our convo @rwaldron. Also rebased onto latest master.

johnnyman727:
Good to go! Thanks @wyze!

PR #690, opened 2016-04-18
Add contributors to package.json

### LINKED ISSUE #686: Adding Contributors to package.json
I think we should add the list of contributors to the `package.json`as described by [NPM](https://docs.npmjs.com/files/package.json#people-fields-author-contributors)

scottgonzalez:
Fixes #686

I added a `.mailmap` to handle some bad author data, but I had to guess for some of the authors. All of Flaki's commits had a bogus address for his laptop, so I chose his public email address from GitHub. Josh Dudley had two commits, each with a different address, so I chose the first one. If anyone wants their info changed, the `.mailmap` can be adjusted and the list will be corrected the next time `grunt update-contributors` is run.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,5 @@
+Flaki <git@flaki.hu> <flaki@Flakis-Air.local>
+Jia Huang <jialiya.huang0@gmail.com>
+Jon McKay <johnnyman727@gmail.com>
+Josh Dudley <jdudley@abrasive-tech.com> <dudley.21@osu.edu>
+Kelsey Breseman <ifoundthemeaningoflife@gmail.com>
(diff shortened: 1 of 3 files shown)


johnnyman727:
Fantastic, thanks @scottgonzalez!

Out of curiosity, where does it pull names from? It looks like I forgot to fill in my last name somewhere.

scottgonzalez:
It pulls the names from the author data in git, so you'd want to update your git config:

```sh
git config --global user.name "Jon McKay"
```

Of course that will only help with future commits. To deal with existing commits, we need to edit `.mailmap`. You can read about `.mailmap` in Git's `shortlog` docs or in [this article by Shane da Silva](https://shane.io/2011/10/07/git-shortlog-and-mailmap.html).

I'll update the PR to add your last name as well.

scottgonzalez:
updated

johnnyman727:
Cool, thanks @scottgonzalez!

PR #692, opened 2016-04-19
Disallow USB connections with `t2 root` command (#673)

### LINKED ISSUE #673: [root] Disallow the --usb option
`t2 root` only works over LAN but we currently allow the `--usb` option to be passed in. This will cause an error message like the one seen below:

```
supaspoida@faraday:~/code/tessel/cylon-bot(master+*)% t2 root --usb
INFO Connected to zaphod.
INFO Starting SSH Session on Tessel. Type 'exit' at the prompt to end.
ERR! TypeError: Cannot read property 'ip' of undefined
```

We should remove the ability to use the `--usb` option with this command.

Cross posted from [the forums](https://forums.tessel.io/t/unable-to-get-wifi-working-on-tessel-2/2367)

wyze:
Fixes #673 

This will hide the flags if someone runs `t2 root --help` but doesn't disable them completely. I saw at least 2 options here, either error out and don't attempt a connection, or force the use of LAN connection and give the user a warning. I chose the 2nd route as I think it is a better user experience. But I am up for discussions on it.

Current output:

```
❯ t2 root --usb
WARN You are trying to connect to Tessel via USB, but this command only works with Wifi.
WARN I will use Wifi to try and look for your Tessel.
INFO Connected to t2-testing.
INFO Starting SSH Session on Tessel. Type 'exit' at the prompt to end.
```

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -432,6 +432,18 @@ makeCommand('root')
   .callback(function(opts) {
     callControllerWith('root', opts);
   })
+  .option('lan', {
+    flag: true,
+    hidden: true
+  })
+  .option('lanPrefer', {
+    flag: true,
+    hidden: true
+  })
+  .option('usb', {
+    flag: true,
+    hidden: true
+  })
   .help('Gain SSH root access to one of your authorized tessels');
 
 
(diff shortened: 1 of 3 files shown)


rwaldron:
LGTM

@HipsterBrown @tikurahul R?

tikurahul:
LGTM

Issue #691, opened 2016-04-19
[crash-reporter] Crashes reported before being displayed


johnnyman727:
I'm running into an issue where a crash is being reported to the crash reporter service but the CLI is not actually displaying the error that was thrown.

Steps to reproduce:
```
mkdir audio
cd audio
t2 init
npm install tessel-av
wget http://www.stephaniequinn.com/Music/Vivaldi%20-%20Spring%20from%20Four%20Seasons.mp3 -O vivaldi.mp3
echo vivaldi.mp3 >> .tesselinclude
t2 run index.js
```
Output:
```
➜  audio t2 run index.js
INFO Looking for your Tessel...
INFO Connected to lyon.
INFO Building project.
INFO Writing project to RAM on lyon (550.4 kB)...
INFO Crash Reported: http://crash-reporter.tessel.io/
```

Perhaps an error is being thrown by the remote Node process and it's being gobbled by the crash reporter?

rwaldron:
Did you try this with the crash reporter off?

tikurahul:
That should not happen. Let me make sure that we don't swallow the real error. Also I should really update the message that says `Crash Report submitted at...` to include the fingerprint. We have that information.

tikurahul:
Oops. Fat fingered the close issue button. My apologies. 🙃

johnnyman727:
> Did you try this with the crash reporter off?

@rwaldron No, not yet but will do.

Also reported here: https://forums.tessel.io/t/red-error-led-on-boot-and-blue-led-blinking-forever/2401/4?u=jon

PR #696, opened 2016-04-20
Fixes https://github.com/tessel/t2-cli/issues/691

tikurahul:
Sometimes remote node crashes are swallowed by the Crash reporter. Now the uncaught exception
handler prints them, and the Crash reporter also returns the correct `fingerprint`, so the
crash can be discovered and reported.

### DIFF
--- a/lib/crash-reporter.js
+++ b/lib/crash-reporter.js
@@ -13,13 +13,14 @@ var Preferences = require('./preferences');
 // the value of the crash reporter preference
 // the value has to be one of 'on' or 'off'
 var CRASH_REPORTER_PREFERENCE = 'crash.reporter.preference';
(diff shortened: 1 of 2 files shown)


lib/crash-reporter.js
```diff
@@ -99,7 +96,8 @@ CrashReporter.post = function(labels, report) {
           if (json.error) {
             reject(json.error);
           } else {
-            resolve();
+            var fingerprint = json.crash_report.fingerprint;
+            resolve(fingerprint);
```
rwaldron:
```js
resolve(json.crash_report.fingerprint);
```

rwaldron:
@tikurahul since there only the one minor change needed, once you've committed that, you can merge this to master

tikurahul:
Sorry, just saw your comments now. Will fix, and merge when I get home :)

PR #702, opened 2016-04-22
Ipv6 link local fix

mattsoftware:
Fix for my local network using local link ip addresses (but from reading should be fine with other networks as well) that will append the network interface to the address if its a) an ipv6 address and b) the network Interface is known


HipsterBrown:
@mattsoftware Thanks for opening a pull request with your fix. You explain the problem a bit in this forum post -> https://forums.tessel.io/t/unable-to-t2-root-update/2417/3 

Could you explain the issue a bit more here, as well as how this PR fixes that issue?

mattsoftware:
The code determines the location of the tessel to connect to by either picking the first internet address in the opts.addresses array, and if that doesn't exist it falls back to opts.host. In my instance the opts.addresses array contained an ipv6 link local address as the first element, and the ipv4 address as the second. From my limited research I cannot connect to a link local address without also referencing the interface that the link local address can be reached on. 

i.e.

ping6 fe80::204:edff:fe1d:c93b # this does not work, as the address is link local
ping6 fe80::204:edff:fe1d:c93b -I eth0 # this works as I have specified the interface to ping on
ping6 fe80::204:edff:fe1d:c93b%eth0 # this also works

My pull request solves my issue by letting the exsting code work out what address to connect to. If the address contains a colon (:) AND there is a networkInterface also on the opts, then we append the % and the network interface to the end of the address.

I have limited resources to test this against other IPv6 networks, so I accept that there will need to be more testing before my pull request gets approved, but I suspect something similar to this will be the solution, even if it needs further refinement.

johnnyman727:
@mattsoftware thanks so much for investigating this! Before we can merge this, we need a unit test or two to make sure it doesn't break in the future. You can add them [in this file](https://github.com/tessel/t2-cli/blob/32564aa5f0b5407733b903577cab6597d36af34d/test/unit/lan-connection.js) and let me know if you have any questions about how `sinon`/`nodeunit` work (or anything).

lib/lan-connection.js
```diff
@@ -25,6 +25,9 @@ var LAN = {};
 LAN.Connection = function(opts) {
   this.auth = {};
   this.ip = Array.isArray(opts.addresses) ? opts.addresses[0] : opts.host;
+  if (this.ip && this.ip.indexOf(':') > -1 && opts.networkInterface) {
```
rwaldron:
```js
this.ip.includes(':')
```

HipsterBrown:
Thanks for the explanation @mattsoftware. You can read more about getting started with our testing setup in the [README](https://github.com/tessel/t2-cli#testing) and feel free to ping me if you have any questions or concerns. 

I'll look at prepping a smoke test this week. Then we can get more folks trying this on other networks.

lib/lan-connection.js
mattsoftware: Thankyou. I updated my pull request with this change.

mattsoftware:
@johnnyman727: Tests added, sorry for the delay. Hope they are acceptable :)

wyze:
You should add `test.expect(x);` to each test to be consistent with other tests.

johnnyman727:
@mattsoftware it looks pretty good! In addition to @wyze's comment above, could you also add some comments explaining the thought process behind each test? Once that's done, we'll be ready to merge.

mattsoftware:
What does the test.expect(x) do? It doesn't seem to be comparing the number to anything real.

wyze:
It sets the expected assertions, and throws an error if they don't add up. Example:

```
Testing index.jsF
>> Ambient - initialization
>> Error: Expected 11 assertions, 10 ran
>> at Object.exports.Ambient.tearDown (test/unit/index.js:12:5)
```

rwaldron:
LGTM!

Issue #708, opened 2016-04-26
`t2` with no command doesn't show the manual


Frijol:
(again)

```sh
➜  t2-cli git:(master) t2

command argument is required

Usage: t2 <command>

command     one of: install-drivers, crash-reporter, provision, restart, run, push, erase, list, init, wifi, key, rename, update, version, ap, root
```

Should show each of those commands with their descriptors.
Probably, one of the commands of the CLI is missing a .help field. To fix, make sure every command on the CLI has a .help field which describes what it is/how to use it.

rwaldron:
I found the missing "help", did you want to fix it, or should I make a PR?

Frijol:
you go ahead! Thanks

PR #709, opened 2016-04-26
t2: add missing help to crash-reporter. Fixes gh-708

rwaldron:
```
$ t2

command argument is required

Usage: t2 <command>

command
  crash-reporter      Crash Reporter enable/disable
  restart             Restart a previously deployed script in RAM or Flash memory (does not rebundle)
  run
    Deploy a script to Tessel and run it with Node.

    Assets that are not directly deployed as a dependency via require analysis,
    for example images or html files of an application (and their directories),
    must be listed in a .tesselinclude in the root of your project.
    This can be created manually or by typing 't2 init'.

    For more information, visit: https://tessel.io/docs/cli#starting-projects

  push
    Pushes the file/dir to Flash memory to be run anytime the Tessel is powered,
    runs the file immediately once the file is copied over.

    Assets that are not directly deployed as a dependency via require analysis,
    for example images or html files of an application (and their directories),
    must be listed in a .tesselinclude in the root of your project.
    This can be created manually or by typing 't2 init'.

    For more information, visit: https://tessel.io/docs/cli#starting-projects

  erase               Erases files pushed to Flash using the tessel push command
  list                Lists all connected Tessels and their authorization status.
  init                Initialize repository for your Tessel project
  wifi                Configure the wireless connection
  key                 Manage ssh keys for connecting to a Tessel
  rename              Change the name of a Tessel to something new
  update              Update the Tessel firmware and openWRT image
  version             Display Tessel's current firmware version
  ap                  Configure the Tessel as an access point
  root                Gain SSH root access to one of your authorized tessels
```

cc @Frijol 

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -115,7 +115,8 @@ parser.command('crash-reporter')
   .option('test', {
     flag: true,
     help: 'Test the Crash Reporter.'
-  });
+  })
+  .help('Crash Reporter enable/disable');
 
 parser.command('provision')
   .callback(callControllerCallback('provisionTessel'))


wyze:
LGTM

rwaldron:
@wyze thanks!

bin/tessel-2.js
```diff
@@ -115,7 +115,8 @@ parser.command('crash-reporter')
   .option('test', {
     flag: true,
     help: 'Test the Crash Reporter.'
-  });
+  })
+  .help('Crash Reporter enable/disable');
```
Frijol: How do you know if it's enabled or disabled? Do you run it with --enable or anything like that?

wyze: A few lines above it there is `--on` and `--off` flags. Not sure how to get the current state, or if there is a way or not.

Reference: https://github.com/tessel/t2-cli/blob/master/bin/tessel-2.js#L107-L114

tikurahul:
Its enabled by default. There is a way to toggle it from the command line. There is nothing that will tell you it's current state via the command line, but we can figure that out programmatically very easily via `preferences.js`. It should be a really easy feature to add, if we need it.

PR #713, opened 2016-04-27
t2: add missing help to crash-reporter (shows status). Fixes gh-708

rwaldron:

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -83,25 +83,24 @@ parser.command('install-drivers')
 
 parser.command('crash-reporter')
   .callback(opts => {
+    var cr = Promise.resolve();
+
     // t2 crash-reporter --on
     if (opts.on) {
-      return CrashReporter.on().then(() => {
-        // t2 crash-reporter --on --test
-        if (opts.test) {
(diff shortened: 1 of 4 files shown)


bin/tessel-2.js
```diff
@@ -116,7 +121,7 @@ parser.command('crash-reporter')
     flag: true,
     help: 'Test the Crash Reporter.'
   })
-  .help('Crash Reporter enable/disable');
+  .help('See Crash Reporter status');
```
johnnyman727: It seems like this help text needs to be a little more general because you can also set the status in addition to seeing it. Maybe it should be "View or set Crash Reporter status" or something to that effect.

rwaldron: I've updated it to use the same language that `t2 wifi` help uses: `Configure the Crash Reporter.`

johnnyman727:
Looks good!

rwaldron:
@johnnyman727 updated the help language + a bunch of tests

johnnyman727:
:+1: good to merge.

tikurahul:
LGTM :+1:

Issue #714, opened 2016-04-27
update-notifier is second worst cause of crashes


rwaldron:
- 11: http://crash-reporter.tessel.io/crashes?fingerprint=-0x20c7a96f
- 6: http://crash-reporter.tessel.io/crashes?fingerprint=-0x115ae372
- 5: http://crash-reporter.tessel.io/crashes?fingerprint=-0x17165b05
- 2: http://crash-reporter.tessel.io/crashes?fingerprint=-0x1c1fc797
- 1: http://crash-reporter.tessel.io/crashes?fingerprint=-0x10ad46c6
- 1: http://crash-reporter.tessel.io/crashes?fingerprint=-0x17b44ab4

rwaldron:
Related: https://github.com/yeoman/configstore/issues/6

johnnyman727:
@LinusU this is entirely your fault 😄

rwaldron:
FYI, there are a lot more of these

LinusU:
Whoa, that's too bad :(

I wonder why the users don't have access to something in their own home directory that probably wasn't there before...

Hmm, I'm going to do some digging...

jimbojw:
> I wonder why the users don't have access to something in their own home directory that probably wasn't there before...

I have a hunch that it's because `t2 install-drivers` has to be run as root. Looking in my own `.config/configstore` there are other files, but the only one that requires root to access is `update-notifier-t2-cli.json`:

```sh
$ ls -al ~/.config/configstore/
total 20
drwxrwxr-x  2 jimbo jimbo 4096 May 10 06:11 .
drwx------ 37 jimbo jimbo 4096 Mar  5 19:08 ..
-rw-rw-r--  1 jimbo jimbo   36 Jun 15  2014 insight-bower.yml
-rw-------  1 jimbo jimbo   45 Jun 15  2014 update-notifier-bower.yml
-rw-------  1 root  root    55 May 10 06:11 update-notifier-t2-cli.json
```

Per discussion in #727, install-drivers appears to be simplifiable to a pair of shell commands that would not require running node as root.

jimbojw:
Also, FYI, users can't even currently discover that this is their problem because the output of crash reporter doesn't give them a fingerprint.

```sh
$ t2 list
INFO Crash Reported: http://crash-reporter.tessel.io/
```

I understand that #730 / #696 addresses this, but until they're released it requires source code hacking to even find the keywords to search for. This is a blocking issue for Ubuntu users.

[...]

PR #715, opened 2016-04-27
Stop crashing when update-notifier fails

LinusU:
*related: #714*

ping @rwaldron @johnnyman727 

This is a quick fix that will stop the t2 cli from crashing for the affected users. It will print a warning and then report the error to our crash server.

screen shot 2016-04-27 at 23 08 50

Another funny thing I noticed is that the crash reporter won't be able to upload crash data if we call `process.exit()` to early, like we do when running only `t2` for example, but that's another issue altogether...

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -23,7 +23,6 @@ const pkg = require('../package.json');
 try {
   updateNotifier({ pkg }).notify();
 } catch (err) {
-  logs.warn('Failed to check for newer version of the CLI');
   CrashReporter.submit(err.stack);
 }
 
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -23,7 +23,7 @@ const pkg = require('../package.json');
 try {
   updateNotifier({ pkg }).notify();
 } catch (err) {
-  CrashReporter.submit(err.stack);
+  CrashReporter.submit(err.stack, { silent: true });
 }
 
 function makeCommand(commandName) {
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -19,9 +19,13 @@ const CLI_ENTRYPOINT = 'cli.entrypoint';
 
 // Check for updates
 const pkg = require('../package.json');
-updateNotifier({
-  pkg
-}).notify();
+
+try {
+  updateNotifier({ pkg }).notify();
+} catch (err) {
+  logs.warn('Failed to check for newer version of the CLI');
+  CrashReporter.submit(err.stack);
+}
 
 function makeCommand(commandName) {
   return parser.command(commandName)
--- a/lib/crash-reporter.js
+++ b/lib/crash-reporter.js
@@ -47,7 +47,11 @@ CrashReporter.on = function() {
     });
 };
 
-CrashReporter.submit = function(report) {
+CrashReporter.submit = function(report, opts) {
+  if (opts === undefined) { opts = {}; }
+
+  const silent = (opts.silent || false);
+
   return Preferences.read(CRASH_REPORTER_PREFERENCE, 'on')
     .then(value => {
       if (value === 'on') {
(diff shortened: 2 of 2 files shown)


bin/tessel-2.js
```diff
@@ -19,9 +19,13 @@ const CLI_ENTRYPOINT = 'cli.entrypoint';
...
 // Check for updates
 const pkg = require('../package.json');
-updateNotifier({
-  pkg
-}).notify();
+
+try {
+  updateNotifier({ pkg }).notify();
+} catch (err) {
+  logs.warn('Failed to check for newer version of the CLI');
```
johnnyman727: I might even be in favor of silently failing with this one because it seems like people who do have the problem will hit this virtually every time they use the CLI.

rwaldron: > in favor of silently failing with this one

Seconded

LinusU: Sounds good!

What we could to would be to only show the error the first time, if only there was some kind of library that made it very easy to store if we have previously shown the error or not... Oh, I know! Nah, just kidding ;)

bin/tessel-2.js
```diff
@@ -19,9 +19,12 @@ const CLI_ENTRYPOINT = 'cli.entrypoint';
...
 // Check for updates
 const pkg = require('../package.json');
-updateNotifier({
-  pkg
-}).notify();
+
+try {
+  updateNotifier({ pkg }).notify();
+} catch (err) {
+  CrashReporter.submit(err.stack);
```
rwaldron: Isn't this going to have the same net outcome?

LinusU: Hmm, do you mean that it will print a warning? Yes that is right, I'll take a look...

johnnyman727: It shouldn't because the error is caught. `CrashReporter.submit` doesn't exit the process.

LinusU: It prints `INFO Crash Reported: http://crash-reporter.tessel.io/crashes?fingerprint=-0x173999f5` which is about as annoying as `WARN Failed to check for newer versions of the CLI`

lib/crash-reporter.js
```diff
@@ -47,7 +47,11 @@ CrashReporter.on = function() {
     });
 };
 
-CrashReporter.submit = function(report) {
+CrashReporter.submit = function(report, opts) {
+  if (opts === undefined) opts = {}
+
+  const silent = (opts.silent || false)
+
```
rwaldron: nits: curly braces for the if-statement's consequent and semi-colons (`grunt` would've caught this :P)

rwaldron: Also, `test/unit/crash-reporter.js` needs tests for this now

LinusU: Whoops, I'll fix it right away!

bin/tessel-2.js
```diff
@@ -19,9 +19,12 @@ const CLI_ENTRYPOINT = 'cli.entrypoint';
...
 // Check for updates
 const pkg = require('../package.json');
-updateNotifier({
-  pkg
-}).notify();
+
+try {
+  updateNotifier({ pkg }).notify();
+} catch (err) {
+  CrashReporter.submit(err.stack, { silent: true });
```
rwaldron: So, now it's quiet for the user, but http://crash-reporter.tessel.io/ is going to continue to amass these reports. I think the real fix is to simply suppress this error :| (or maybe the authors of `configstore` could just make the damn thing work without barfing all over the place

LinusU: I thought it was a good idea to keep collecting the crash data, that way we can see how many users are impacted by this, and also help out with additional information that might be required for yeoman/update-notifier#76...

LinusU:
@rwaldron I need to go to bed now but feel free to make modifications to this branch and land it 👍

bin/tessel-2.js
rwaldron: Ok, I can agree to that then.

rwaldron:
> I need to go to bed now but feel free to make modifications to this branch and land it

Good night!

rwaldron:
Tests added: https://github.com/tessel/t2-cli/commit/43db821f00741c5bf4d0710caf40e8b830e3dbe9

PR #733, opened 2016-05-10
Add custom hostname flag to uci when using `t2 rename`

### LINKED ISSUE #252: t2 update: Sometimes the hostname changes after update
We need to figure out why the hostname changes sometimes after update. Once we know why, we can take steps to set it back to what it was previously.

wyze:
This is the other half of the puzzle for https://github.com/tessel/openwrt-tessel/pull/59 and completes https://github.com/tessel/openwrt-tessel/issues/19.

Also would close #252 and make https://github.com/tessel/t2-start/pull/117 not necessary.

### DIFF
--- a/lib/tessel/commands.js
+++ b/lib/tessel/commands.js
@@ -92,6 +92,9 @@ module.exports.getHostname = function() {
 module.exports.setHostname = function(hostname) {
   return ['uci', 'set', 'system.@system[0].hostname=' + hostname];
 };
+module.exports.setHostnameFlag = function() {
+  return ['uci', 'set', 'system.@system[0].custom_hostname_set=1'];
+};
 module.exports.commitHostname = function() {
   return ['uci', 'commit', 'system'];
 };
--- a/lib/tessel/name.js
+++ b/lib/tessel/name.js
@@ -36,6 +36,10 @@ Tessel.prototype.setName = function(name) {
     // Write the new hostname to the device
     return this.simpleExec(commands.setHostname(name))
       .then(() => {
+        // Set a flag to remember our hostname on upgrade
+        return this.simpleExec(commands.setHostnameFlag());
+      })
+      .then(() => {
         // Commit the new hostname
         return this.simpleExec(commands.commitHostname());
       })
--- a/test/unit/name.js
+++ b/test/unit/name.js
@@ -20,6 +20,7 @@ exports['Tessel.prototype.rename'] = {
     this.commitHostname = this.sandbox.spy(commands, 'commitHostname');
     this.openStdinToFile = this.sandbox.spy(commands, 'openStdinToFile');
     this.setHostname = this.sandbox.spy(commands, 'setHostname');
+    this.setHostnameFlag = this.sandbox.spy(commands, 'setHostnameFlag');
     this.getHostname = this.sandbox.spy(commands, 'getHostname');
     this.logsWarn = this.sandbox.stub(logs, 'warn', function() {});
     this.logsInfo = this.sandbox.stub(logs, 'info', function() {});
@@ -81,7 +82,7 @@ exports['Tessel.prototype.rename'] = {
   },
 
   resetName: function(test) {
-    test.expect(7);
+    test.expect(8);
 
     this.tessel.rename({
         reset: true
@@ -94,6 +95,7 @@ exports['Tessel.prototype.rename'] = {
         test.equal(this._getMACAddress.callCount, 1);
         test.equal(this.setName.callCount, 1);
         test.equal(this.setHostname.callCount, 1);
+        test.equal(this.setHostnameFlag.callCount, 1);
         test.equal(this.commitHostname.callCount, 1);
         test.equal(this.openStdinToFile.callCount, 1);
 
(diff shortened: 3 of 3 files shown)


johnnyman727:
@wyze so glad you're working on this.

Just so I'm sure I understand your proposed solution - you're basically creating a new flag that notes whether the name has been rewritten and then checking for that flag during the update process?

Have you confirmed that the flag is not overwritten during an update? I think it probably would not be overwritten since other `uci` configs are not overwritten but I just wanted to check.

What do you think we should do about the `t2 rename --reset` command? Should we remove this new flag or set it to zero? I personally don't see a need to do that because I think the only case a user might expect a new name is if they ran `t2 update -n` to completely erase all configs.

I might suggest that we name the new flag `custom_hostname_set` because it sounds more like a flag to me.

We may want to consider just writing to a flag whenever a new hostname is made. Like I said, the only case where we want to change the hostname is on `t2 update -n` which would delete that flag anyway (so a new name would be set).

wyze:
# Smoke Test

---

#### Rename

```
❯ t2 rename Tester
INFO Connected to Tessel-02A30510B2C5.
INFO Changed name of device Tessel-02A30510B2C5 to Tester
```

#### Make sure flag is set

```
❯ t2 root
INFO Connected to Tester.
INFO Starting SSH Session on Tessel. Type 'exit' at the prompt to end.


BusyBox v1.23.2 (2016-05-10 21:55:24 EDT) built-in shell (ash)

Tessel 2  /  Built on OpenWrt
root@Tester:~# cat /etc/config/system

config system
	option timezone 'UTC'
	option hostname 'Tester'
	option custom_hostname '1'
```

#### Update

```
❯ t2 update --openwrt-path openwrt-ramips-mt7620-tessel-squashfs-sysupgrade-node-4.4.3.bin
INFO Connected to Tester.
INFO Updating OpenWRT (1/2)
INFO Transferring image of size 19.14 MB. This will take 2-4 minutes...
INFO Transfer complete.
INFO Starting OpenWRT update.
INFO Please do not remove power from Tessel.
INFO This process will take at least two minutes...
INFO Configuration is saved during update.
WARN No firmware binary loaded... skipping firmware update
INFO Finished updating Tessel with local builds.
```

#### List

```
❯ t2 list
	USB	Tester
```

#### `t2 update -n`

```
❯ t2 update --openwrt-path openwrt-ramips-mt7620-tessel-squashfs-sysupgrade-node-4.4.3.bin -n
INFO Connected to Tester.
INFO Updating OpenWRT (1/2)
INFO Transferring image of size 19.14 MB. This will take 2-4 minutes...
INFO Transfer complete.
INFO Starting OpenWRT update.
INFO Please do not remove power from Tessel.
INFO This process will take at least two minutes...
INFO Configuration is not saved during update.
WARN No firmware binary loaded... skipping firmware update
INFO Finished updating Tessel with local builds.
```

#### List

```
❯ t2 list
	USB	Tessel-02A30510B2C5
```

wyze:
> Just so I'm sure I understand your proposed solution - you're basically creating a new flag that notes whether the name has been rewritten and then checking for that flag during the update process?

The CLI will set the flag, and then the uci defaults file, [50-tessel-hostname](https://github.com/tessel/openwrt-tessel/blob/custom-hostname-flag/files/etc/uci-defaults/50-tessel-hostname) performs the check during the update on whether or not to change the hostname.

> Have you confirmed that the flag is not overwritten during an update? I think it probably would not be overwritten since other uci configs are not overwritten but I just wanted to check.

See the smoke test above. 😄  

> What do you think we should do about the `t2 rename --reset` command? Should we remove this new flag or set it to zero? I personally don't see a need to do that because I think the only case a user might expect a new name is if they ran `t2 update -n` to completely erase all configs.

The `t2 rename --reset` command still uses the underlying `setName` function which sets this flag, so even on reset the flag would still be set. My smoke test above confirms that passing `-n` to `t2 update` does reset the hostname.

> I might suggest that we name the new flag custom_hostname_set because it sounds more like a flag to me.

Works for me, I'll get that changed tomorrow.

> We may want to consider just writing to a flag whenever a new hostname is made. Like I said, the only case where we want to change the hostname is on `t2 update -n` which would delete that flag anyway (so a new name would be set).

I think it might already do this, because `t2 rename <newname>` and `t2 rename --reset` use the same `setName` function which sets the flag. I am not aware of any other way that a new hostname is made. Unless you mean by when the `50-tessel-hostname` file gets executed, and it updates the hostname, set the flag there too?

johnnyman727:
> See the smoke test above. 😄

:+1: 

> Unless you mean by when the 50-tessel-hostname file gets executed, and it updates the hostname, set the flag there too?

Yes, this is exactly what I mean. This would solve the one sort of corner case where someone updates their Tessel, never changes the default name but becomes accustomed to it (or writes some code to access THEIR_TESSEL.local or whatever) and then is surprised to find the name updated and code broken after an update. This would just be an OpenWRT patch and not a CLI patch.

wyze:
Okay. I will add that to the `uci-default` file in the PR over there: https://github.com/tessel/openwrt-tessel/pull/59

wyze:
@johnnyman727, Is this good to go? I merged https://github.com/tessel/openwrt-tessel/pull/59 since more smoke tests were successful.

johnnyman727:
@wyze we don't need this anymore, right? OpenWRT will set its own custom_hostname field as soon as it's updated without a `-n` flag.

wyze:
Ah, you might be right. Let me double check.

wyze:
Yes, this is no longer needed.

PR #741, opened 2016-05-16
Remove (unnecessary?) `+` at the beginning of the udev rule

ryanplusplus:
The `+` caused the udev rule to fail to load on Fedora. Removing it allowed it to work. I'm assuming this was just a typo.


johnnyman727:
Yes, based on www.github.com/tessel/t1-cli/blob/master/install/85-Tessel.rules I think this was a copy paste error. 

Thanks for the catch!

Would you mind signing our CLA if you haven't already? www.dojofoundation.org/claForm. Once you do that we can merge.

ryanplusplus:
I'm happy to sign the CLA, but the link you provided 404s.

johnnyman727:
Sorry about that, tried to type it out from memory on a phone. It's actually www.dojofoundation.org/about/claForm

ryanplusplus:
Okay, I've signed the CLA. Thanks!

johnnyman727:
Thanks you!

PR #744, opened 2016-05-18
t2 restore: restore/unbrick openwrt on tessel 2

### LINKED ISSUE #493: RFC: tessel unbrick
I think we should add a command to unbrick a Tessel if either the openwrt image or the firmware image is somehow corrupted. 

A corrupted openWRT image can be easily fixed with a script like [this one](https://github.com/tessel/t2-firmware/blob/master/scripts/flash.py) because @kevinmehall already wrote a utility to write directly to the OpenWRT Flash from the coprocessor. We could either port it to JS or just include that linked Python file directly. We will need to place the u-boot.bin image either on our build server or ship it with the CLI. 

A corrupted coprocessor image can be fixed using `openocd` from the OpenWRT image but this is more involved and you would *need* to have LAN access to OpenWRT. openocd would have to be installed with opkg (or included by default). Then we would use openocd to load an image file. Basic instructions for this [can be found here](https://gist.github.com/johnnyman727/23de244b454fd382e57d).

We can do the same as the above for a corrupted bootloader by substituting the bootloader image.

I suggest the following commands:
```
tessel unbrick --openwrt // Unbricks the linux bootloader and file system
tessel unbrick --firmware // Unbricks coprocessor firmware
tessel unbrick --bootloader // Unbricks coprocess bootloader 
```

majgis:
This fixes #493

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -145,11 +145,6 @@ parser.command('restore')
     log.level(options.loglevel);
     callControllerWith('restoreTessel', options);
   })
-  .option('openwrt', {
-    abbr: 'o',
-    flag: true,
-    help: 'Restore a factory version of OpenWrt'
-  })
   .help('Restore your Tessel by installing the factory version of OpenWrt.');
 
 makeCommand('restart')
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -140,6 +140,15 @@ parser.command('provision')
   })
   .help('Authorize your computer to control the USB-connected Tessel');
 
+parser.command('restore')
+  .callback(callControllerCallback('restoreTessel'))
+  .option('openwrt', {
+    abbr: 'o',
+    flag: true,
+    help: 'Restore a factory version of OpenWrt'
+  })
+  .help('Restore your Tessel by installing the factory version of OpenWrt.');
+
 makeCommand('restart')
   .callback(options => {
     log.level(options.loglevel);
@@ -583,7 +592,7 @@ module.exports.closeFailedCommand = function(status, options) {
   }
 
   log.spinner.stop();
-  process.exit(options.code || status.code || code);
+  process.exit(options.code || (status && status.code) || code);
 };
 
 
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -141,7 +141,10 @@ parser.command('provision')
   .help('Authorize your computer to control the USB-connected Tessel');
 
 parser.command('restore')
-  .callback(callControllerCallback('restoreTessel'))
+  .callback(options => {
+    log.level(options.loglevel);
+    callControllerWith('restoreTessel', options);
+  })
   .option('openwrt', {
     abbr: 'o',
     flag: true,
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -18,7 +18,6 @@ var updates = require('./update-fetch');
 var provision = require('./tessel/provision');
 var Tessel = require('./tessel/tessel');
 
-
 var controller = {};
 var responses = {
   noAuth: 'No Authorized Tessels Found.',
@@ -171,7 +170,8 @@ Tessel.get = function(opts) {
       timeout: (opts.timeout || 2) * 1000,
       usb: opts.usb,
       lan: opts.lan,
-      authorized: true
+      authorized: true,
+      altSetting: opts.altSetting
     };
 
     if (opts.authorized !== undefined) {
@@ -553,6 +553,28 @@ controller.provisionTessel = function(opts) {
         return tessel.provisionTessel(opts);
       });
     });
+};
+
+
+controller.restoreTessel = function(opts) {
+  if (!opts.openwrt) {
+    return Promise.reject('You must use the --openwrt option to restore.');
+  }
+
+  opts.usb = true;
+  opts.lan = false;
+  opts.altSetting = 1;
+  return controller.get(opts).then(function(tessel) {
+    return new Promise(function(resolve, reject) {
+      tessel.restore(opts)
+        .then(function(result) {
+          resolve(result);
+        })
+        .catch(function(err) {
+          reject(err);
+        });
+    });
+  });
 };
 
 controller.deploy = function(opts) {
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -557,10 +557,6 @@ controller.provisionTessel = function(opts) {
 
 
 controller.restoreTessel = function(opts) {
-  if (!opts.openwrt) {
-    return Promise.reject('You must use the --openwrt option to restore.');
-  }
-
   opts.usb = true;
   opts.lan = false;
   opts.altSetting = 1;
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -560,13 +560,13 @@ controller.restoreTessel = function(opts) {
   opts.usb = true;
   opts.lan = false;
   opts.altSetting = 1;
-  return controller.get(opts).then(function(tessel) {
-    return new Promise(function(resolve, reject) {
+  return controller.get(opts).then((tessel) => {
+    return new Promise((resolve, reject) => {
       tessel.restore(opts)
-        .then(function(result) {
+        .then((result) => {
           resolve(result);
         })
-        .catch(function(err) {
+        .catch((err) => {
           reject(err);
         });
     });
--- a/lib/flash.js
+++ b/lib/flash.js
@@ -25,7 +25,7 @@ function factoryPartition(mac1, mac2) {
   );
 }
 
-function transaction(usbConnection, write, read, status_poll, wren, next) {
+function transaction (usbConnection, write, read, status_poll, wren, next) {
   read = read || 0;
   status_poll = status_poll || false;
   wren = wren || false;
@@ -69,8 +69,8 @@ function eraseChip(usbConnection, next) {
 }
 
 // Poll for the WIP bit in the status register to go low
-function waitTransactionComplete(usbConnection, next) {
-  setTimeout(function onWait() {
+function waitTransactionComplete (usbConnection, next) {
+  setTimeout(function onWait () {
     getStatus(usbConnection, (err, status) => {
       if (err) {
         return next(err);
@@ -138,39 +138,43 @@ function randbyte() {
   return randint(0, 255);
 }
 
-function asyncLog(msg, next) {
-  log.info(msg);
-  next();
+function asyncLog(msg) {
+  return function onAsyncLog(next) {
+    log.info(msg);
+    next();
+  }
 }
 
-module.exports = function flashDevice(usbConnection, ubootBuffer, squashfsBuffer, next) {
-  var uid = [randbyte(), randbyte(), randbyte(), randbyte()];
-  var mac1 = [0x02, 0xa3].concat(uid);
-  var mac2 = [0x02, 0xa4].concat(uid);
+module.exports = function flashDevice(usbConnection, ubootBuffer, squashfsBuffer) {
+  return new Promise((resolve, reject)=> {
+    var uid = [randbyte(), randbyte(), randbyte(), randbyte()];
+    var mac1 = [0x02, 0xa3].concat(uid);
+    var mac2 = [0x02, 0xa4].concat(uid);
 
-  async.waterfall([
-    asyncLog.bind(null, 'Checking the chip id...'),
-    readChipId.bind(null, usbConnection),
-    checkChipId,
+    async.waterfall([
+      asyncLog('Checking the chip id...'),
+      readChipId.bind(null, usbConnection),
+      checkChipId,
 
-    asyncLog.bind(null, 'Erasing the chip...'),
-    setWriteEnabled.bind(null, usbConnection),
-    eraseChip.bind(null, usbConnection),
-    waitTransactionComplete.bind(null, usbConnection),
+      asyncLog('Erasing the chip...'),
+      setWriteEnabled.bind(null, usbConnection),
+      eraseChip.bind(null, usbConnection),
+      waitTransactionComplete.bind(null, usbConnection),
 
-    asyncLog.bind(null, 'Writing uboot...'),
-    write.bind(null, usbConnection, 0, ubootBuffer),
-    waitTransactionComplete.bind(null, usbConnection),
+      asyncLog('Writing uboot...'),
+      write.bind(null, usbConnection, 0, ubootBuffer),
+      waitTransactionComplete.bind(null, usbConnection),
 
-    asyncLog.bind(null, 'Writing mediatek factory partition...'),
-    write.bind(null, usbConnection, 0x40000, factoryPartition(mac1, mac2)),
-    waitTransactionComplete.bind(null, usbConnection),
+      asyncLog('Writing mediatek factory partition...'),
+      write.bind(null, usbConnection, 0x40000, factoryPartition(mac1, mac2)),
+      waitTransactionComplete.bind(null, usbConnection),
 
-    asyncLog.bind(null, 'Writing squashfs...'),
-    write.bind(null, usbConnection, 0x50000, squashfsBuffer),
-    waitTransactionComplete.bind(null, usbConnection),
-    asyncLog.bind(null, 'The update was successful.'),
-    asyncLog.bind(null, 'Please power cycle your Tessel.')
+      asyncLog('Writing squashfs...'),
+      write.bind(null, usbConnection, 0x50000, squashfsBuffer),
+      waitTransactionComplete.bind(null, usbConnection),
+      asyncLog('The update was successful.'),
+      asyncLog('Please power cycle your Tessel.')
 
-  ], next);
+    ], (err) => err ? reject(err) : resolve());
+  })
 };
--- a/lib/flash.js
+++ b/lib/flash.js
@@ -1,6 +1,14 @@
+// System Objects
+// ...
+
+// Third Party Dependencies
 var async = require('async');
-var PAGE = 256;
+
+// Internal
 var log = require('./log');
+
+// Constants
+var PAGE = 256;
 var CHIP_ID = new Buffer([0x01, 0x02, 0x19]);
 
 function address(addr) {
--- a/lib/flash.js
+++ b/lib/flash.js
@@ -38,7 +38,7 @@ function transaction(usbConnection, write, read, status_poll, wren, next) {
 
   var data = concatToBuffer(hdr, write);
 
-  usbConnection.epOut.transfer(data, function(err) {
+  usbConnection.epOut.transfer(data, (err) => {
     if (err) {
       return next(err);
     }
@@ -71,7 +71,7 @@ function eraseChip(usbConnection, next) {
 // Poll for the WIP bit in the status register to go low
 function waitTransactionComplete(usbConnection, next) {
   setTimeout(function onWait() {
-    getStatus(usbConnection, function onGetStatus(err, status) {
+    getStatus(usbConnection, (err, status) => {
       if (err) {
         return next(err);
       }
@@ -85,7 +85,7 @@ function waitTransactionComplete(usbConnection, next) {
 
 // Read the status register
 function getStatus(usbConnection, next) {
-  transaction(usbConnection, [0x05], 1, null, null, function onGetStatus(err, data) {
+  transaction(usbConnection, [0x05], 1, null, null,(err, data) => {
     next(err, data ? data[0] : data);
   });
 }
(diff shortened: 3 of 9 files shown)


majgis:
The unbrick command is shown in the help:
```
$ t2 --help

Usage: t2 <command>

command     
  crash-reporter      Configure the Crash Reporter.
  unbrick             Unbrick your Tessel by restoring to factory settings.
 ...
```

The `t2 unbrick` command prompts the user to use the `--openwrt` option.  There was an error here that most likely came after the last rebase.  The second commit fixed this error.
```
$ t2 unbrick
INFO You must use the --openwrt option to proceed with unbricking.
```

Unbricking:
```
$ t2 unbrick --openwrt
INFO Connected to your Tessel.
INFO Proceeding with updating OpenWrt...
INFO Beginning download. This could take a couple minutes..
  Downloading [====================] 100% 0.0s remaining
INFO Download complete!
INFO Checking the chip id...
INFO Erasing the chip...
INFO Writing uboot...
INFO Writing mediatek factory partition...
INFO Writing squashfs...
INFO Please power cycle your Tessel

```

johnnyman727:
@majgis this looks great. My only suggestion si that we add one more `logs.info` that mentions that the update was successful preceeding the "Please power cycle your Tessel" message. Otherwise, it may look like it failed.

@rwaldron can you review once that's done? I will be out of internet range for the next 10 days.

majgis:
@rwaldron The success message was added, let me know if there is anything else.

lib/controller.js
```diff
@@ -557,6 +557,29 @@ controller.provisionTessel = function(opts) {
     });
 };
 
+
+controller.unbrickTessel = function(opts) {
+  if (!opts.openwrt) {
+    logs.info('You must use the --openwrt option to proceed with unbricking.');
+    return Promise.reject();
```
johnnyman727: In order to be consistent with existing error handing, I would recommend calling `Promise.reject` with the info string you are printing out just before. So replace the above two lines with:
```
return Promise.reject('You must use the --openwrt option to proceed with unbricking.');
```
Because a string is returned instead of an `Error`, the CLI will print out a WARN instead of ERR. I'm not particularly pleased with the way the CLI has different error handling based on return type but we can clean that up with one comprehensive PR in the future.

lib/controller.js
```diff
@@ -1020,4 +1043,5 @@ module.exports = controller;
 
 // Shared exports
 module.exports.listTessels = controller.list;
+module.exports.listTessels = controller.list;
```
johnnyman727: It looks like this line was duplicated?

lib/unbrick.js
```diff
@@ -0,0 +1,27 @@
+var update = require('./update-fetch');
+var logs = require('./logs');
+var flash = require('./flash');
+
+module.exports = function unbrick(opts, tessel) {
```
johnnyman727: For the sake of consistency with other commands, I recommend making the `unbrick` command a method on the `Tessel` prototype. We have been doing this for all commands that actually manipulate the Tessel. See [the version command](https://github.com/tessel/t2-cli/blob/master/lib/tessel/version.js) as an example. If you do this, please move this file to `lib/tessel/unbrick.js`.

lib/unbrick.js
```diff
@@ -0,0 +1,27 @@
...
+      return update.fetchUnbrick()
+        .then(function(result) {
+          flash(tessel.connection, result.uboot, result.squashfs, function onFlash(err) {
+            if (err) {
+              return reject(err);
+            }
+            resolve();
+          });
+        })
+        .catch((e) => {
```
johnnyman727: Are you sure this `catch` is needed if you are `return`ing the promise returned by `fetchUnbrick`?

lib/controller.js
```diff
@@ -91,7 +91,6 @@ Tessel.list = function(opts) {
 
     // When a Tessel is found
     seeker.on('tessel', function displayResults(tessel) {
-
```
johnnyman727: We generally try to avoid whitespace changes like this to keep the history cleaner.

johnnyman727:
@majgis sorry for the delay! Back from the wilderness and had a chance to review. For the most part it looks great! Ping me once you've had a chance to look over my feedback.

bin/tessel-2.js
```diff
@@ -144,6 +144,15 @@ parser.command('provision')
...
 
+parser.command('unbrick')
+  .callback(callControllerCallback('unbrickTessel'))
+  .option('openwrt', {
+    abbr: 'o',
+    flag: true,
+    help: 'Restore a factory version of OpenWrt'
+  })
+  .help('Unbrick your Tessel by restoring to factory settings.');
+
```
rwaldron: Any opposition to along this "restore"?

bin/tessel-2.js
```diff
@@ -492,7 +501,7 @@ module.exports.closeFailedCommand = function(status, options) {
     }
   }
 
-  process.exit(options.code || status.code || code);
+  process.exit( options.code || (status && status.code) || code);
```
rwaldron: Please run `grunt`

majgis:
I will address these issues this evening.

bin/tessel-2.js
majgis: @johnnyman727 let me know if you would like to rename all occurrences of unbrick to restore.

majgis:
All the issues should be addressed.  Let me know if you want the rename of unbrick to restore too.

rwaldron:
I'd prefer if this was `restore`, please ping me when updated—thanks!!

majgis:
@rwaldron I changed unbrick to restore.  Let me know if there is anything else.

rwaldron:
I will put this on my list for tomorrow :)

johnnyman727:
ping @rwaldron

rwaldron:
Whoops, sorry I missed that ping. I will add this to my list for today.

lib/flash.js
```diff
@@ -0,0 +1,184 @@
+var async = require('async');
+var PAGE = 256;
+var logs = require('./logs');
+var CHIP_ID = new Buffer([0x01, 0x02, 0x19]);
```
rwaldron: Take a look at the deps section in other files—they are split into categories and separate from other var decls. I'm not going to block this patch on that, as I can just do a fix up after it lands, but good to know for future reference :)

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+        return next();
+      }
+      process.nextTick(waitTransactionComplete, usbConnection, next);
+    });
+  }, 200);
+}
+
+// Read the status register
+function getStatus(usbConnection, next) {
+  transaction(usbConnection, [0x05], 1, null, null, function onGetStatus(err, data) {
```
rwaldron: Use arrow function

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+}
+
+function eraseChip(usbConnection, next) {
+  transaction(usbConnection, [0x60], null, null, null, next);
+}
+
+// Poll for the WIP bit in the status register to go low
+function waitTransactionComplete(usbConnection, next) {
+  setTimeout(function onWait() {
+    getStatus(usbConnection, function onGetStatus(err, status) {
```
rwaldron: Use arrow function

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+
+  var byteIndex = 0;
+  for (i = 0; i < buffers.length; i++) {
+    buffer = buffers[i];
+    buffer.copy(result, byteIndex);
+    byteIndex += buffer.length;
+  }
+
+  return result;
+}
```
rwaldron: This can be reduced to eliminate all of the extra buffer allocations: 

```js
function toBuffer() {
  return new Buffer(
    Array.from(arguments).reduce(
      (accum, entry) => accum.concat(entry)
    )
  );
}


var result = toBuffer(
  [0x00, 0x01, 0x02, 0x03],
  [0x04],
  [0x05],
  [0x06, 0x07, 0x08, 0x09],
  [0x0A, 0x0B, 0x0C, 0x0D],
  [0x0E],
  [0x0F]
);

console.log(result);



// <Buffer 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f>

```

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+  return randint(0, 255);
+}
+
+function asyncLog(msg, next) {
+  logs.info(msg);
+  next();
+}
+
+module.exports = function flashDevice(usbConnection, ubootBuffer, squashfsBuffer, next) {
+  var uid = [randbyte(), randbyte(), randbyte(), randbyte()];
```
rwaldron: Use: `Array.from(crypto.randomBytes(4));`

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+  var mac1 = [0x02, 0xa3].concat(uid);
+  var mac2 = [0x02, 0xa4].concat(uid);
+
+  async.waterfall([
```
rwaldron: I hate to drop this mega bomb on you, but you'll need to convert all of this to a Promise based operation. The rest of the CLI uses Promises throughout. Let me know if you have any questions about the process of coversion

lib/controller.js
```diff
@@ -557,6 +557,28 @@ controller.provisionTessel = function(opts) {
...
+  opts.usb = true;
+  opts.lan = false;
+  opts.altSetting = 1;
+  return controller.get(opts).then(function(tessel) {
+    return new Promise(function(resolve, reject) {
+      tessel.restore(opts)
+        .then(function(result) {
+          resolve(result);
+        })
+        .catch(function(err) {
```
rwaldron: Use arrow function

lib/controller.js
```diff
@@ -557,6 +557,28 @@ controller.provisionTessel = function(opts) {
...
+    return Promise.reject('You must use the --openwrt option to restore.');
+  }
+
+  opts.usb = true;
+  opts.lan = false;
+  opts.altSetting = 1;
+  return controller.get(opts).then(function(tessel) {
+    return new Promise(function(resolve, reject) {
+      tessel.restore(opts)
+        .then(function(result) {
```
rwaldron: Use arrow function

lib/controller.js
```diff
@@ -557,6 +557,28 @@ controller.provisionTessel = function(opts) {
...
+
+controller.restoreTessel = function(opts) {
+  if (!opts.openwrt) {
+    return Promise.reject('You must use the --openwrt option to restore.');
+  }
+
+  opts.usb = true;
+  opts.lan = false;
+  opts.altSetting = 1;
+  return controller.get(opts).then(function(tessel) {
```
rwaldron: Use arrow function

lib/controller.js
```diff
@@ -557,6 +557,28 @@ controller.provisionTessel = function(opts) {
...
+    return new Promise(function(resolve, reject) {
```
rwaldron: Use arrow function

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+    getStatus(usbConnection, function onGetStatus(err, status) {
+      if (err) {
+        return next(err);
+      }
+      if (status === 0) {
+        return next();
+      }
+      process.nextTick(waitTransactionComplete, usbConnection, next);
+    });
+  }, 200);
```
rwaldron: Where does this number come from?

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+
+module.exports = function flashDevice(usbConnection, ubootBuffer, squashfsBuffer, next) {
+  var uid = [randbyte(), randbyte(), randbyte(), randbyte()];
+  var mac1 = [0x02, 0xa3].concat(uid);
+  var mac2 = [0x02, 0xa4].concat(uid);
+
+  async.waterfall([
+    asyncLog.bind(null, 'Checking the chip id...'),
+    readChipId.bind(null, usbConnection),
+    checkChipId,
```
rwaldron: This information is actually how the CLI determines which USB peripheral is a Tessel, so I believe it's unnecessary, since the operation wouldn't have reached this point if the id was incorrect. Note that this value is called "TESSEL_VID" (0x1209) elsewhere.

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+  return concatToBuffer(
+    [0x20, 0x76, 0x03, 0x01],
+    mac1,
+    Array(30).fill(0xff),
+    mac2
+  );
+}
+
+function transaction(usbConnection, write, read, status_poll, wren, next) {
+  read = read || 0;
```
rwaldron: The  argument value of this parameter is often `null`—it should just be a number all the time

lib/flash.js
rwaldron: For now, the good news is that I think this can actually be done fairly easily. This section roughly changes to: 


```
return new Promise(resolve => {
  async.waterfall([
    asyncLog.bind(null, 'Erasing the chip...'),
    setWriteEnabled.bind(null, usbConnection),
    eraseChip.bind(null, usbConnection),
    waitTransactionComplete.bind(null, usbConnection),

    asyncLog.bind(null, 'Writing uboot...'),
    write.bind(null, usbConnection, 0, ubootBuffer),
    waitTransactionComplete.bind(null, usbConnection),

    asyncLog.bind(null, 'Writing mediatek factory partition...'),
    write.bind(null, usbConnection, 0x40000, factoryPartition(mac1, mac2)),
    waitTransactionComplete.bind(null, usbConnection),

    asyncLog.bind(null, 'Writing squashfs...'),
    write.bind(null, usbConnection, 0x50000, squashfsBuffer),
    waitTransactionComplete.bind(null, usbConnection),
    asyncLog.bind(null, 'The update was successful.'),
    asyncLog.bind(null, 'Please power cycle your Tessel.')
  ], resolve);  
}); 
```

The bad news is that once this change set lands, we need to refactor all the rest of the operations to be promises as well and get rid of `async`. I've spend some time on my end working out how we can simplify this stuff and clean it up a bit, but I don't think that work should block landing this. (the other stuff I've written does need to be fixed/updated)

lib/flash.js
majgis: https://github.com/tessel/t2-firmware/blob/master/scripts/flash.py#L81  The original value was 100ms.  Shall I change it to that?

lib/flash.js
majgis: I'll go ahead and convert everything to promises.

[Async was already in use](https://github.com/tessel/t2-cli/search?utf8=%E2%9C%93&q=async), you might consider removing it from the project altogether to reduce temptation.

majgis:
I'll get to these changes as soon as I can, latest it will be ready Mon Jul 18.

lib/flash.js
rwaldron: Yeah, I'd definitely like that, but time :\

Of those three uses, there is actually no real point in async being used for any of them, but sometimes "if it ain't broke..."

lib/flash.js
rwaldron: Well, then my question would be: where does _that_ value come from? I'm just curious to learn the origin of this time frame. I'm going to look at the datasheet and see if there's anything about it

bin/tessel-2.js
```diff
@@ -144,6 +144,15 @@ parser.command('provision')
...
   .help('Authorize your computer to control the USB-connected Tessel');
 
+parser.command('restore')
+  .callback(callControllerCallback('restoreTessel'))
+  .option('openwrt', {
+    abbr: 'o',
+    flag: true,
+    help: 'Restore a factory version of OpenWrt'
+  })
+  .help('Restore your Tessel by installing the factory version of OpenWrt.');
```
HipsterBrown: Why do we need the `openwrt` option on this command if there are no other options? From a user perspective, it looks like an unnecessary extra step.

lib/controller.js
HipsterBrown: Can this be changed to just return `tessel.restore(opts)` since that is a Promise?

lib/usb-connection.js
```diff
@@ -144,13 +147,22 @@ USB.Connection.prototype.open = function() {
...
+            resolve();
           }
         });
       });
-    })
-    .then((data) => {
-      this.serialNumber = data;
+    });
+
+  if (altSetting === 1) {
```
HipsterBrown: I don't understand why this logic is done outside the Promise? `this.setAltSetting(altSetting)` could be returned instead of saved to a variable, and this logic could be done inside the next `.then()` statement:
 
from line 156
```js
...
.then(() => {
    if (altSetting === 1) {
        return Promise.resolve(this);
    } else {
        // everything else past line 167
        ...
    }
});
```

This should reduce some of the complexity of passing around a Promise, as well as ensure the error case is handled in one place.

lib/flash.js
```diff
@@ -0,0 +1,184 @@
...
+  });
+}
+
+function readChipId(usbConnection, next) {
+  transaction(usbConnection, [0x9f], 3, null, null, next);
+}
+
+function checkChipId(chipId, next) {
+  if (Buffer.compare(chipId, CHIP_ID) !== 0) {
+    return next(new Error('Invalid chip ID (flash communication error)'));
```
HipsterBrown: - In what case would this error get thrown? Why would the `chipId` not match?
- What does this mean to someone using this command on their Tessel? 
- Is there a way to resolve it?

lib/tessel/restore.js
```diff
@@ -0,0 +1,25 @@
...
+        .then((result) => {
+          flash(usbConnection, result.uboot, result.squashfs, (err) => {
+            if (err) {
+              return reject(err);
+            }
+            resolve();
+          });
+        });
+    }
+    return reject(new Error('You need to specify --openwrt to proceed.'));
```
HipsterBrown: I believe this is already handled at the `controller` level, right?

lib/update-fetch.js
```diff
@@ -95,58 +99,68 @@ function loadLocalBinary(path) {
...
     extract.once('finish', function() {
-      if (!binaries.firmware.length || !binaries.openwrt.length) {
-        return reject(new Error('Fetched binary wasn\'t formatted properly.'));
-      } else {
-        logs.info('Download complete!');
-        return resolve(binaries);
+      for (var key in files) {
+        var file = files[key];
+        if (!file.length) {
+          return reject(new Error('Fetched file wasn\'t formatted properly.'));
```
HipsterBrown: I know this was an existing error message, but can we improve it in this build? Can this be resolved by someone using the CLI or does someone need to push a reformatted image to S3?

johnnyman727: @HipsterBrown someone with AWS permissions would need to push a properly formatted build up. Perhaps we can just suggest they create an issue so we're aware there's a problem?

HipsterBrown: That would be a great update. Basically letting the user know it's not their fault and how the problem can be resolved will make a more constructive error message.

rwaldron:
I've just attempted to rebase this and was successful, but it required a delicate, commit-by-commit resolution.

rwaldron:
After rebasing...

- I had to update occurrences of "logs" to "log"
- "Erasing chip" takes about a minute, the logged message should say that


I tried using this to restore a much older Tessel 2 (TM-T2-03) ([this one](https://github.com/tessel/t2-firmware/issues/122)), but was unsuccessful. I suppose I'll have to risk one of my presently functional T2's to smoke test this when ready :(

rwaldron:
Here's a commit containing the fixes I made after rebasing: https://github.com/rwaldron/t2-cli/commit/2e0de76d1329cedab5892423d9657228e7e94ee6

johnnyman727:
@rwaldron my guess is that the old T2 had firmware pre-0.0.6 which has USB comm incompatibility with more recent CLI releases.

majgis:
Thanks.  I had a busy weekend and I will try to tackle this soon

rwaldron:
@johnnyman727 

> my guess is that the old T2 had firmware pre-0.0.6 which has USB comm incompatibility with more recent CLI releases.

Yeah, that's where I landed as well. It was a long shot ;)

lib/flash.js
majgis: In addition to arrays, there are also buffers coming into this function (from `  var pageData = data.slice(sliceStart, sliceEnd);` in the `write` function) so I went with this:

```
function toBuffer() {
  return Buffer.concat( Array.from(arguments).map (
    (arg) => Array.isArray(arg) ? new Buffer(arg) : arg
  ));
}
```

majgis:
The interface for flash.js now uses a promise, but I did not convert the internals to promises.
I won't be doing any more revisions to this pull request.  Feel free to rebase it to your project if you want to continue to fiddle, or you can close it and it can serve as a reference.  Thank you for your time in reviewing the changes.

rwaldron:
Nice work! I will try to dig in and do a full review this week :)

rwaldron:
I was present when @reconbot tested this branch. My goal is to land this for the next release—thanks for all the patience <3

rwaldron:
I'm currently reviewing and writing tests for this

johnnyman727:
🎊  Nice job @majgis and @rwaldron! I'm glad to see this landed.

Issue #745, opened 2016-05-21
Automatic crash submission vs privacy


flaki:
This was raised (legitimately) as a issue at today's workshop.

We should have an opt-in for this (or something) during CLI install, alternatively we could ask _before_ submitting each crash.

It would be also nice to log the data to the console that's going to be submitted in the crash report.

tikurahul:
We have an opt out which is exposed via the CLI. We should make that option more prominent. 

We also log the unsanitized stack that we are about to submit as long as you are using a recent version of the CLI, as well as the URL to the crash report (which should tell you exactly what we collected). 

I agree that the first time we submit a crash report, we might want to prompt them and tell them about what is going to happen. I will take that as an action item. Do you think that is sufficient?

--

Making the crash reporter opt-in would greatly reduce its usefulness. We have discovered many problems proactively.

johnnyman727:
Fixed with #749.

PR #749, opened 2016-05-30
Add a prompt before you submit a crash report.

tikurahul:
Addresses: https://github.com/tessel/t2-cli/issues/745

After this change:

```
➜  CLI git:(master) t2 crash-reporter --test
ERR! Detected CLI crash [Error: Testing the crash reporter] Error: Testing the crash reporter
    at CrashReporter.off.CrashReporter.on.CrashReporter.sanitize.CrashReporter.sanitize.redactions.CrashReporter.prompt.CrashReporter.submit.CrashReporter.prompt.then.CrashReporter.post.CrashReporter.status.CrashReporter.test (/Users/rahulrav/Workspaces/Tessel2/CLI/lib/crash-reporter.js:177:25)
    at process._tickCallback (node.js:406:9)
    at Function.Module.runMain (module.js:449:11)
    at startup (node.js:141:18)
    at node.js:933:3
INFO Detected a crash in the CLI. Submit crash (y/n) ?
   If yes(y), subsequent crashes will be submitted automatically.
n
INFO Did not submit crash report.
```

### DIFF
--- a/test/unit/menu.js
+++ b/test/unit/menu.js
@@ -1,6 +1,7 @@
 // Test dependencies are required and exposed in common/bootstrap.js
+/*global Menu */
 
-exports['controller.menu'] = {
+exports['Menu.prompt'] = {
   setUp: function(done) {
     this.sandbox = sinon.sandbox.create();
     this.logsWarn = this.sandbox.stub(logs, 'warn', function() {});
@@ -35,7 +36,7 @@ exports['controller.menu'] = {
       });
     });
 
-    controller.menu({
+    Menu.prompt({
       prompt: {
         name: 'selected',
         type: 'list',
@@ -78,7 +79,7 @@ exports['controller.menu'] = {
       });
     });
 
-    controller.menu({
+    Menu.prompt({
       prompt: {
         name: 'selected',
         type: 'list',
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -5,12 +5,12 @@ var util = require('util');
 // Third Party Dependencies
 var async = require('async');
 var colors = require('colors');
-var inquirer = require('inquirer');
 var semver = require('semver');
 
 // Internal
 var discover = require('./discover');
 var logs = require('./logs');
+var Menu = require('./menu');
 var updates = require('./update-fetch');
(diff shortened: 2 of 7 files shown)


rwaldron:
User input things should be uniformly implemented with `controller.menu` (which is a wrapper for `inquirer`).

- [controller.menu](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L995-L1017)
- [example T2 selection](https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L219-L249)

Basically this: 

```js
controller.menu({
  prompt: {
    name: 'selected',
    type: 'confirm',
    message: CRASH_PROMPT_MESSAGE,
  }
}).then(selection => {
    console.log(selection);
});
```


But you'll probably have to put `controller.menu` in a different module now, let's call it `menu.js` and then update both `lib/controller.js` and `lib/crash-reporter.js`

johnnyman727:
@tikurahul I think the output could have more info about what submitting a crash report actually means. Why would the user want to do it? When would they not?

tikurahul:
Yes, I agree. I can add a blurb about how crash reports help us improve the quality of the CLI. 

Do you have any suggestions on what the blurb should state ?

johnnyman727:
Perhaps "Submit Crash Report to help Tessel Developers improve the CLI?"

Also, it looks like we print out "Detected a Crash in the CLI" twice - we could probably strip that from the submission prompt.

tikurahul:
Sounds good. Will clean up the logs and add the blurb.

tikurahul:
Interestingly enough, `sinon.spy` is not able to deal with modules which are pure exported functions. 

For e.g.

```javascript
var Module = (input) => { ... };
module.exports = Module;
```
In tests, `sinon.spy(Module, function() { });` causes unresolved symbol issues.

test/unit/menu.js
```diff
@@ -78,7 +79,7 @@ exports['controller.menu'] = {
       });
     });
 
-    controller.menu({
+    Menu.prompt({
       prompt: {
         name: 'selected',
         type: 'list',
```
rwaldron: Very clean.

PR #761, opened 2016-06-03
Only access point or wifi is enabled at one time

### LINKED ISSUE #755: `t2 ap` isn't creating a network my computer can find
I've accomplished `t2 ap` success in the past on my machine, but this time it's not working.

I have:
```
➜  t2 ap on
INFO Connected to daniel.
INFO SSID: TesselRouter7
INFO Security: none
INFO IP Address: 192.168.1.101
INFO State: Enabled
```

If I dterm in I can read:

```
root@daniel:/# cat etc/config/wireless

config wifi-device 'radio0'
	option type 'mac80211'
	option channel '11'
	option hwmode '11g'
	option path '10180000.wmac'
	option htmode 'HT20'
	option disabled '0'

config wifi-iface
	option device 'radio0'
	option network 'wifi'
	option mode 'sta'
	option ssid 'OSCON'
	option encryption 'none'
	option disabled '0'

config wifi-iface
	option device 'radio0'
	option network 'lan'
	option mode 'ap'
	option encryption 'none'
	option disabled '0'
	option ssid 'TesselRouter7'

config wifi-iface

config wifi-iface

config wifi-iface

config wifi-iface

config wifi-iface

config wifi-iface

config wifi-iface

config wifi-iface
```

(that many-line print of `config wifi-iface` looks like a problem...)

I'm not familiar with how to debug this piece, can someone please help me?

johnnyman727:
Fixes #755.

* Ensures that extra `wifi-iface` lines aren't generated.

@Frijol found that you can't have both wifi and access points enabled at the same time or else the access point doesn't show up.

### DIFF
--- a/test/unit/access-point.js
+++ b/test/unit/access-point.js
@@ -16,7 +16,7 @@ exports['Tessel.prototype.createAccessPoint'] = {
     this.setLanNetworkNetmask = this.sandbox.spy(commands, 'setLanNetworkNetmask');
     this.commitNetwork = this.sandbox.spy(commands, 'commitNetwork');
     this.setAccessPoint = this.sandbox.spy(commands, 'setAccessPoint');
-    this.getAccessPointSSID = this.sandbox.spy(commands, 'getAccessPointSSID');
+    this.getAccessPoint = this.sandbox.spy(commands, 'getAccessPoint');
     this.setAccessPointDevice = this.sandbox.spy(commands, 'setAccessPointDevice');
     this.setAccessPointNetwork = this.sandbox.spy(commands, 'setAccessPointNetwork');
     this.setAccessPointMode = this.sandbox.spy(commands, 'setAccessPointMode');
@@ -29,6 +29,12 @@ exports['Tessel.prototype.createAccessPoint'] = {
     this.reconnectDhcp = this.sandbox.spy(commands, 'reconnectDhcp');
 
     this.tessel = TesselSimulator();
+    // These are needed because the sheer number of commands run within
+    // each test exceeds the default maximum number of listeners. This
+    // is only an issue with tests because we re-use the same remote
+    // process simulator across all command calls
+    this.tessel._rps.stdout.setMaxListeners(100);
+    this.tessel._rps.stderr.setMaxListeners(100);
 
     done();
   },
@@ -47,7 +53,7 @@ exports['Tessel.prototype.createAccessPoint'] = {
       security: 'psk2'
     };
 
-    // Test is expecting two closes...;
+    // Immediately close any opened connections
     this.tessel._rps.on('control', () => {
       setImmediate(() => {
         this.tessel._rps.emit('close');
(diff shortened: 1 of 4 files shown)


johnnyman727:
@Frijol want to give this a try?

johnnyman727:
Don't listen to AppVeyer. It's trying to clone an old branch.

HipsterBrown:
@johnnyman727 

> Ensures that extra wifi-iface lines aren't generated.

Awesome fix! I'm glad that shouldn't be an issue after this PR lands. 

>Disables wifi when an access point is enabled
Disabled access points when wifi is enabled

I don't think this needs to be a constraint on the Tessel. I've successfully run both wifi and an access point connections on the Tessel. Having both enabled means Tessel can connect to wifi configured by a user through a web interface served by the AP. Plus, while a user may not access the external internet through Tessel (yet), the Tessel can connect to outside services while serving a web app through an AP. 

If we just fix the extra `wifi-iface`, does the contraint around wifi and ap both being enabled go away?

johnnyman727:
> I don't think this needs to be a constraint on the Tessel. I've successfully run both wifi and an access point connections on the Tessel.

Now that I'm on a separate network, I see that you're right. *sigh*.

Previously, I had Tessel on a paywalled network so it didn't actually have access to the internet. I wonder if that had some impact. I could consistently connect to wifi, then create an AP and have the AP undiscoverable.

johnnyman727:
@Frijol and I are running a workshop right now and several people have had issues where they connected to wifi then made an AP and the AP didn't show up in discovered networks until they ran `t2 wifi --off`.

HipsterBrown:
@johnnyman727 For the workshop, was there anything unique about the wifi conditions? Was the wifi-blocking issue immediately reproducible on other networks, besides the paywall and workshop networks?

rwaldron:
What's the status?

rwaldron:
This needs to be rebased to include an updated/corrected git history

johnnyman727:
@HipsterBrown I've heard from a couple other people anecdotally that having an AP enabled was blocking the wifi access but I'm not able to reproduce reliably. :( I'm not sure what the best path forward is.

HipsterBrown:
@johnnyman727 I would pull out the fix for the extra `wifi-iface` lines being generated into its own PR so we can get that pushed out as soon as possible. Then I would like to give the other issue of having an AP enabled was blocking the wifi access isolated for more research, but we can add a note around the AP/Wifi docs to help with troubleshooting for folks who run into that issue.

johnnyman727:
@HipsterBrown sorry about the long delay - updated to just fix the multiple ifaces issue.

johnnyman727:
ping @HipsterBrown

HipsterBrown:
Ok great. LGTM :+1: I updated the description for this PR for future reference.

test/unit/access-point.js
```diff
@@ -16,7 +16,7 @@ exports['Tessel.prototype.createAccessPoint'] = {
     this.setLanNetworkNetmask = this.sandbox.spy(commands, 'setLanNetworkNetmask');
     this.commitNetwork = this.sandbox.spy(commands, 'commitNetwork');
     this.setAccessPoint = this.sandbox.spy(commands, 'setAccessPoint');
-    this.getAccessPointSSID = this.sandbox.spy(commands, 'getAccessPointSSID');
+    this.getAccessPoint = this.sandbox.spy(commands, 'getAccessPoint');
```
rwaldron: Cleeeean <3

Issue #607, opened 2016-02-26
[dep] Investigate using npmlog instead of rolling own


rwaldron:
Just an exercise to see what I can learn, maybe it's good for something, maybe not. https://github.com/npm/npmlog

johnnyman727:
:+1:

PR #768, opened 2016-06-12
Logging Reform. Integrate npmlog, char-spinner. Resolves gh-607

rwaldron:
This replaces our logging with npmlog and introduces a standardized integration of "char-spinner" which is used to indicate "something is happening".

This reformed logging system will allow making all logging levels configurable, ie. any type of logging can be enabled/disabled. Note that this capability is not yet exposed to the CLI directly. A few questions need to be answered: 

- [ ] Do we persist logging flags in preferences?
- [ ] If yes, _how_ are the logging flags/preferences exposed? I think we need to have `t2 config` (and it should match [npm config](https://docs.npmjs.com/cli/config)

Demos:

- With and without: `--loglevel=debug`
  
- https://www.youtube.com/watch?v=2dj0ylLgAf8
- https://www.youtube.com/watch?v=nz6x_Gr8eKA
- https://www.youtube.com/watch?v=nz6x_Gr8eKA

Notes: 

- npmlog is presently using a branch on my fork, pending acceptance of https://github.com/npm/npmlog/pull/34
- There may be more places that the spinner can be added, so please feel free to suggest
- There are currently no `http` or `trace` logs, but now that this is in place, I will spend some time implementing those.


bin/tessel-2.js
```diff
@@ -77,42 +79,50 @@ function makeCommand(commandName) {
...
   return controller[methodName](opts)
     .then(module.exports.closeSuccessfulCommand, module.exports.closeFailedCommand);
 }
 
-function callControllerCallback(methodName) {
-  return function(opts) {
-    return callControllerWith(methodName, opts);
-  };
-}
-
```
rwaldron: No longer used. All calls consolidated to `callControllerWith`

bin/tessel-2.js
```diff
@@ -77,42 +79,50 @@ function makeCommand(commandName) {
...
 parser.command('install-drivers')
-  .callback(function() {
+  .callback(options => {
+    if (options.loglevel !== log.level()) {
+      log.level(options.loglevel);
+    }
```
rwaldron: This is implemented for all `callback(...)`. Whatever args handling we replace nom-nom with should include a "global" callback.

bin/tessel-2.js
```diff
@@ -77,42 +79,50 @@ function makeCommand(commandName) {
...
       cr = CrashReporter.on();
-    } else if (opts.off) {
+    } else if (options.off) {
       // t2 crash-reporter --off
       cr = CrashReporter.off();
     }
 
     // t2 crash-reporter --test
-    if (opts.test) {
+    if (options.test) {
```
rwaldron: I renamed `opts` => `options` everywhere.

rwaldron:
cc @HipsterBrown

HipsterBrown:
Will review this tonight.

lib/controller.js
```diff
@@ -66,11 +66,11 @@ controller.setupLocal = function(opts) {
 
 
 Tessel.list = function(opts) {
-
   return new Promise(function(resolve, reject) {
-
+    log.spinner.stop();
```
HipsterBrown: Is this just making sure the spinner is defaulted to stop before starting it, in case it was not stopped by `callControllerWith()`?

lib/discover.js
```diff
@@ -5,14 +5,17 @@ var util = require('util');
 var Emitter = events.EventEmitter;
 
 // Third Party Dependencies
-var debug = require('debug')('discovery');
+// ...
```
HipsterBrown: Keeping this for a reason?

lib/discover.js
```diff
@@ -5,14 +5,17 @@ var util = require('util');
...
 // Internal
 var lan = require('./lan-connection');
-var logs = require('./logs');
+var log = require('./log');
 var Tessel = require('./tessel/tessel');
 var usb = require('./usb-connection');
 
+function debug(message) {
+  log.debug(`(discovery) ${message}`);
+}
```
HipsterBrown: Smart 👍

lib/init.js
```diff
@@ -176,7 +182,8 @@ module.exports = function(opts) {
   // Set interactive boolean off of CLI flags
   options = opts;
 
-  console.log('Initializing Tessel 2 Project...');
+  log.info('Initializing Tessel 2 Project...');
+  log.spinner.start();
```
HipsterBrown: If this is started here, does `log.spinner.start()` need to be in all the async functions like `loadNpm`? Probably doesn't hurt to call it in each async function, just wondering.

lib/discover.js
HipsterBrown: Nevermind, I see that it's a pattern in the other files as well. 👍

lib/tessel/access-point.js
```diff
@@ -126,7 +128,10 @@ Tessel.prototype.createAccessPoint = function(opts) {
 
     var turnAccessPointOn = () => this.simpleExec(commands.turnAccessPointOn());
 
-    var commitAndClosePromise = () => new Promise((resolve, reject) => commitAndClose(this, status, resolve, reject));
+    var commitAndClosePromise = () => {
+      log.spinner.stop();
```
HipsterBrown: Could this be called in `commitAndClose` since that's another async action that restarts a few system processes like `wifi`, `dnsmasq`, and `dhcp`, which could take a few moments worthy of a spinner.

lib/tessel/access-point.js
```diff
@@ -146,12 +151,16 @@ Tessel.prototype.createAccessPoint = function(opts) {
...
     .then(() => {
+      log.spinner.stop();
+
       // When an AP exists, change the status to
       // reflect an "update" vs. "create":
       status = 'Updated Access Point successfully. ';
       return setupAccessPoint();
     })
     .catch(() => {
+      log.spinner.stop();
```
HipsterBrown: `setupAccessPoint` calls this after the other async actions of `setAccessPoint` and `setLanNetwork`. Could this be removed so the spinner continues until the access point is actually finished being setup?

lib/tessel/access-point.js
```diff
@@ -146,12 +151,16 @@ Tessel.prototype.createAccessPoint = function(opts) {
 
   return this.simpleExec(commands.getAccessPointSSID())
     .then(() => {
+      log.spinner.stop();
```
HipsterBrown: Similar note as this comment -> https://github.com/tessel/t2-cli/pull/768/files#r66899373

lib/tessel/deploy.js
```diff
@@ -333,14 +333,17 @@ exportables.sendBundle = function(tessel, opts) {
...
 
             // Log write
-            logs.info('Writing project to %s on %s (%d kB)...', memtype, tessel.name, bundle.length / 1000);
+            log.info('Writing project to %s on %s (%d kB)...', memtype, tessel.name, bundle.length / 1000);
+            log.spinner.start();
 
             // Calling receive to know when the process closes
             tessel.receive(remoteProcess, (err) => {
               if (err) {
                 return reject(err);
```
HipsterBrown: Will the spinner stop if this `Promise` is rejected?

lib/tessel/update.js
```diff
@@ -54,18 +54,20 @@ Tessel.prototype.updateOpenWRT = function(opts, image) {
   return this.fixOldUpdateScripts()
     .then(() => {
       return new Promise((resolve, reject) => {
-        logs.info('Updating OpenWRT (1/2)');
+        log.info('Updating OpenWRT (1/2)');
+        log.spinner.start();
         // Write the new image to a file in the /tmp dir
         this.connection.exec(commands.openStdinToFile(updatePath), (err, remoteProc) => {
           if (err) {
             return reject(err);
```
HipsterBrown: Will the spinner stop if this `Promise`  is rejected?

lib/tessel/update.js
```diff
@@ -77,20 +79,21 @@ Tessel.prototype.updateOpenWRT = function(opts, image) {
...
-          logs.info('Configuration is saved during update.');
+          log.info('Configuration is saved during update.');
           sysupgradeCommand = commands.sysupgrade(updatePath);
         }
 
-        // The USBDaemon will cut out or the SSH command will close        
+        // The USBDaemon will cut out or the SSH command will close
         this.connection.exec(sysupgradeCommand, (err, remoteProc) => {
           if (err) {
             return reject(err);
```
HipsterBrown: Same notes about [rejected Promises](https://github.com/tessel/t2-cli/pull/768/files#r66899802) and stopping the spinner.

lib/tessel/wifi.js
```diff
@@ -6,14 +6,14 @@
...
-var logs = require('../logs');
+var log = require('../log');
 var Tessel = require('./tessel');
 
 
 Tessel.prototype.findAvailableNetworks = function() {
-  logs.info('Scanning for available networks...');
   return this.simpleExec(commands.scanWiFi())
     .then(function wifiScanResults(result) {
+      log.spinner.stop();
```
HipsterBrown: When does this spinner start?

lib/tessel/wifi.js
```diff
@@ -63,6 +63,7 @@ Tessel.prototype.getWifiInfo = function() {
 
       return this.simpleExec(commands.getIPAddress())
         .then(function(ipResults) {
+          log.spinner.stop();
```
HipsterBrown: When does this spinner start?

HipsterBrown:
> Do we persist logging flags in preferences?

I say yes and agree with following the `npm config` pattern. This leaves room to grow into other config flags if needed.

rwaldron:
After reading the review comments, I decided to test another approach. As a result, I've pushed a completely revised version of this patch.

HipsterBrown:
Reviewing and smoke testing now.

bin/tessel-2.js
```diff
@@ -41,6 +41,45 @@ if (!isRoot()) {
...
+        module.exports.closeFailedCommand
+      );
+  }
+
+  return cr;
+};
+
+
+controller.installDrivers = function() {
+  require('./tessel-install-drivers');
```
HipsterBrown: Is this being required again for a reason? It's also on line 16:

```js
var drivers = require('./tessel-install-drivers');
```

lib/usb-connection.js
```diff
@@ -23,7 +29,7 @@ try {
   // var VENDOR_REQ_IN  = usb.LIBUSB_REQUEST_TYPE_VENDOR | usb.LIBUSB_RECIPIENT_DEVICE | usb.LIBUSB_ENDPOINT_IN;
```
HipsterBrown: Do we still need this?

lib/usb-connection.js
```diff
@@ -23,7 +29,7 @@ try {
...
 } catch (e) {
   haveusb = false;
-  logs.err('WARNING: No usb controller found on this system.');
+  log.error('WARNING: No usb controller found on this system.');
 }
```
HipsterBrown: I ran into this error while testing this patch - `[Error: Error: Module version mismatch. Expected 47, got 46.]`, which came from the internals of the `usb` module and lefts the process hanging until I force quit it with `kill -9`. That error didn't come through until I added `log.error(e)` above line 31. I'm still not sure why it was left hanging open though because this didn't happen on `master`. Maybe there should be a `Promise.reject()`? 

The solution for that error was to reinstall my `node_modules` by the way.

rwaldron: Seems like an entirely unrelated problem that normal use can't possibly warrant handling for :P


that is... i think you changed node versions, npm installed, then changed back.

lib/usb-connection.js
rwaldron: I don't know, but that should be addressed in a different patch

lib/tessel/update.js
rwaldron: I've updated this so that both `closeSuccessfulCommand` and `closeFailedCommand` can stop the spinner

lib/usb-connection.js
HipsterBrown: Yea, I did something screwy with `nvm` when creating new bash sessions. The comment was mostly recognizing some missing error feedback around this part of logging. Could we also have a line that logs the error coming from the `try...catch` around `require('usb')`? Or would that be part of another patch?

bin/tessel-2.js
rwaldron: Pfft. I never even noticed that. I just copied this from the old code.

rwaldron: Fixed!

lib/usb-connection.js
rwaldron: That's a fail fast error, I don't think we need to do anything

HipsterBrown:
Overall, I love the simplification of spinner state management. This is going to be a very useful patch. 

LGTM 👍

rwaldron:
@HipsterBrown thank you so much for the several-days-long review, I appreciate your help enormously <3

PR #772, opened 2016-06-16
[Jumping Off Code] Adds simple, untested implementation of rust deploys

johnnyman727:
This is a jumping off branch for someone who wants to finish adding support for deploying Rust projects. This code has been untested but essentially hands off a project directory to the [cross compilation server](https://github.com/tessel/rust-compilation-server) and returns a tarred executable to the regularly scheduled deploy process.

There are two functions that need to be completed and these functions are called automatically by the logic that gets run when `t2 run/pass` is provided to the CLI. First, `exportables.tarBundle` provides an `opts` object with info about the directory being deployed (including path) and needs to resolve a Promise with a `Buffer` of the tarred executable. `exportables.preRun` needs to make sure the permissions of that executable are set properly on Tessel 2 for it to be run. There is also an `exportables.preBundle` function but we don't need to worry about that for Rust (it's used for special cases in JavaScript projects, primarily).

So here is a checklist of what needs to get done:
- [ ] Make `exportables.tarBundle` production-ready. This means making sure it's working, checking error conditions, considering corner cases, etc. I also recommend changing the cross compilation server to tar up the executable server side before sending it down as that would speed up transfers and reduce logic in the CLI.
- [ ] Set up the cross compilation server at a publicly available address
- [ ] Ensure that `exportables.preRun` works
- [ ] Reduce redundant mentions of `rust_executable`. I believe that name should be a variable instead of hardcoding it several times in this file and in `/tessel/commands.js`. 
- [ ] Add some unit tests
- [ ] Make sure `grunt` is happy.


cflewis:
I'll grab the lock on this and see what I can do :)

lib/tessel/deployment/rust.js
```diff
@@ -1,10 +1,17 @@
 // System Objects
 var path = require('path');
+var http = require('http');
+var path = require('path');
 
 // Third Party Dependencies
 var fs = require('fs-extra');
 var tags = require('common-tags');
 var toml = require('toml');
+require('toml-require').install();
```
rwaldron: Where is this used and why? Presently, the module `toml` is handling the parsing of toml files: https://github.com/tessel/t2-cli/blob/a7c32f6eeaf74c6f616b84c4db005e93b4c3cd13/lib/tessel/deployment/rust.js#L30-L36

johnnyman727:
Closing in favor of #774.

PR #779, opened 2016-06-22
t2 init: broad cleanup

rwaldron:
WIP. I still have more JS tests to write. 

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -9,11 +9,9 @@ const updateNotifier = require('update-notifier');
 const isRoot = require('is-root');
 
 // Internal
-var controller = require('../lib/controller');
 var CrashReporter = require('../lib/crash-reporter');
-var init = require('../lib/init');
+var controller = require('../lib/controller');
 var log = require('../lib/log');
-var drivers = require('./tessel-install-drivers');
 var Preferences = require('../lib/preferences');
 
 const CLI_ENTRYPOINT = 'cli.entrypoint';
@@ -40,40 +38,6 @@ if (!isRoot()) {
     });
   }
 }
-
-controller.t2Init = init.initProject;
-
-controller.crashReporter = function(options) {
-  var cr = Promise.resolve();
-
-  // t2 crash-reporter --on
-  if (options.on) {
-    cr = CrashReporter.on();
-  } else if (options.off) {
-    // t2 crash-reporter --off
-    cr = CrashReporter.off();
-  }
-
-  // t2 crash-reporter --test
-  if (options.test) {
-    // not handling failures, as we want to trigger a crash
-    cr.then(CrashReporter.test)
-      .then(module.exports.closeSuccessfulCommand);
-  } else {
-    cr.then(CrashReporter.status)
-      .then(
-        module.exports.closeSuccessfulCommand,
-        module.exports.closeFailedCommand
-      );
-  }
-
-  return cr;
-};
-
-
-controller.installDrivers = function() {
-  return drivers.install();
-};
 
 function makeCommand(commandName) {
   return parser.command(commandName)
@@ -323,7 +287,7 @@ parser.command('init')
   .callback(options => {
     log.level(options.loglevel);
 
-    callControllerWith('t2Init', options);
+    callControllerWith('createNewProject', options);
   })
   .option('interactive', {
     flag: true,
@@ -334,7 +298,7 @@ parser.command('init')
     metavar: 'LANG',
     abbr: 'l',
     default: 'js',
-    help: 'The language to use <javascript|rust|js|rs>. Javascript by default'
+    help: 'The language to use <javascript|rust|js|rs>. JavaScript by default'
   })
   .help('Initialize repository for your Tessel project');
 
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -8,7 +8,10 @@ var colors = require('colors');
 var semver = require('semver');
 
 // Internal
+var CrashReporter = require('./crash-reporter');
 var discover = require('./discover');
+var drivers = require('../bin/tessel-install-drivers');
+var init = require('./init');
 var log = require('./log');
 var Menu = require('./menu');
 var updates = require('./update-fetch');
@@ -979,6 +982,41 @@ controller.tesselEnvVersions = opts => {
       });
   });
 };
+
+controller.createNewProject = init.createNewProject;
+
+controller.crashReporter = function(options) {
+  var cr = Promise.resolve();
+
+  // t2 crash-reporter --on
+  if (options.on) {
+    cr = CrashReporter.on();
+  } else if (options.off) {
+    // t2 crash-reporter --off
+    cr = CrashReporter.off();
+  }
+
+  // t2 crash-reporter --test
+  if (options.test) {
+    // not handling failures, as we want to trigger a crash
+    cr.then(CrashReporter.test)
+      .then(module.exports.closeSuccessfulCommand);
+  } else {
+    cr.then(CrashReporter.status)
+      .then(
+        module.exports.closeSuccessfulCommand,
+        module.exports.closeFailedCommand
+      );
+  }
+
+  return cr;
+};
+
+controller.installDrivers = function() {
+  return drivers.install();
+};
+
+
 
 module.exports = controller;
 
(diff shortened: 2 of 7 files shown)


bin/tessel-2.js
```diff
@@ -41,40 +39,6 @@ if (!isRoot()) {
...
-  }
-
-  return cr;
-};
-
-
-controller.installDrivers = function() {
-  return drivers.install();
-};
-
```
rwaldron: This is just being moved to `lib/controller.js`

bin/tessel-2.js
```diff
@@ -323,7 +287,7 @@ parser.command('init')
   .callback(options => {
     log.level(options.loglevel);
 
-    callControllerWith('t2Init', options);
+    callControllerWith('createNewProject', options);
```
rwaldron: This renaming choice was subjective, chosen to be more explicit about the operation performed

bin/tessel-2.js
```diff
@@ -334,7 +298,7 @@ parser.command('init')
     metavar: 'LANG',
     abbr: 'l',
     default: 'js',
-    help: 'The language to use <javascript|rust|js|rs>. Javascript by default'
+    help: 'The language to use <javascript|rust|js|rs>. JavaScript by default'
```
rwaldron: Just a typo

lib/controller.js
```diff
@@ -8,7 +8,10 @@ var colors = require('colors');
 var semver = require('semver');
 
 // Internal
+var CrashReporter = require('./crash-reporter');
 var discover = require('./discover');
+var drivers = require('../bin/tessel-install-drivers');
+var init = require('./init');
```
rwaldron: See previous note about moving all `controller` stuff into `lib/controller.js`

lib/controller.js
```diff
@@ -980,6 +983,41 @@ controller.tesselEnvVersions = opts => {
...
+
+  return cr;
+};
+
+controller.installDrivers = function() {
+  return drivers.install();
+};
+
+
+
```
rwaldron: See previous note about moving all `controller` stuff into `lib/controller.js`

lib/controller.js
```diff
@@ -980,6 +983,41 @@ controller.tesselEnvVersions = opts => {
...
+        module.exports.closeSuccessfulCommand,
+        module.exports.closeFailedCommand
+      );
+  }
+
+  return cr;
+};
+
+controller.installDrivers = function() {
+  return drivers.install();
```
rwaldron: I removed the `then(...)` here, because that's handled by `callControllerWith(...)` in `bin/tessel-2.js`

lib/init/index.js
```diff
@@ -12,42 +13,45 @@ var languages = {
...
   } else {
     // Otherwise generate a project in that language
-    return lang.generateProject(opts);
+    return lang.generateProject(options);
   }
 };
 
 // Determine the langauge to initialize the project with
-exportables.resolveLanguage = (opts) => {
+exportables.resolveLanguage = (input) => {
```
rwaldron: Deployment also has an  `exportables.resolveLanguage`, which accepts a string. I plan to refactor both to use the same functionality, so this is just a first step towards that.

lib/init/rust.js
```diff
@@ -22,45 +22,36 @@ exportables.generateProject = (opts) => {
...
+            return reject(new Error(`Could not create ${srcDir}; ${error.toString()}`));
           }
           // Copy over config file, the blinky main, and the toml file
-          fs.createReadStream(path.resolve(__dirname, './../../resources/rust/', file_toml)).pipe(fs.createWriteStream(path_toml));
+          fs.createReadStream(path.resolve(__dirname, './../../resources/rust/', cargoToml)).pipe(fs.createWriteStream(cargoTomlDir));
           log.info('Initialized Cargo project...');
-          fs.createReadStream(path.resolve(__dirname, './../../resources/rust/', file_src)).pipe(fs.createWriteStream(path_src));
-          log.info(`Wrote "Hello World" to ${path_src}`);
+          fs.createReadStream(path.resolve(__dirname, './../../resources/rust/', mainRs)).pipe(fs.createWriteStream(mainRsDir));
+          log.info(`Wrote "Hello World" to ${mainRsDir}`);
```
rwaldron: This one function does the work that several function do in the JS version. For now, I've left it as is, but we should revisit this

test/unit/bin-tessel-2.js
```diff
@@ -924,42 +922,84 @@ exports['Tessel (init)'] = {
...
-      test.equal(this.initProject.callCount, 1);
-      // Ensure it checks the language being requested
-      test.equal(this.resolveLanguage.callCount, 1);
-      // It should generate a js project
-      test.equal(this.generateJavaScriptProject.callCount, 1);
-      // It should not generate a Rust project
-      test.equal(this.generateRustProject.callCount, 0);
-      // Ensure it continues to call our exit function
-      test.equal(this.successfulCommand.callCount, 1);
+    resolve.then(() => {
```
rwaldron: All CLI operations result in a call to a function that returns a promise, therefore all must be tested by observing the outcome once resolved or rejected.

lib/init/javascript.js
```diff
@@ -33,14 +31,14 @@ exportables.loadNpm = () => {
 };
 
 // Resolve an npm cofig list, or nothing (existance is not needed)
-exportables.getNpmConfig = (npm) => {
+exportables.resolveNpmConfig = (npm) => {
```
rwaldron: In general, I think things that are "resolved" by a promise should be named accordingly.

lib/init/javascript.js
```diff
@@ -214,8 +211,8 @@ exportables.generateProject = (opts) => {
     .then(exportables.getDependencies)
     .then(exportables.npmInstall)
     .then(exportables.createTesselinclude)
```
rwaldron: In a follow up, this should be made more "general purpose"

lib/init/index.js
```diff
@@ -12,42 +13,45 @@ var languages = {
...
 
   // If somehow a language option wasn't provided
-  if (!opts.lang) {
-    // Return JS as default
-    return languages['js'];
+  if (!input) {
+    return languages.js;
+  }
+
+  input = String(input).toLowerCase();
```
HipsterBrown: Is there a case where `input` wouldn't be a `String`? I'm guessing it's coerced to one just in case. 

If it's not a `String`, could we `return null` earlier?

rwaldron: I believe this is an artifact of an early iteration of the cleanup and can be removed.

lib/init/rust.js
```diff
@@ -22,45 +22,36 @@ exportables.generateProject = (opts) => {
...
-        fs.mkdir(dir_src, (err) => {
-          if (err) {
-            return reject(new Error(mkdir_err(dir_src)));
+        fs.mkdir(srcDir, (error) => {
+          if (error) {
+            return reject(new Error(`Could not create ${srcDir}; ${error.toString()}`));
           }
           // Copy over config file, the blinky main, and the toml file
-          fs.createReadStream(path.resolve(__dirname, './../../resources/rust/', file_toml)).pipe(fs.createWriteStream(path_toml));
+          fs.createReadStream(path.resolve(__dirname, './../../resources/rust/', cargoToml)).pipe(fs.createWriteStream(cargoTomlDir));
```
HipsterBrown: Could this be simplified to use `fs.copy` or `fs.copySync` is we want to keep it synchronous, like in `javascript.js`?

lib/init/index.js
rwaldron: Confirmed. Revision inbound

lib/init/rust.js
rwaldron: Yes, you're right. Revision to follow

lib/init/rust.js
```diff
@@ -22,45 +22,47 @@ exportables.generateProject = (opts) => {
...
+
+            fs.copy(path.resolve(__dirname, './../../', 'resources/rust', mainRs), dirAndMainRs, (error) => {
+              if (error) {
+                return reject(new exportables.CreateError(dirAndMainRs, error));
+              }
+
+              log.info(`Wrote "Hello World" to ${dirAndMainRs}`);
+              resolve();
+            });
+          });
```
rwaldron: @HipsterBrown ^^

HipsterBrown:
Awesome. Thanks for tackling this @rwaldron. Once tests pass LGTM :+1:

rwaldron:
Thanks @HipsterBrown!

PR #787, opened 2016-07-02
t2 run: move JS-specific "post run" operation to a language specific definition

rwaldron:
@johnnyman727 This should be reviewed and merged before any of the pending Rust support changes.

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -355,23 +355,24 @@ exportables.sendBundle = function(tessel, opts) {
 };
 
 
-exportables.run = function(tessel, opts) {
-  if (opts.resolvedEntryPoint === undefined) {
-    opts.resolvedEntryPoint = opts.entryPoint;
+exportables.run = function(tessel, options) {
+  if (options.resolvedEntryPoint === undefined) {
+    options.resolvedEntryPoint = options.entryPoint;
   }
 
-  log.info('Running %s...', opts.resolvedEntryPoint);
+  log.info('Running %s...', options.resolvedEntryPoint);
 
   return new Promise((resolve, reject) => {
-    var prom = Promise.resolve();
+    var preRun = Promise.resolve();
+    var lang = options.lang;
 
-    if (opts.lang.preRun) {
-      prom = opts.lang.preRun(tessel, opts);
+    if (lang.preRun) {
+      preRun = lang.preRun(tessel, options);
     }
 
-    return prom.then(() => {
+    return preRun.then(() => {
       tessel.connection.exec(
-        commands[opts.lang.meta.extname].execute(Tessel.REMOTE_RUN_PATH, opts.resolvedEntryPoint), {
+        commands[lang.meta.extname].execute(Tessel.REMOTE_RUN_PATH, options.resolvedEntryPoint), {
           pty: true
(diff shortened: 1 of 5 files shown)


lib/tessel/deployment/rust.js
```diff
@@ -60,5 +60,6 @@ exportables.preRun = function(tessel) {
   });
 };
 
+// exportables.postRun = function(tessel, remote) {};
```
rwaldron: This is here as "pseudo documentation" until the language support API churn slows and we can start documenting properly

lib/tessel/deployment/javascript.js
```diff
@@ -68,6 +68,17 @@ exportables.preBundle = function(opts) {
   return exportables.resolveBinaryModules(opts);
 };
 
+// exportables.preRun = function(tessel) {};
+
+exportables.postRun = function(tessel, remote) {
```
johnnyman727: It's not great that the arguments for `preRun`, `preBundle, `tarBundle`, and `postRun` all have very different arguments. Is there any way we can make them more consistent and intuitive?

rwaldron: I'm not sure that holds up very well, considering:

- There is no such thing as a "remote process" when `preRun` is called, but would also be irrelevant to that operation.
- `preBundle` and `tarBundle` are about files and directories, not Tessel or "remote process". 


The best that I think can be done is:

- `preBundle(options)` 
- `tarBundle(options)` 
- `preRun(tessel, options)`
- `postRun(tessel, options)` (`options` will contain `remoteProcess`)

lib/tessel/deploy.js
```diff
@@ -355,23 +355,24 @@ exportables.sendBundle = function(tessel, opts) {
...
-    if (opts.lang.preRun) {
-      prom = opts.lang.preRun(tessel, opts);
+    if (lang.preRun) {
+      preRun = lang.preRun(tessel, options);
     }
 
-    return prom.then(() => {
+    return preRun.then(() => {
       tessel.connection.exec(
-        commands[opts.lang.meta.extname].execute(Tessel.REMOTE_RUN_PATH, opts.resolvedEntryPoint), {
```
rwaldron: This is just renaming, nothing interesting

lib/tessel/deploy.js
```diff
@@ -355,23 +355,24 @@ exportables.sendBundle = function(tessel, opts) {
...
+        commands[lang.meta.extname].execute(Tessel.REMOTE_RUN_PATH, options.resolvedEntryPoint), {
```
rwaldron: This is just renaming, nothing interesting

lib/tessel/deployment/javascript.js
```diff
@@ -63,11 +63,21 @@ var exportables = {
   lists: lists,
 };
 
-
-exportables.preBundle = function(opts) {
-  return exportables.resolveBinaryModules(opts);
+/*
+exportables.preRun = function(tessel, options) {
+  return Promise.resolve();
 };
```
rwaldron: This is not removed, just relocated next to `exportables.tarBundle`

johnnyman727:
LGTM, I think the API changes are as good as can be.

PR #789, opened 2016-07-04
Add an advisory in package.json/engines that Node needs to be version 4

cflewis:
Add a warning that you need to be using Node 4 LTS

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
   },
   "preferGlobal": true,
   "engines": {
-    "node": ">=4.2.0"
+    "node": ">=4.2.0 <5.0"
   },
   "contributors": [
     "Jon McKay <johnnyman727@gmail.com>",


LinusU:
I've been using Node 6 with t2-cli, why wouldn't it work? Was there something breaking in 6.2.0?

johnnyman727:
@LinusU I'm really curious how you compiled the `usb` module with Node 6!

johnnyman727:
I think a fix that might be more maintainable is by updating the [`engines` field of the package.json](https://docs.npmjs.com/files/package.json). That way, `npm` will automatically print out a warning when the CLI is installed with an incompatible version of Node and we won't have to explicitly remember to update the README when we make the CLI Node 6.x compatible (because we'll get a reminder every time a person installs).

LinusU:
Hmmm, maybe I'm dreaming and I actually used nvm...

I think that the engines field would be nice 👍

johnnyman727:
LGTM!

PR #790, opened 2016-07-06
Rust deployment

johnnyman727:


cflewis:
LGTM

tikurahul:
LGTM

badboy:
I took a look at the code and also ran it.
I had to change a previously mentioned thing to make it work on Node v6:

```patch
diff --git i/lib/tessel/deployment/rust.js w/lib/tessel/deployment/rust.js
-    var details = exportables.meta.checkConfiguration(opts.target);
+    var details = exportables.meta.checkConfiguration(opts.target, null, opts.resolvedEntryPoint);
 ```

I know v6 might not be officially supported. But if that is the only thing holding it back, can it be added?

File: bin/tessel-2.js [hunk 1]
```diff
@@ -203,8 +203,13 @@ makeCommand('run')
     default: false,
     help: 'Deploy a project containing all files within, including those not used by the program, excluding any files matched by non-negated rules in .tesselignore and including any files matched by rules in .tesselinclude. Program is started from specified file.'
   })
+  .option('rustCC', {
```
rwaldron: I'd like to suggest making this `rustcc`, so it's easier for end users to type

bin/tessel-2.js
```diff
@@ -203,8 +203,13 @@ makeCommand('run')
...
+    // TODO: make this point to a long-lived server somewhere
+    default: 'http://localhost:8080',
+    help: 'Specify the location and port of the Rust cross-compilation server.'
+  })
   .help(`
-    Deploy a script to Tessel and run it with Node.
+    Deploy a script to Tessel and run it.
```
rwaldron: Is this accurate? I think it should probably say "Compile and deploy a Rust program" (or something like that?

lib/tessel/deployment/rust.js
```diff
@@ -1,23 +1,27 @@
...
-
+var Reader = require('fstream').Reader;
+var tar = require('tar');
 
 // Internal
 var commands = require('../commands');
 var lists = require('./lists/rust');
 var Tessel = require('../tessel');
-
+var log = require('../../log');
```
rwaldron: nit: let's try our best to keep these alphabetical :)

bin/tessel-2.js
LinusU: or possibly `rust-cc`? Unix command line options are almost always in kebab-case

lib/tessel/deployment/rust.js
```diff
@@ -33,32 +37,134 @@ var exportables = {
         program
       };
     },
-    shell: () => {
+    shell: (opts) => {
       return tags.stripIndent `
         #!/bin/sh
-        exec /app/remote-script/rust_executable
+        exec /app/remote-script/${opts.resolvedEntryPoint}
```
rwaldron: I think I made a mistake when I wrote "/app/remote-script/", because the paths are presently defined as:

```js
Tessel.REMOTE_PUSH_PATH = '/app/';
Tessel.REMOTE_RUN_PATH = '/tmp/remote-script/';
```

lib/tessel/deployment/rust.js
```diff
@@ -33,32 +37,134 @@ var exportables = {
...
+    // Save our incoming compiled buffer to this array
+    var buffers = [];
+
+    // Parse out the components of the remote rust compilation server
+    if (!(opts.rustCC.startsWith('http://') || opts.rustCC.startsWith('https://'))) {
+      // Without an HTTP protocol definition, url.parse will fail;
+      // assume http:// was implied
+      opts.rustCC = 'http://' + opts.rustCC;
+    }
+    var destination = url.parse(opts.rustCC);
```
rwaldron: If you agree with my recommendation re: `rustCC => rustcc`, don't forget to update here.

lib/tessel/deployment/rust.js
```diff
@@ -33,32 +37,134 @@ var exportables = {
...
+exportables.remoteRustCompilation = (opts) => {
+  log.info('Compiling Rust code remotely...');
+  return new Promise(function(resolve, reject) {
+
+    // Save our incoming compiled buffer to this array
+    var buffers = [];
+
+    // Parse out the components of the remote rust compilation server
+    if (!(opts.rustCC.startsWith('http://') || opts.rustCC.startsWith('https://'))) {
+      // Without an HTTP protocol definition, url.parse will fail;
```
rwaldron: I'm not able to reproduce a failure?

rwaldron: Wait, I see now...

```js
var url = require("url");
var fixture = "failure.com";
var destination = url.parse(fixture);

console.log(destination);

/*
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: 'failure.com',
  path: 'failure.com',
  href: 'failure.com' }
 */
```

So, not a failure, but it's missing the information that's necessary, ie. hostname

lib/tessel/deployment/rust.js
```diff
@@ -33,32 +37,134 @@ var exportables = {
...
+      port: destination.port,
+      path: '/rust-compile',
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/octet-stream',
+      }
+    };
+
+    // Set up the request
+    var post_req = http.request(post_options, function(res) {
```
rwaldron: If it's safe to use arrow functions, then we should use them (since the rest of the code base is migrating that way)

lib/tessel/deployment/rust.js
```diff
@@ -33,32 +37,134 @@ var exportables = {
...
 
+exportables.remoteRustCompilation = (opts) => {
+  log.info('Compiling Rust code remotely...');
+  return new Promise(function(resolve, reject) {
+
+    // Save our incoming compiled buffer to this array
+    var buffers = [];
+
+    // Parse out the components of the remote rust compilation server
+    if (!(opts.rustCC.startsWith('http://') || opts.rustCC.startsWith('https://'))) {
```
rwaldron: This can be simplified to:

```js
 if (!opts.rustcc.startsWith('http'))  {
``

File: lib/tessel/deployment/rust.js (hunk 7 above)
Comment (rwaldron) [inline lib/tessel/deployment/rust.js]: I don't think the vars need to be prefixed with `post_`. `request` and `options` are sufficient (since we immediately know what they are)

**Comment by rwaldron** [comment]
I've just landed #787, this will need to be rebased.

File: bin/tessel-2.js (hunk 1 above)
Comment (johnnyman727) [inline bin/tessel-2.js]: I'm good with `rust-cc`. That's a little more intuitive to me.

File: bin/tessel-2.js (hunk 2 above)
Comment (johnnyman727) [inline bin/tessel-2.js]: The command (`run`) is language agnostic so we probably shouldn't mention any engines or pre-bundle steps.

File: lib/tessel/deployment/rust.js (hunk 4 above)
Comment (johnnyman727) [inline lib/tessel/deployment/rust.js]: This is related to https://github.com/tessel/t2-cli/issues/785 and will have to be fixed when that linked issue is. Basically, we untar to `/remote-script` whether we `push` or `run` so `/app/remote-script` gets created.

File: lib/tessel/deployment/rust.js (hunk 8 above)
Comment (johnnyman727) [inline lib/tessel/deployment/rust.js]: Ah yes, sorry about that misdirection @cflewis.

File: bin/tessel-2.js (hunk 1 above)
Comment (rwaldron) [inline bin/tessel-2.js]: ¯\_(ツ)_/¯

> Unix command line options are almost always in kebab-case

When they are distinctly two words, yes... but "cc" is an abbreviation right? I'm assuming "cross compiler"? [That's two words](https://en.wikipedia.org/wiki/Cross_compiler), so by this logic, you want "rust-c-c" which is just cruel.

File: bin/tessel-2.js (hunk 2 above)
Comment (rwaldron) [inline bin/tessel-2.js]: Oh right... For some reason I thought this was some Rust specific help text—likely brought on by the new option just above. Sorry for confusion.

I suppose can generalize this further: "Deploy an application to Tessel and run it."

File: lib/tessel/deployment/rust.js (hunk 4 above)
Comment (rwaldron) [inline lib/tessel/deployment/rust.js]: I'll get to work on that

**Comment by rwaldron** [comment]
Needs to be rebased.

File: test/unit/deploy.js [hunk 9]
```diff
@@ -1,3 +1,5 @@
+var javascript = require('../../lib/tessel/deployment/javascript');
```
rwaldron: This object is already globally available in this file as `deployment.js` (same thing for `deployment.rs` and `deployment.py`)

lib/tessel/deployment/rust.js
rwaldron: This no longer needs to be changed. In https://github.com/tessel/t2-cli/commit/e334d8eb23dd4eb902fe0b82ba1babd8e8174871, I opted to simply accept the flash path as-is, since it's been like that for since https://github.com/tessel/t2-cli/commit/4eb17770a20e68ad699b88e47f035ea03b41558f and it's really not a problem.

bin/tessel-2.js
johnnyman727: I'll go with `rustcc` because it's shortest and, let's be honest, most users are very unlikely to go through the trouble of setting up their own cross compilers. It's pretty much just a developer flag.

johnnyman727:
Cleaned up commits and addressed PR comments. I'm going to work on getting the cross compilation server deployed on digital ocean, add the permanent IP here, and then merge.

johnnyman727:
Smoke test for anyone that wants to try this:
```
mkdir rust-fun; cd rust-fun
t2 init --lang=rust
t2 run Cargo.toml
```
Watch the LEDs blink to your heart's content.

johnnyman727:
:clap:

cflewis:
Sweet! Awesome work Jon! 🤗

PR #792, opened 2016-07-10
AP/WIFI check password validity for encryption type

### LINKED ISSUE #760: AP/WIFI check password validity for encryption type
OpenWRT will silently fail if you try to set the password on an access point (or wifi connection but that's less of an issue) to something that is not valid for the encryption type. For example, I tried to create an access point with the encryption type of `psk2` and the password of `stones` but that password was not long enough to be valid. The CLI reported that it was set correctly but the network was never created.

Ideally, we should be able to check for password-encryption validity before we attempt to set it in OpenWRT. The CLI should WARN the user if it is invalid and abort.

laiedwin:
This pull request fixes #760.

**Changes**:
* Removes WPA Enterprise options (`wpa`, `wpa2`) from `securityOptions` array.
* Adds network passphrase validation and relevant tests.

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -664,7 +664,7 @@ controller.connectToNetwork = function(opts) {
   var ssid = opts.ssid;
   var password = opts.password;
   var security = opts.security;
-  var securityOptions = ['none', 'wep', 'psk', 'psk2', 'wpa', 'wpa2'];
+  var securityOptions = ['none', 'wep', 'psk', 'psk2'];
   return new Promise(function(resolve, reject) {
       if (!ssid) {
         return reject('Invalid credentials: must set SSID with the -n or --ssid option.');
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -681,9 +681,8 @@ controller.connectToNetwork = function(opts) {
       if (security === 'wep') {
         // WEP passphrases can be 10, 26, or 58 hexadecimal digits long.
         // Match for hexadecimal characters:
-        const regex = /^[0-9a-f]+$/i;
-        if (!regex.test(password)) {
-          return reject('Invalid passphrase: WEP keys must consist of hexadecimal digits.');
+        if (isNaN(`0x${password}`)) {
+          return reject('Invalid passphrase: WEP keys must consist of hexadecimal digits, i.e. 0 through 9 and "a" through "f".');
         }
         // Then test for length:
         const length = password.length;
@@ -697,9 +696,9 @@ controller.connectToNetwork = function(opts) {
         // Match ASCII codes for all 127 printable ASCII characters:
         const asciiRegex = /^[\x00-\x7F]{8,63}$/;
         // Match exactly 64 hexadecimal digits:
-        const hexRegex = /^[0-9a-f]{64}$/i;
-        // Reject if both regexes fail.
-        if (!asciiRegex.test(password) && !hexRegex.test(password)) {
+        const hexTest = !isNaN(`0x${password}`) && password.length === 64;
+        // Reject if both tests fail.
+        if (!asciiRegex.test(password) && !hexTest) {
           return reject('Invalid passphrase: WPA/WPA2-PSK passkeys must be 8-63 ASCII characters, or 64 hexadecimal digits.');
         }
       }
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -677,6 +677,32 @@ controller.connectToNetwork = function(opts) {
       if (security && securityOptions.indexOf(security) < 0) {
         return reject(`"${security}" is not a valid security option. Please choose on of the following: ${securityOptions.join(', ')}`);
       }
+      
+      if (security === 'wep') {
+        // WEP passphrases can be 10, 26, or 58 hexadecimal digits long.
+        // Match for hexadecimal characters:
+        const regex = /^[0-9a-f]+$/i;
+        if (!regex.test(password)) {
+          return reject('Invalid passphrase: WEP keys must consist of hexadecimal digits.');
+        }
+        // Then test for length:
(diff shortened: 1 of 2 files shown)


lib/controller.js
```diff
@@ -677,6 +677,32 @@ controller.connectToNetwork = function(opts) {
       if (security && securityOptions.indexOf(security) < 0) {
         return reject(`"${security}" is not a valid security option. Please choose on of the following: ${securityOptions.join(', ')}`);
       }
+      
+      if (security === 'wep') {
+        // WEP passphrases can be 10, 26, or 58 hexadecimal digits long.
+        // Match for hexadecimal characters:
+        const regex = /^[0-9a-f]+$/i;
+        if (!regex.test(password)) {
+          return reject('Invalid passphrase: WEP keys must consist of hexadecimal digits.');
```
HipsterBrown: Could you clarify "hexadecimal digits" in this error message? Something like:

>  Invalid passphrase: WEP keys must consist of hexadecimal digits, i.e 0 through 9 and "a" through "f".

HipsterBrown:
Great work so far @laiedwin! I made a few comments for one of the error messages and some stuff for the error test cases. Could you add some successful password tests as well for the various security options? I'll do a smoke test (trying it out on the hardware) tonight and comment with my results. 

Thanks again for working on this! :clap:

lib/controller.js
```diff
@@ -677,6 +677,32 @@ controller.connectToNetwork = function(opts) {
       if (security && securityOptions.indexOf(security) < 0) {
         return reject(`"${security}" is not a valid security option. Please choose on of the following: ${securityOptions.join(', ')}`);
       }
+      
+      if (security === 'wep') {
+        // WEP passphrases can be 10, 26, or 58 hexadecimal digits long.
+        // Match for hexadecimal characters:
+        const regex = /^[0-9a-f]+$/i;
+        if (!regex.test(password)) {
```
rwaldron: Instead of regex parsing, you could alternatively check if the value is a valid hexadecimal number:

```js
isNaN(`0x${password}`)
```

test/unit/controller.js
```diff
@@ -1065,6 +1065,104 @@ exports['controller.closeTesselConnections'] = {
...
+    controller.createAccessPoint({
+      ssid: 'test',
+      password: 'DEADDEADDEADDEADDEADDEADDEADDEADDEADDEADDEADDEADDEADDEADDEADDEADDEAD',
+      security: 'psk'
+    })
+    .catch(error => {
+      test.ok(error);
+      test.done();
+    });
+  },
```
rwaldron: Great tests!

HipsterBrown:
My Smoke Test:

- Pulled down the PR 
- `npm link .` (link this build of the cli)
`t2 ap --ssid wepNetwork --pass ThisShouldThrow --security wep`

**Result:**
```
Nicks-MacBook-Pro:tessel-test hipsterbrown$ t2 ap --ssid wepNetwork --pass ThisShouldError --security wep
INFO Testing...
INFO Updated Access Point successfully. SSID:  wepNetwork, password ThisShouldError, security mode: wep
```

After seeing the smoke test not work correctly, I checked the changed files again. In the GitHub diff view, I totally missed that the password & security validation was added to `controller.connectToNetwork` instead of `controller.createAccessPoint`, which is what [the tests cover](https://github.com/tessel/t2-cli/pull/792/files#diff-579f2a5c7de143123729e47453faf76bR1072) (not sure why they still passed though). 

Referring back to [the original issue](https://github.com/tessel/t2-cli/issues/760#issuecomment-223842520), the security & password validation should go into `controller.createAccessPoint`. I should have caught that earlier, sorry @laiedwin.

HipsterBrown:
@laiedwin Did you get a chance to read my previous comment yet? No rush, just want to make sure you're aware of why this hasn't been merged yet.

laiedwin:
@HipsterBrown Sorry, I've been a lot busier than usual lately -- I'll try to get around to addressing the later comments this weekend!

HipsterBrown:
@laiedwin No worries at all. Just wanted to check in.

johnnyman727:
@laiedwin how are things going? Can we help move things along in any way?

HipsterBrown:
@laiedwin @johnnyman727 I've opened a PR against this build to help get it over the finish line. Just a few test fixes mostly. 

https://github.com/laiedwin/t2-cli/pull/1

lib/controller.js
```diff
@@ -677,6 +677,31 @@ controller.connectToNetwork = function(opts) {
...
+      if (security === 'psk' || security === 'psk2') {
+        // WPA/WPA2-PSK passphrases can be 8-63 ASCII characters, or 64 hexadecimal digits.
+        // Match ASCII codes for all 127 printable ASCII characters:
+        const asciiRegex = /^[\x00-\x7F]{8,63}$/;
+        // Match exactly 64 hexadecimal digits:
+        const hexTest = !isNaN(`0x${password}`) && password.length === 64;
+        // Reject if both tests fail.
+        if (!asciiRegex.test(password) && !hexTest) {
+          return reject('Invalid passphrase: WPA/WPA2-PSK passkeys must be 8-63 ASCII characters, or 64 hexadecimal digits.');
+        }
```
rwaldron: I'm not going to block on this, but it would be awesome if the comments for each set of criteria included links to canonical "sources of truth".

rwaldron: Specifically, I'd like to see links to the published protocols, in which things like ` if (length !== 10 || length !== 26 || length !== 58) {` are clearly defined. Thanks!!

rwaldron:
Per discussion with @HipsterBrown, we're going to land this and then he'll do a follow up.

Issue #557, opened 2016-02-10
[unit tests] Flaky tests in access-point.js


LinusU:
Ever so often I get this output when running the tests locally. Rerunning fixes it.

```text
Running "nodeunit:tests" (nodeunit) task
Testing access-point.jsFFFFFF
>> Tessel.prototype.createAccessPoint - newAccessPoint
>> TypeError: Cannot read property 'mockClose' of undefined
>> at Object.exports.Tessel.prototype.createAccessPoint.tearDown (test/unit/access-point.js:34:16)

>> Tessel.prototype.createAccessPoint - noPasswordNoSecurity
>> TypeError: Cannot read property 'mockClose' of undefined
>> at Object.exports.Tessel.prototype.createAccessPoint.tearDown (test/unit/access-point.js:34:16)

>> Tessel.prototype.createAccessPoint - properCredentials
>> TypeError: Cannot read property 'mockClose' of undefined
>> at Object.exports.Tessel.prototype.createAccessPoint.tearDown (test/unit/access-point.js:34:16)

>> Tessel.prototype.createAccessPoint - passwordNoSecurity
>> TypeError: Cannot read property 'mockClose' of undefined
>> at Object.exports.Tessel.prototype.createAccessPoint.tearDown (test/unit/access-point.js:34:16)

>> Tessel.prototype.enableAccessPoint - turnsOn
>> TypeError: Cannot read property 'mockClose' of undefined
>> at Object.exports.Tessel.prototype.enableAccessPoint.tearDown (test/unit/access-point.js:221:16)

>> Tessel.prototype.disableAccessPoint - turnsOn
>> TypeError: Cannot read property 'mockClose' of undefined
>> at Object.exports.Tessel.prototype.disableAccessPoint.tearDown (test/unit/access-point.js:267:16)
```

johnnyman727:
Thanks for reporting this. I've also seen this a bunch of times but didn't record it in our issues.

rwaldron:
@HipsterBrown assigning to you. Please investigate and report (either can repro, or cannot)

johnnyman727:
From my experience, this has nothing to do with the access point tests. I believe I saw this issue intermittently before we even had the access point feature. It's pretty hard to reproduce (maybe 1 in 100 test runs will hit it).

johnnyman727:
[...]

PR #795, opened 2016-07-14
Update Rust cross compilation server API to 1.1

johnnyman727:
`result.stderr` from the cross compilation server now has compilation output which was a change between `rustc` 1.5.0 and 1.10.0.

### DIFF
--- a/lib/tessel/deployment/rust.js
+++ b/lib/tessel/deployment/rust.js
@@ -135,14 +135,14 @@ exportables.remoteRustCompilation = (opts) => {
             // If there was, abort with the provided message
             return reject(new Error(result.error));
           }
-          // Check if compilation failed via stderr
+
+          // Print out any stderr output
           else if (result.stderr !== null && result.stderr !== '') {
-            return reject(new Error(result.stderr));
+            log.info(result.stderr);
           }
 
-          // If output from the compilation process was provided
+          // Print out any stdout output
           if (result.stdout !== null) {
-            // Print it out
             log.info(result.stdout);
           }
 
--- a/test/unit/deployment/rust.js
+++ b/test/unit/deployment/rust.js
@@ -164,7 +164,7 @@ exports['deploy.rust'] = {
 
   remoteRustCompilationStderr: function(test) {
 
-    test.expect(3);
+    test.expect(2);
 
     // stderr will be written to when compilation fails
     var stderr = 'Compilation failed: you need to camelcase everything.';
@@ -173,7 +173,7 @@ exports['deploy.rust'] = {
     var result = JSON.stringify({
       stderr: stderr,
       stdout: '',
-      binary: null,
+      binary: new Buffer(0),
     });
 
     // When we get a write to the post request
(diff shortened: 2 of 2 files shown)


Frijol:
LGTM except the failed Appveyor build

lib/tessel/deployment/rust.js
```diff
@@ -122,11 +122,17 @@ exportables.remoteRustCompilation = (opts) => {
         })
         // Reject on failure
         .on('error', function(postReqError) {
+          // Stop the spinner if we error
+          log.spinner.stop();
```
HipsterBrown: I don't think this is needed because rejected or resolved commands from the cli stop the `spinner` -> https://github.com/tessel/t2-cli/blob/master/bin/tessel-2.js#L510-L534

lib/tessel/deployment/rust.js
```diff
@@ -159,6 +165,9 @@ exportables.remoteRustCompilation = (opts) => {
         });
     });
 
+    // Start the spinner to folks know things are happening
+    log.spinner.start();
```
HipsterBrown: The `spinner` is started in `bin/tessel-2.js` -> https://github.com/tessel/t2-cli/blob/master/bin/tessel-2.js#L87

test/unit/access-point.js
```diff
@@ -3,38 +3,44 @@
...
+      // each test exceeds the default maximum number of listeners. This
+      // is only an issue with tests because we re-use the same remote
+      // process simulator across all command calls
+      this.tessel._rps.stdout.setMaxListeners(100);
+      this.tessel._rps.stderr.setMaxListeners(100);
+
+    }
+    catch(err) {
+      console.log('would have failed with error', err);
+    }
```
HipsterBrown: Did you mean to commit this? Or is this to make sure the flaky tests don't fail?

johnnyman727: Yeah, currently making progress on https://github.com/tessel/t2-cli/issues/557#issuecomment-232969903. Kinda stuck on reading through `grunt` source though.

HipsterBrown: Ok so we want to get that flakiness resolved before moving forward with this PR?

I'll try to review and dig into #557 a bit today as well.

johnnyman727: No, it's not a blocker but it's annoying because somehow AppVeyer hits that race condition pretty often and fails perfectly fine PRs/branches.

lib/tessel/deployment/rust.js
johnnyman727: Ah yes! I'll go ahead and remove it.

test/unit/access-point.js
HipsterBrown: Looks like AppVeyer is still failing on some JavaScript Code Styling errors.

johnnyman727: Ah whoops, meant to remove that change. Will fix.

johnnyman727:
I'm going to go ahead and merge this since it was reviewed by two Team Members. AppVeyer fails inconsistently on various versions of Node because of [this unrelated issue](https://github.com/tessel/t2-cli/issues/557#issuecomment-232969903).

PR #798, opened 2016-07-15
Shrinkwraps dependencies so we can secure our own fork of nodeunit

johnnyman727:
Fixes https://github.com/tessel/t2-cli/issues/557#issuecomment-232981469.

See the above issue for commentary on the source of a bug that's causing intermittent test failures.

Basically, there is a bug in one of our well-maintained dependency's unmaintained dependency (`nodeunit`). I have [opened a PR](https://github.com/caolan/nodeunit/pull/317) on `nodeunit` but it's unlikely to be merged and published to `npm`. That means our dependency will be unlikely to pull the new version in to its project.

[`npm shrinkwrap`](https://nodejs.org/en/blog/npm/managing-node-js-dependencies-with-shrinkwrap/) gets around that because it lets us specify where we are pulling those dependencies from. Instead of pulling `nodeunit` from the original repo, this `shrinkwrap.json` pulls it in from [our own fork](https://github.com/tessel/t2-cli/blob/a65db7e5a593a88e2e304548ee561bd815015bc3/npm-shrinkwrap.json#L1691) which already has a fix from the bug.

I'm honestly not sure if this is the most maintainable way to fix the problem but it does seem to work for the short term

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,4669 @@
+{
+  "name": "t2-cli",
+  "version": "0.0.23",
+  "dependencies": {
+    "abbrev": {
+      "version": "1.0.9",
+      "from": "abbrev@>=1.0.0 <2.0.0",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz"
+    },
+    "acorn": {
+      "version": "3.2.0",
+      "from": "acorn@>=3.0.2 <4.0.0",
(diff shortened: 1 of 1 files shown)


johnnyman727:
GAHHH maybe this didn't fix the issue? https://ci.appveyor.com/project/rwaldron/t2-cli/build/1095/job/j657nph6t9b5sjht

rwaldron:
Naw, this isn't necessary. The appropriate fix is to explicitly require our dependencies where they are depended on ;)

Since [I agreed with you here](https://github.com/tessel/t2-cli/issues/557#issuecomment-233056281), and I introduced the bug, I will do the work to update the files.

Issue #776, opened 2016-06-21
Reserved option to disable compression


huseyinkozan:
I have trouble at runtime with apidoc. Few days ago I searched the sources for disabling the compression, but cannot find.

Is it possible to add an option to disable copression for the Node files ?

Here the apidoc error while requesting `http://192.168.2.2/api/vendor/jquery.min.js?v=1466512139991`. It works like this : `http://192.168.2.2/api/vendor/jquery.min.js`, but couldnt find what is adding non-cache stuff : ^ /
```
TypeError: o is not a function
   at o (/app/remote-script/node_modules/qs/lib/parse.js:1:960)
   at i (/app/remote-script/node_modules/qs/lib/parse.js:1:1426)
   at Object.module.exports [as parse] (/app/remote-script/node_modules/qs/lib/parse.js:1:2530)
   at s (/app/remote-script/node_modules/express/lib/utils.js:1:305)
   at /app/remote-script/node_modules/express/lib/middleware/query.js:1:344
   at e.handle_request (/app/remote-script/node_modules/express/lib/router/layer.js:1:867)
   at k (/app/remote-script/node_modules/express/lib/router/index.js:1:2971)
   at /app/remote-script/node_modules/express/lib/router/index.js:1:2608
   at Function.e.process_params (/app/remote-script/node_modules/express/lib/router/index.js:1:3963)
   at n (/app/remote-script/node_modules/express/lib/router/index.js:1:2525)
```

HipsterBrown:
@huseyinkozan Can I ask why a documentation tool like `apidoc` would be running on a Tessel? Could this be done as a pre-deploy step? 

If this is a needed feature, how do you envision it working with the current `t2-cli` deploy workflow? Is it a flag passed on `t2 run` or `t2 push`, i.e. `t2 run index.js --no-compress`?

huseyinkozan:
@HipsterBrown You are right about apidoc deployment. Just packing all in one; api, gui, doc.

`t2 run index.js --no-compress` seems good to me.

rwaldron:
Before I start implementing, I want to know more about this use case. If compression is turned off, it will mean that _all_ of the JS will be deployed without any compression. It's also important to know that the only files that are compressed are those that are discovered via the dependency graph—which means they are `require("foo")` or `require("/path/to/module")`.

huseyinkozan:
Thank you @rwaldron for asking. 

[...]

PR #801, opened 2016-07-18
Make code compression optional, default to —compress=true. Fixes gh-776

rwaldron:
- `t2 run <index.js> --compress=false` will skip the code compression step

@huseyinkozan sorry for the delay on this

@HipsterBrown @carlbaron review?

## Smoke Test

1. Create a new project: `mkdir t2-776 && cd t2-776 && t2 init`
2. Run: `t2 run index.js --compress=false` 
3. Once it's loaded and running, open a separate terminal and run: `t2 root`
4. Once you are SSH'ed to the board, run: `cd /tmp/remote-scripts/ && cat index.js`. The content should be the same as it appears in your local project directory. 

### DIFF
--- a/test/unit/deployment/javascript.js
+++ b/test/unit/deployment/javascript.js
@@ -172,6 +172,7 @@ exports['Deployment: JavaScript'] = {
         // Actually deploy the script
         this.tessel.deploy({
             entryPoint: path.relative(process.cwd(), DEPLOY_FILE_JS),
+            compress: true,
             push: false,
             single: false
           })
@@ -542,6 +543,7 @@ exports['deployment.js.tarBundle'] = {
 
     deployment.js.tarBundle({
       target: path.normalize(target),
+      compress: true,
       full: true,
     }).then(bundle => {
 
@@ -612,6 +614,7 @@ exports['deployment.js.tarBundle'] = {
     deployment.js.tarBundle({
       target: path.normalize(target),
       resolvedEntryPoint: entryPoint,
+      compress: true,
       slim: true,
     }).then(bundle => {
       // These things happen in the --slim path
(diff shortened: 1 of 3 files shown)


test/unit/deployment/javascript.js
```diff
@@ -690,6 +694,34 @@ exports['deployment.js.tarBundle'] = {
...
+    var entryPoint = 'index.js';
+    var target = 'test/unit/fixtures/syntax-error';
+
+    deployment.js.tarBundle({
+      target: path.normalize(target),
+      resolvedEntryPoint: entryPoint,
+      compress: false,
+      slim: true,
+    }).then(bundle => {
+      extract(bundle, (error, entries) => {
```
HipsterBrown: Looks like this is causing an linter error.

HipsterBrown:
The code looks good to me. :+1: I can do a smoke test tonight but I'm fairly confident this is good to go.

rwaldron:
Strange... those lines don't match the code I'm looking at locally...

HipsterBrown:
Smoke test worked as it should. Also opened up a quick PR to this branch to fix that `jshint` error -> https://github.com/rwaldron/t2-cli/pull/2

huseyinkozan:
@rwaldron No problem. 
Thanks for the feature.
Will this disable compression for all *.js files (includes node_modules/**/*.js), or only user files ?

rwaldron:
> Will this disable compression for all .js files (includes node_modules//.js), or only user files ?

Yes, it will _completely_ bypass the compression.

huseyinkozan:
Thanks @rwaldron.

rwaldron:
Thanks for the review, this will be in the release today

Issue #719, opened 2016-04-29
t2 list --lan should dedupe output


boneskull:
When your local machine has *n* network interfaces connected to the same network, each Tessel will show *n* times in the resulting list (in theory):

```shell
$ t2 list --lan
INFO Searching for nearby Tessels...
	LAN	bloath	(USB connect and run `t2 provision` to authorize)
	LAN	bloath	(USB connect and run `t2 provision` to authorize)
```

See the single machine:

```shell
$ arp bloath.local
? (10.0.0.22) at 2:a3:60:d1:47:44 on en4 ifscope [ethernet]
? (10.0.0.22) at 2:a3:60:d1:47:44 on en0 ifscope [ethernet]
```

I'm also wondering if you have your Tessel plugged into ethernet *and* connected via WiFi, it won't show 2*n* times...?  It'd certainly have two different IP's and MAC addresses...

rwaldron:
Yikes, definitely needs to be deduped. 

Start here: https://github.com/tessel/t2-cli/blob/master/lib/controller.js#L93-L107

johnnyman727:
Thanks for bringing this up @boneskull!

PR #805, opened 2016-07-20
Dedupe Tessels found on LAN connections. Fixes gh-719

rwaldron:
The problem here was:

1. Nothing was actually being put into `this.discovered`
2. Even if there had been, the test we did have would've passed anyway because it just put a number in the list and then emitted a number. That would assume that the thing put into `this.discovered` had the same object identity as the thing being checked, which is wrong.

# Smoke Test

This is easier to do with 2 or more Tessels, so I've just taken screen shots to show the result of the changes: 

Before: 

After: 

### DIFF
--- a/lib/lan-connection.js
+++ b/lib/lan-connection.js
@@ -183,17 +183,18 @@ LAN.Scanner.prototype.start = function() {
     });
 
     // When the browser finds a new device
-    this.browser.on('update', (data) => {
+    this.browser.on('update', (discovered) => {
       try {
-        debug(`Device found: ${data.fullname}`);
-        // Check if it's a Tessel
-        if (this.discovered.indexOf(data) > -1) {
+        debug(`Device found: ${discovered.fullname}`);
+
+        // Check against already discovered devices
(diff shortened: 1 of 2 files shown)


rwaldron:
@HipsterBrown @carlbaron

HipsterBrown:
The code looks good to me. Great addition to the tests as well. I can smoke test this tonight if needed. I appreciate the screenshots for previewing the issue and solution.

HipsterBrown:
Smoke test results:

```
Nicks-MacBook-Pro:tessel-test hipsterbrown$ t2 list
 	USB	j5
 	USB	tessel-router
 	LAN	j5
 	LAN	tessel-router
INFO Multiple Tessels found.
INFO Set default Tessel with environment variable (e.g. "export TESSEL=bulbasaur") or use the --name flag.
```

Tried it a couple times and always got the same result. :+1:

carlbaron:
Looks good to me. Have only 1 Tessel so couldn't easily smoke test it.

👍

rwaldron:
Again, I appreciate the reviews and smoke tests where possible :)

Issue #732, opened 2016-05-09
Support passing node options


ebruchez:
For example `--expose-gc`.

This should work with:

- `t2 run`
- `t2 push`

@johnnyman727 confirms that there isn't a way to send Node args at the moment.

PR #806, opened 2016-07-20
run/push: support "subargs" for forwarding to remote process. Fixes gh-732

rwaldron:
Uses the ["subarg"](https://www.npmjs.com/package/subarg) syntax:

```
t2 run index.js [foo bar baz -a 1 -b 2 --named --key=value]
```

Will result in process.argv on the deployed program:

```
[ '/usr/bin/node',
  '/tmp/remote-script/index.js',
  'foo',
  'bar',
  'baz',
  '-a',
  '1',
  '-b',
  '2',
  '--named',
  '--key=value' ]
```

# Smoke Test

(pull this branch locally)

1. Create a new project: 
  ```
  mkdir t2-732 && cd t2-732 && t2 init;
  npm install minimist --save
  ```
2. Open the `index.js` just now and replace the contents with: 
  ```js
  var argv = require('minimist')(process.argv.slice(2));

  console.log(argv);
  ```
3. Run `t2 run index.js [foo bar baz -a 1 -b 2 --named --key=value]`

Deployment should look something like this: 

```
$ t2 run index.js [foo bar baz -a 1 -b 2 --named --key=value]
INFO Connected to ash.
INFO Building project.
INFO Writing project to RAM on ash (9.728 kB)...
INFO Deployed.
INFO Running index.js...
{ _: [ 'foo', 'bar', 'baz' ],
  a: 1,
  b: 2,
  named: true,
  key: 'value' }
```

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -185,6 +185,7 @@ makeCommand('run')
     options.push = false;
     // Overridden in tarBundle if options.full is `true`
     options.slim = true;
+    options.subargs = parser.subargs || [];
 
     callControllerWith('deploy', options);
   })
@@ -241,6 +242,7 @@ makeCommand('push')
     options.push = true;
     // Overridden in tarBundle if options.full is `true`
     options.slim = true;
+    options.subargs = parser.subargs || [];
 
     callControllerWith('deploy', options);
   })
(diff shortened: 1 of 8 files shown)


bin/tessel-2.js
```diff
@@ -516,6 +518,38 @@ makeCommand('root')
...
+      args[i] = arg.slice(1, arg.length);
+      sIndexOfSA = i;
+    }
+
+    if (arg.endsWith(']') && sIndexOfSA !== -1) {
+      // Remove the trailing ']'
+      args[i] = arg.slice(0, arg.length - 1);
+      eIndexOfSA = i;
+    }
+  }
```
rwaldron: It would've been nice to use `subarg` to parse this stuff, but in reality we don't actually want to parse these yet because there is no clear path to reassembling them as the string they will need to be when the remote process is invoked.

rwaldron:
@ebruchez wdyt?

@HipsterBrown review?

test/unit/deploy.js
```diff
@@ -284,7 +284,8 @@ exports['Tessel.prototype.deploy'] = {
     test.expect(12);
     deployTestCode(this.tessel, {
       push: true,
-      single: false
+      single: false,
+      subargs: [],
```
HipsterBrown: This should pass without the `subargs` property as well, right?

HipsterBrown:
Code looks good to me. I can't do a smoke test tonight. 

A few questions:

- Would you suggest using subargs to pass through things like API tokens instead of environment variables? 
- Could we also pass through environment variables in a similar way?

HipsterBrown:
Smoke test results:

```
Nicks-MacBook-Pro:tessel-test hipsterbrown$ mkdir t2-732 && cd t2-732 && t2 init; npm install minimist --save
INFO Initializing new Tessel project for JavaScript...
INFO Created package.json.
INFO Created .tesselinclude.
INFO Wrote "Hello World" to index.js
minimist@1.2.0 node_modules/minimist
Nicks-MacBook-Pro:t2-732 hipsterbrown$ vim index.js
Nicks-MacBook-Pro:t2-732 hipsterbrown$ t2 run index.js [foo bar baz -a 1 -b 2 --named --key=value]
INFO Which Tessel do want to use? 	LAN	tessel-router
INFO Building project.
INFO Writing project to RAM on tessel-router (9.728 kB)...
INFO Deployed.
INFO Running index.js...
{ _: [ 'foo', 'bar', 'baz' ],
  a: 1,
  b: 2,
  named: true,
  key: 'value' }
Nicks-MacBook-Pro:t2-732 hipsterbrown$
```

Looks like a successful test! :+1:

test/unit/deploy.js
rwaldron: No, because this test runs code that exists at a point in which all default properties are resolved, so the `subargs` property will always be present, either empty or with contents

rwaldron:
> Could we also pass through environment variables in a similar way?

TBH, I'm not sure and this is an interesting question. I will do some tests and report back

rwaldron:
> Could we also pass through environment variables in a similar way?

I've spent some time with this question and have determined that this patch is not an appropriate mechanism to shoe-horn other features into. I'm going to stop at "just subargs" here.

HipsterBrown:
@rwaldron Sounds good. I appreciate you taking the time to investigate that concept. LGTM :+1:

Issue #820, opened 2016-08-19
preferences fails write when .tessel folder doesn't exist


jandrieu:
I ran into this trying to get the t2 working on WIndows 10, node 4.3, following this gist https://gist.github.com/tcr/992978a5dbe5bff2e18f495c5c0973c3

Unfortunately, the preferences file handling fails if there is no ~/.tessel folder.

```
C:\Users\Joe\Documents\Software\Tessel>t2 list
INFO Searching for nearby Tessels...
ERR! Detected CLI crash { [Error: LIBUSB_TRANSFER_STALL] errno: 4 } Error: LIBUSB_TRANSFER_STALL
ERR!     at Error (native)
undefined
Submit Crash Report to help Tessel Developers improve the CLI ?
    If yes(y), subsequent crashes will be submitted automatically. Yes
ERR! Error writing preference crash.reporter.prompt false
ERR! Detected CLI crash { [Error: ENOENT: no such file or directory, open 'C:\Users\Joe\.tessel\preferences.json']
ERR!   errno: -4058,
ERR!   code: 'ENOENT',
ERR!   syscall: 'open',
ERR!   path: 'C:\\Users\\Joe\\.tessel\\preferences.json' } Error: ENOENT: no such file or directory, open 'C:\Users\Joe\.tessel\preferences.json'
ERR!     at Error (native)
undefined
Submit Crash Report to help Tessel Developers improve the CLI ?
```

rwaldron:
Closed by https://github.com/tessel/t2-cli/issues/821

PR #821, opened 2016-08-19
fs.ensureFile to fix missing .tessel dir in Preferences.write

jandrieu:
Fix for issue #820


rwaldron:
Thanks for the patch. I added a test to cover the new behavior

PR #834, opened 2016-08-23
t2 crash-reporter: adds argv (sans first two entries)

rwaldron:
@tikurahul what do you think about this? There will be a new `argv` property that gets posted to the crash report server. We can try it out for a bit before I write a bunch of tests.

### DIFF
--- a/lib/crash-reporter.js
+++ b/lib/crash-reporter.js
@@ -107,6 +107,10 @@ CrashReporter.submit = function(report, opts) {
   }
 
   const silent = (opts.silent || false);
+  // Removes:
+  // - The path to node
+  // - The path to t2
+  const argv = process.argv.slice(2).join(', ');
 
   return CrashReporter.prompt()
     .then(success => {
@@ -124,7 +128,7 @@ CrashReporter.submit = function(report, opts) {
 
               var stack = CrashReporter.sanitize(report.stack || String(report));
 
-              return CrashReporter.post(labels, stack)
+              return CrashReporter.post(labels, stack, argv)
                 .then(fingerprint => {
                   if (!silent) {
                     log.info(`Crash Reported: ${CRASH_REPORTER_BASE_URL}/crashes?fingerprint=${fingerprint}`);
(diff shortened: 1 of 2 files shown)


tikurahul:
@rwaldron This looks good to me. I can add support for `argv` in the server side. I will add it to the core model and also make it searchable in the search index. I will mark `argv` parameter as optional in the REST API.

rwaldron:
Sounds like a plan!

lib/crash-reporter.js
```diff
@@ -142,14 +146,18 @@ CrashReporter.submit = function(report, opts) {
...
+
   return new Promise((resolve, reject) => {
     request.post({
-      url: SUBMIT_CRASH_URL,
+      url,
       form: {
-        crash: report,
-        labels: labels,
-        f: 'json'
+        argv,
```
tikurahul: Can you do an `argv.join(',')` to make it consistent with labels.

lib/crash-reporter.js
```diff
@@ -107,6 +107,10 @@ CrashReporter.submit = function(report, opts) {
   }
 
   const silent = (opts.silent || false);
+  // Removes:
+  // - The path to node
+  // - The path to t2
+  const argv = process.argv.slice(2).join(', ');
```
rwaldron: @tikurahul updated here

test/unit/crash-reporter.js
```diff
@@ -206,6 +206,28 @@ exports['CrashReporter.submit'] = {
...
+    }).then(() => {
+      test.equal(this.crPost.callCount, 1);
+      test.equal(this.crPost.lastCall.args[2], 'foo');
+      process.argv[2] = original;
+      test.done();
+    }).catch(error => {
+      test.ok(false, error.message);
+      test.done();
+    });
+  },
```
rwaldron: @HipsterBrown @johnnyman727 test added.

tikurahul:
:+1: LGTM

rwaldron:
@tikurahul thanks!

PR #837, opened 2016-08-26
t2 restore: progress bars for all write operations

rwaldron:
Needs tests, but wondering what folks think. The result is pretty nice: 

### DIFF
--- a/lib/update-fetch.js
+++ b/lib/update-fetch.js
@@ -169,7 +169,7 @@ exportables.downloadTgz = function(tgzUrl, fileMap) {
       var contentLength = parseInt(res.headers['content-length'], 10);
 
       // Create a new progress bar
-      var bar = new Progress('  Downloading [:bar] :percent :etas remaining', {
+      var bar = new Progress('     [:bar] :percent :etas remaining', {
         clear: true,
         complete: '=',
         incomplete: ' ',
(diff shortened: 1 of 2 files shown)


johnnyman727:
I think it's a great UI improvement! If only we could get this for deploys as well :/

HipsterBrown:
Love it. :+1: 

Definitely agree with @johnnyman727, it would be nice to get a clearer idea of how long program deploys take to run eventually. This is a good starting point.

rwaldron:
Thanks. I've updated the display a bit, with new images in the PR above. Please re-review

lib/update-fetch.js
```diff
@@ -127,7 +127,7 @@ exportables.fetchBuild = function(build) {
 
 exportables.downloadTgz = function(tgzUrl, fileMap) {
   return new Promise((resolve, reject) => {
-    log.info('Downloading files. This may take a few minutes...');
+    log.info('Downloading files (this may take a few minutes)...');
```
johnnyman727: If you deleted the `(this may take a few minutes)` in the squashfs write (which I think is fair now that there is a progress bar), you might as well delete it here too.

HipsterBrown: Do we still need "(this may take a few minutes)" here if it was removed in the other file?

HipsterBrown: @johnnyman727 I think we left the same comment at the same time. :P

PR #845, opened 2016-09-01
Adhoc network support

HipsterBrown:
This build adds an `adhoc` command, which acts as an alias for `ap` because the `adhoc` mode uses the same `wifi-face` and changes the `mode` property to "adhoc". 

After installing this branch, run `t2 adhoc --help` to get more information about using the command. 

**To test the command:**
Works best with 2 Tessels running an adhoc network. Instructions will be the same for both Tessels.

- install this branch locally
- run `t2 adhoc --ssid TestAdHoc --name <name of one tessel>`
- run `t2 adhoc --ssid TestAdHoc --name <name of other tessel>`
- both Tessels should have their amber LEDs on

You can make sure the adhoc connection works using `ssh` or `screen` to access one of the Tessel terminals and running `ping 192.168.1.101` and see successful data returns. 192.168.1.101 is the static IP address of the Tessel `lan` network.

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -482,6 +482,46 @@ makeCommand('version')
 
 makeCommand('ap')
   .callback(options => {
+    options.mode = 'ap';
+    log.level(options.loglevel);
+
+    if (options.on || options.off) {
+      if (options.on) {
+        callControllerWith('enableAccessPoint', options);
+      } else {
+        callControllerWith('disableAccessPoint', options);
+      }
+    } else if (options.ssid) {
+      callControllerWith('createAccessPoint', options);
+    } else {
+      callControllerWith('getAccessPointInfo', options);
+    }
+  })
+  .option('ssid', {
+    abbr: 'n',
+    help: 'Name of the network.'
+  })
+  .option('password', {
+    abbr: 'p',
+    help: 'Password to access network.'
+  })
+  .option('security', {
+    abbr: 's',
+    help: 'Encryption to use on network (i.e. wep, psk, psk2, wpa, wpa2).'
+  })
+  .option('off', {
+    flag: true,
+    help: 'Disable the access point'
+  })
+  .option('on', {
+    flag: true,
+    help: 'Enable the access point'
+  })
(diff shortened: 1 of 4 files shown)


Frijol:
Testing.

* [ ] Connected both to existing local internet, then ran test above. Doesn't work with wifi on, need handling or error.
* [ ] Disabled Wifi on both and tried again, worked great

Comments

* [ ] Would be nice to give more info than "Configure the Tessel as an adhoc network" in the `.help` – I'm not sure what that really means
* [ ] Not clear what it's good for - feels like it needs more builtin functionality to be useful/discoverable
* [ ] Not clear how to map IPs to devices
* [ ] Help output doesn't quite match Wifi options help output.

Wifi help output:

```sh
   -n SSID, --ssid SSID               Set the SSID of the network to connect to
   -p PASSWORD, --password PASSWORD   Set the password of the network to connect to
```

Adhoc help output:

```sh
   -p, --password                    Password to access network.
```

Overall, functions as expected except for needing to turn Wifi off first

Frijol:
Re mapping IP addresses, I guess you could run `t2 adhoc` on each connected t2 in order to print the IP address.

HipsterBrown:
I've updated this PR with the ability to set an IP address for both `t2 ap` and `t2 adhoc` commands.

@Frijol 
> Help output doesn't quite match Wifi options help output.

`adhoc` networking is kind of a combination of wifi and access points because devices in the adhoc network can connect to it, while also emitting the network. So I've added validation for `connectToNetwork` and `createAccessPoint` (which is used by the `adhoc` command as well) to make sure wifi is not or cannot be enabled at the same time as an adhoc network.

arushi019:
Hi @Frijol @HipsterBrown ! How should we go about this?

Frijol:
@HipsterBrown re @arushi019's comment above, where are we in this PR? What would you recommend as next steps?

HipsterBrown:
@Frijol @arushi019 I think we can start by pulling the flag for setting ip of the `ap` network into a separate PR. I need to do some more research into the usability of the adhoc networking before adding it to the CLI and docs.

PR #847, opened 2016-09-02
one-line badges

Frijol:
I think this is more standard. Ok by you @tcr or @johnnyman727?


johnnyman727:
LGTM. Don't worry about the Appveyer complaints.

PR #967, opened 2016-09-09
Minor fixes to Rust deployment.

tcr:
* 	Fixes undefined variable when running t2 run Cargo.toml
* 	Updates Rust init script to support tessel 0.2.0 crate.
* Adds alloc_system directive to rust initialization script.

### DIFF
--- a/lib/tessel/deployment/rust.js
+++ b/lib/tessel/deployment/rust.js
@@ -28,7 +28,7 @@ var exportables = {
       var cargoToml = toml.parse(fs.readFileSync(path.join(pushdir, 'Cargo.toml'), 'utf8'));
 
       if (cargoToml.package) {
-        basename = path.basename(program);
+        basename = pushdir;
         program = cargoToml.package.name;
       }
 
--- a/resources/rust/main.rs
+++ b/resources/rust/main.rs
@@ -1,5 +1,10 @@
 //! A blinky example for Tessel
 
+// This enables the system allocator in your script.
+// This is required to run Rust binaries on Tessel.
+#![feature(alloc_system)]
+extern crate alloc_system;
+
 // Import the tessel library
 extern crate tessel;
 // Import the Tessel API
--- a/resources/rust/main.rs
+++ b/resources/rust/main.rs
@@ -1,9 +1,9 @@
 //! A blinky example for Tessel
 
 // Import the tessel library
-extern crate rust_tessel;
+extern crate tessel;
 // Import the Tessel API
-use rust_tessel::Tessel;
+use tessel::Tessel;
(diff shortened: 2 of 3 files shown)


tcr:
Also see <https://github.com/tessel/rust-compilation-server/pull/12>.

lib/tessel/deployment/rust.js
```diff
@@ -28,7 +28,7 @@ var exportables = {
       var cargoToml = toml.parse(fs.readFileSync(path.join(pushdir, 'Cargo.toml'), 'utf8'));
 
       if (cargoToml.package) {
-        basename = path.basename(program);
+        basename = pushdir;
```
johnnyman727: Out of curiosity, was this field used anywhere?

johnnyman727:
LGTM!

lib/tessel/deployment/rust.js
tcr: I have no idea! :speak

johnnyman727: :+1: I think I remember that property not being defined but didn't fix it because it wasn't breaking anything and then just forgot. 🙈

rwaldron:
@johnnyman727 @tcr needs tests ;)

lib/tessel/deployment/rust.js
rwaldron: The "package" field is in our own rust init generated Cargo.toml... IIRC, this path is what allows for deployment from the root of a project, but where the program code is in a subdirectory, eg. `t2 run src/whatever.rs`. If that's correct, then the change here would break that—someone should smoke test. Also, this change should have its own tests.

rwaldron: I am unable to reproduce any failures with `t2 run Cargo.toml`, as noted in: 

> Fixes undefined variable when running t2 run Cargo.toml

resources/rust/main.rs
```diff
@@ -1,9 +1,14 @@
 //! A blinky example for Tessel
 
+// This enables the system allocator in your script.
+// This is required to run Rust binaries on Tessel.
+#![feature(alloc_system)]
+extern crate alloc_system;
```
johnnyman727: @tcr just FYI, this breaks remote code deployment because the cross compilation server is running Rust `v1.10.0` but not nightly.

tcr: Oh.

johnnyman727: In the [process of fixing on the build server](https://github.com/tessel/rust-compilation-server/pull/14).

johnnyman727: Okay, this is fixed!

Issue #972, opened 2016-09-12
radio0 is disabled


tcr:
This happened on two fully updated Tessels (0.0.14), so I'm not sure what's happening here, but...

`t2 wifi -n ...` was complaining my credentials were wrong. That message was incorrect so I dtermed in:

```
$ wifi
'radio0' is disabled
'radio0' is disabled
```

It warned me radio0 was disabled. I did `uci show wireless` and found this:

```
uci show wireless
wireless.radio0=wifi-device
wireless.radio0.type='mac80211'
wireless.radio0.channel='11'
wireless.radio0.hwmode='11g'
wireless.radio0.path='10180000.wmac'
wireless.radio0.htmode='HT20'
wireless.radio0.disabled='1'
wireless.@wifi-iface[0]=wifi-iface
wireless.@wifi-iface[0].device='radio0'
wireless.@wifi-iface[0].network='wifi'
wireless.@wifi-iface[0].mode='sta'
wireless.@wifi-iface[0].encryption='psk2'
wireless.@wifi-iface[0].key='...'
wireless.@wifi-iface[0].disabled='0'
wireless.@wifi-iface[0].ssid='...'
```

So radio0 was disabled? I then ran:

```
uci set wireless.radio0.disabled=0
```

And my wifi just turned on.

HipsterBrown:
So I'm not sure how the wireless config got into that state because the cli should have stopped disabling the wifi radio after `ap` functionality was added. However, just in case this happens to anyone else, I've created a fix that re-enables `radio0` if that message occurs. https://github.com/tessel/t2-cli/pull/987

tcr:
@HipsterBrown 👏

PR #987, opened 2016-09-20
Check if wireless.radio0 is disabled

HipsterBrown:
Fixes https://github.com/tessel/t2-cli/issues/972

Smoke test:

The only way I've figured out how to recreate this issue is to manually disable the `radio0` wireless interface. I would suggest doing this over USB using `dterm` or `screen` because an `ssh` session will cut off as soon as the radio is disabled, i.e. `screen /dev/tty.usb<tab complete from here>` and hit the "enter" key twice.

Once rooted into your Tessel, run the following commands:

```
uci set wireless.radio0.disabled='1'
uci commit wireless
```

While still rooted, running `wifi` should give you the error `'radio0' is disabled`. To leave the `screen` session, hit Control+a+\ then type "y".

Now that the wireless radio has been disabled:

1. Check out this branch
2. `npm link .` (if you haven't linked this locally yet)
3. `t2 wifi -n "Your Network Name" -p PasswordIfNeeded123`

You should get the following output:

```
hipsterbrown:t2-cli (handle-wifi-radio) $ t2 wifi -n "My Network Name" -p MyPassword
INFO Wifi Enabled.
INFO Wifi Connected. SSID: My Network Name, password: MyPassword, security: psk2
```

Let me know if there are any questions.

### DIFF
--- a/lib/tessel/wifi.js
+++ b/lib/tessel/wifi.js
@@ -163,7 +163,18 @@ Tessel.prototype.setWiFiState = function(enable) {
   return new Promise((resolve, reject) => {
     return this.simpleExec(commands.turnOnWifi(enable))
       .then(() => this.simpleExec(commands.commitWirelessCredentials()))
-      .then(() => this.simpleExec(commands.reconnectWifi()))
+      .then(() => {
+        return this.simpleExec(commands.reconnectWifi())
+          .then((result) => {
+            // check if the actual wifi radio (wireless.radio0) is disabled
+            if (result.includes("'radio0' is disabled")) {
+              return this.simpleExec(commands.turnRadioOn())
+                .then(() => this.simpleExec(commands.commitWirelessCredentials()))
+                .then(() => this.simpleExec(commands.reconnectWifi()))
+                .catch(reject);
+            }
+          });
+      })
       .then(() => {
         var settle = (rejection) => {
           if (rejection) {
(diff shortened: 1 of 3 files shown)


lib/tessel/wifi.js
```diff
@@ -163,7 +163,18 @@ Tessel.prototype.setWiFiState = function(enable) {
...
+      .then(() => {
+        return this.simpleExec(commands.reconnectWifi())
+          .then((result) => {
+            // check if the actual wifi radio (wireless.radio0) is disabled
+            if (result.includes("'radio0' is disabled")) {
+              return this.simpleExec(commands.turnRadioOn())
+                .then(() => this.simpleExec(commands.commitWirelessCredentials()))
+                .then(() => this.simpleExec(commands.reconnectWifi()))
+                .catch(reject);
+            }
```
rwaldron: What happens if there is an error that doesnt include `"'radio0' is disabled"`? I think this needs to return `Promise.resolve()`?

rwaldron: What happens if there is an error that doesnt include `"'radio0' is disabled"`? I think this needs to return `Promise.resolve()`?


(Sorry, I reposted this because I'm not sure if we're officially using Github Review yet)

HipsterBrown: I believe, and after experimenting, once the execution of a `thenable` function finishes, the Promise will continue to the next chained method. Returning in a `thenable` function will pass that value to the next `thenable`.

HipsterBrown:
@rwaldron Pushed a test to confirm your comment.

lib/tessel/wifi.js
rwaldron: Keeping track of promises is not my cup of tea. Thanks for clarifying

PR #996, opened 2016-09-23
better phrasing of t2 reboot

dbuentello:
This is more in line with the rest of `t2 -h`:
```sh

Usage: t2 <command>

command     
  install-homedir     Install homedir: ~/.tessel
  crash-reporter      Configure the Crash Reporter.
  restore             Restore your Tessel by installing the factory version of OpenWrt.
  restart             Restart a previously deployed script in RAM or Flash memory (does not rebundle)
  reboot              Reboot your Tessel
  run                 

```

### DIFF
--- a/bin/tessel-2.js
+++ b/bin/tessel-2.js
@@ -195,7 +195,7 @@ makeCommand('reboot')
 
     callControllerWith('reboot', options);
   })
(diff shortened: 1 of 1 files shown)


johnnyman727:
LGTM

Issue #1012, opened 2016-10-05
t2 run/push: subargs issues


rwaldron:
- with only one arg will retain leading [
- with unnecessary whitespace will create empty strings

PR #1013, opened 2016-10-05
t2 run/push: subargs issues. Fixes gh-1012

rwaldron:
Fixes:

- with only one arg will retain leading [
- with unnecessary whitespace will create empty strings

cc @HipsterBrown 


rwaldron:
The failures are consistently: 

- only on travis 
- only on osx
- completely bogus

rwaldron:
Here is evidence that the change set does not cause the failed tests: https://travis-ci.org/tessel/t2-cli/builds/165619948

PR #1014, opened 2016-10-05
Adds SDK installation and Rust local compilation.

tcr:
Adds the following capabilities:

* `t2 sdk install` — installs T2 SDK and rustlib.
* `t2 sdk remove` — removes the above.
* `t2 run Cargo.toml` — locally compiles a Cargo crate and deploys on Tessel.

## Usage

Use this branch of `t2-cli`.

In a new directory, run `t2 init --lang=rust`.

Run `t2 sdk install`. If it errors out, make sure you're using rust stable v1.11.0 or v1.12.0. If you have `rustup` on your machine, run `rustup override set stable`.

Run `t2 run Cargo.toml`. It should output the following and start blinking:

```
$ t2 run Cargo.toml
INFO Connected to Tessel-02A34C9AB2FE.
    Updating registry `https://github.com/rust-lang/crates.io-index`
   Compiling libc v0.2.16
   Compiling cfg-if v0.1.0
   Compiling unix_socket v0.5.0
   Compiling tessel v0.2.0
   Compiling blinky v0.0.1 (file:///Users/tim/tcr/test/rust-blinky)
    Finished release [optimized] target(s) in 3.38 secs
INFO Writing project to RAM on Tessel-02A34C9AB2FE (338.432 kB)...
INFO Deployed.
INFO Running blinky...
```

## Testing

Setup: run `cargo-tessel sdk uninstall` to reset your local context.

**NOTE:** The remote server *compilation* may fail until this is updated. Interpret a lot of `__stack_chk_guard` errors as `--rustcc` working properly.

* Run `t2 init --lang=rust` in a new directory to get a blinky example.
* Run `t2 run blinky` when `rustc` is not available. You should be informed A) you should install `rustc` and Cargo and B) you can run `t2 run Cargo.toml --rustcc` to use the Rust Cross Compiler. We exit with an error code.
* Run `t2 run blinky --rustcc` when `rustc` is not available. The remote compilation server should be used and succeed, and you will see the blinky code run on Tessel. (You can `mv /Users/tim/.cargo/bin/rustc` to some other path to simulate rustc not being installed. Make sure to move it back after.)
* Run `t2 run blinky` when `rustc` *is* available. You should be informed A) **the Rust SDK needs to be installed to compile locally** and B) you can run `t2 run Cargo.toml --rustcc` to use the Rust Cross Compiler. We exit with an error code.
* Run `cargo tessel sdk install` when your `rustc` version is nightly or < 1.11.0. It should install the Tessel build tools. (If you have `rustup`, run `rustup override set nightly` in your blinky dir first). It should fail to install the MIPS libstd with a warning to install 1.11.0 or greater stable. We exit with an error code.
* Run `t2 run blinky`. You should be informed A) **the MIPS libstd needs to be installed to compile locally** and B) you can run `t2 run Cargo.toml --rustcc` to use the Rust Cross Compiler. We exit with an error code.
* Run `cargo tessel sdk install` when your `rustc` version is >= 1.11.0 and is stable.  (If you have rustup, run `rustup override set stable` in your blinky dir first). It should install the Tessel build tools and the MIPS libstd successfully.
* Run `t2 run blinky`. The compilation should be successful, using the locally installed SDK and cargo to compile the blinky code, then run on Tessel.
* Run `t2 run blinky --rustcc`. The compilation should be successful, using the remote server to compile the blinky code, then run on Tessel.


bin/tessel-2.js
```diff
@@ -541,6 +541,24 @@ makeCommand('root')
...
+  .callback(options => {
+    log.level(options.loglevel);
+
+    if (options.subcommand === 'install') {
+      options.operation = 'installSdk';
+      callControllerWith('installer', options);
+    } else if (options.subcommand === 'remove') {
+      options.operation = 'removeSdk';
+      callControllerWith('installer', options);
+    }
```
rwaldron: What happens if `options.subcommand` is undefined?

bin/tessel-2.js
```diff
@@ -541,6 +541,24 @@ makeCommand('root')
...
+  })
+  .option('subcommand', {
+    position: 1,
+    required: true,
```
rwaldron: re: `options.subcommand` being undefined... this didn't register on first read.

bin/tessel-2.js
```diff
@@ -541,6 +541,24 @@ makeCommand('root')
...
+    choices: ['install', 'remove']
+  })
+  .help('SDK management');
```
rwaldron: Needs more info. SDK for what?

rwaldron:
- [ ] Create a new `install` directory under `lib/tessel/` (ie. `lib/tessel/install/`) and move the file presently called `sdk.js` to `lib/tessel/install/rust.js`

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+var Progress = require('progress');
+var Promise = require('bluebird');
+var request = require('request');
+var spawn = require('child_process').spawn;
+var tar = require('tar-fs');
+var tmp = require('tmp');
+var Transform = require('stream').Transform;
+var zlib = require('zlib');
+
+var log = require('./log');
```
rwaldron: Please organize these as they are organized throughout the code base

lib/installer.js
```diff
@@ -73,3 +76,25 @@ module.exports.homedir = function() {
...
+      log.error(e.message);
+      log.warn('Continuing with SDK installation...');
+    }))
+    .then(() => {
+      log.info('SDK installed.');
+    });
+};
+
+module.exports.removeSdk = function() {
+  return new Promise((resolve, reject) => { // jshint ignore:line
```
rwaldron: Is this ignoring the unused `reject`? Just remove it ;)

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
+var blocks = require('block-stream2');
+var bz2 = require('unbzip2-stream');
+var createHash = require('sha.js');
+var fs = require('fs-extra-promise');
+var osenv = require('osenv');
+var path = require('path');
+var Progress = require('progress');
+var Promise = require('bluebird');
```
rwaldron: Remove this, there is no reason to pave over the built-in Promise class

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
+var blocks = require('block-stream2');
+var bz2 = require('unbzip2-stream');
+var createHash = require('sha.js');
+var fs = require('fs-extra-promise');
```
rwaldron: t2-cli is using `fs-extra` exclusively everywhere else—why does this file need to have a different thing?

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+var osenv = require('osenv');
+var path = require('path');
+var Progress = require('progress');
+var Promise = require('bluebird');
+var request = require('request');
+var spawn = require('child_process').spawn;
+var tar = require('tar-fs');
```
rwaldron: Please use `var tar = require('tar');`

package.json
```diff
@@ -48,10 +48,13 @@
     "acorn": "^3.0.2",
     "async": "^1.4.2",
     "bindings": "^1.2.1",
+    "block-stream2": "^1.1.0",
+    "bluebird": "^3.4.6",
```
rwaldron: Remove

package.json
```diff
@@ -48,10 +48,13 @@
...
     "char-spinner": "^1.0.1",
     "colors": "^1.1.0",
     "common-tags": "0.0.1",
     "fs-extra": "^0.26.2",
+    "fs-extra-promise": "^0.4.1",
```
rwaldron: Remove

package.json
```diff
@@ -63,21 +66,25 @@
...
     "shell-escape": "^0.2.0",
     "ssh2": "^0.4.2",
     "sshpk": "^1.6.0",
     "stream-to-buffer": "^0.1.0",
-    "npmlog": "^4.0.0",
     "t2-project": "^0.4.0",
     "tar": "^2.1.1",
+    "tar-fs": "^1.13.2",
     "tar-stream": "^1.3.0",
+    "tmp": "0.0.29",
```
rwaldron: Remove

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+};
+
+var paths = {
+  sdk: path.join(osenv.home(), '.tessel/sdk'),
+  rustlib: path.join(osenv.home(), '.tessel/rust'),
+};
+
+var sdkPath = {
+  'macos': 'https://builds.tessel.io/t2/sdk/t2-sdk-macos-x86_64.tar.bz2',
+  'linux': 'https://builds.tessel.io/t2/sdk/t2-sdk-linux-x86_64.tar.bz2',
```
rwaldron: Keys don't need quoting

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+};
+
+var rustlibUrl = 'https://builds.tessel.io/t2/sdk/t2-rustlib-VERSION.tar.gz';
```
rwaldron: Use const casing, eg. `RUST_LIB_TGZ_URL` (eg. see `RESTORE_TGZ_URL` in update-fetch.js)

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+    callback();
+  };
+  stream.on('finish', function() {
+    stream.emit('sha256', sha256.digest('hex'));
+  });
+  return stream;
+}
+
+function sha256file(hash, name) {
+  return hash + '  ' + name + '\n';
```
rwaldron: Use template string

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+    });
+  });
+
+  return req;
+}
+
+function downloadString(url) {
+  return new Promise((resolve, reject) => {
+    request({
+      url: url,
```
rwaldron: Use property shorthand wherever possible (prevents pointless renaming): 

```diff
+    request({
+      url,
```

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+        reject(err);
+      } else {
+        resolve({
+          path: tmppath,
+          cleanup: cleanup,
+        });
+      }
+    });
+  });
+};
```
rwaldron: Why is this exported? (also, why is everything exported on `exports` and not `module.exports`, per the project convention?

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+      if (check.exists && check.checked) {
+        log.info(`Latest ${pkgname} already installed.`);
+        return;
+      } else if (!check.exists) {
+        log.info(`Installing ${pkgname}...`);
+      } else {
+        log.info(`Updating ${pkgname}...`);
+      }
+
+      return fs.mkdirpAsync(paths.rustlib)
```
rwaldron: Eventually t2-cli can be refactored to use fs-extra-promise, but until it's done across the entire project, all fs operations should use fs-extra. (This PR is not an appropriate vehicle for such changes)

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+            // Write out CHECKSUM file.
+            fs.writeFileSync(path.join(destdir.path, 'CHECKSUM'), checksum);
+
+            // Remove the old SDK directory.
+            fs.removeAsync(root)
+              .then(() => {
+                // Move temporary directory to target destination.
+                return fs.moveAsync(destdir.path, root);
+              })
+              .finally(() => {
```
rwaldron: t2-cli should use https://www.npmjs.com/package/promise.prototype.finally (which is based on actual proposal)

rwaldron: Shim it to the built-in Promise.prototype: 

```js
require('promise.prototype.finally').shim();
```

tcr: This can just be removed—this is the only instance of finally.

lib/sdk.js
```diff
@@ -0,0 +1,414 @@
...
+  });
+};
+
+exports.buildTessel = (config) => {
+  var env = Object.assign({}, process.env);
+  Object.assign(env, {
+    STAGING_DIR: config.stagingDir,
+    RUST_TARGET_PATH: config.rustlibPath,
+    PATH: path.join(config.toolchainPath, 'bin') + ':' + env.PATH,
+    RUSTFLAGS: '-L ' + config.rustlibPath,
```
rwaldron: Use template strings

tcr:
Addressed your early feedback, thanks @rwaldron. (I'm tempted to submit a PR to drop in fs-extra-promise for fs-extra, but I'll just do it the sadder way)

---

Here's the issue with `tar-fs`: the `tar` module has issues with the SDK bundle. This bundle is auto-generated by the OpenWRT build toolchain and uploaded unmodified to our servers. When we ask to unpack it, it gets confused on symlinks and chokes. I can dig further into this (perhaps it's trying to unpack symlinks as hardlinks, or the header parsing is weird, or something else), but my thoughts are:

* `tar-fs` works with the SDK bundle with no modification
* `tar-fs` claims and benchmarks faster than `tar` does, which is important for the 200mb SDK file.
* `tar-fs` is built on `tar-stream`, which we also include in the project; it might make sense to switch over to `tar-fs` for other operations. 

In any case, what approach should I take here?

rwaldron:
> Should we fallback to the Rust compilation server if the SDK isn't installed?

Since the machinery is already in place, I think we should. I want @johnnyman727's input as well.

rwaldron:
w/r to tar-fs: just leave it in your code for now and we can revisit a broader refactoring later.

tcr:
Spitballing a proposal for rustcc:

* User runs `t2 run Cargo.toml`. It informs the error that the Rust SDK needs to be installed to compile. It also suggests using `t2 run Cargo.toml --rustcc` to use the remote Rust Cross Compiler. It exits with an error code. This follows rule of "no surprises".
* User wants to use a custom cross-compilation server. User runs `t2 run Cargo.toml --rustcc=127.0.0.1:14450`.
* User wants to locally compile. Gets error in `t2 run Cargo.toml` to use the SDK. User runs `t2 sdk install`. User runs `t2 run Cargo.toml`, code succeeds.

I want to see how use of the cross-compilation server pans out over the next few months; I'm weary of having it on by "default" but behind an obvious argument flag, with good error messages, will allow users to quickly get started at hackathons and allow Windows users to use the Rust server as well, while letting users who want local cross-compilation know explicitly what the `t2 run` command will do at all times.

rwaldron:
Curious: why was the command changed from `t2 run src/main.rs` to `t2 run Cargo.toml`? That seems on par with something like `t2 run package.json`, which is definitely not allowed.

tcr:
@rwaldron While I didn't make that change, I see three options here:

1. Specify the Cargo.toml file. As you mention, it's not symmetrical with package.json, though it does work (and the CLI could assume `t2 run [Cargo.toml]` for just `t2 run`.)
1. Specify the binary name. Rust binaries can live at `src/main.rs`, `src/bin/*.rs`, or be specified using `[[bin]]` blocks in the Cargo.toml file. Each of these has a interpreted name, seen by `cargo metadata`; `src/main.rs` would be the name of the Cargo crate itself, for example. Again we can assume `t2 run` runs the default binary if nothing is specified. Right now, we have to specify `t2 run Cargo.toml --bin name-of-bin` regardless, if there are more than one binaries. It might make sense to just specify *only* the binary name, imo.
1. Specify the path to the Rust file itself. This is very unnatural for Rust / Cargo code, most would use the bin name directly, so I'd advocate some solution which doesn't do this. Especially since there's not always a logical mapping between rust filename and output binary.

rwaldron:
Ok, `t2 install x` is merged, so you'll need to refactor that portion of code in this branch. As part of proving out the use case, I framed the `t2 install rust-sdk` and `t2 uninstall rust-sdk`—but they don't do anything but return a resolved Promise.

rwaldron:
Ping me when ready for next review

bin/tessel-2.js
```diff
@@ -541,6 +541,27 @@ makeCommand('root')
...
+      // Unreachable: this case should not be allowed by nomnom.
+      throw new Error('Expected subcommand "t2 sdk install" or "t2 sdk remove".');
+    }
+  })
+  .option('subcommand', {
+    position: 1,
+    required: true,
+    choices: ['install', 'remove']
+  })
+  .help('Installs the Tessel SDK for binary cross-compilation.');
```
rwaldron: This portion can can be removed, as there is now `t2 install x` (with a `rust-sdk` waiting to be implemented

lib/install/rust.js
```diff
@@ -0,0 +1,424 @@
...
+  var stream = new Transform();
+  stream._transform = function(chunk, encoding, callback) {
+    this.push(chunk);
+    sha256.update(chunk);
+    callback();
+  };
+  stream.on('finish', function() {
+    stream.emit('sha256', sha256.digest('hex'));
+  });
+  return stream;
```
rwaldron: Can use arrow functions inside `sha256stream`

lib/install/rust.js
```diff
@@ -0,0 +1,424 @@
...
+function tmpdir() {
+  return new Promise((resolve) => {
+    var dir = fsTemp.template('t2-sdk-%s').mkdirSync();
+    resolve({
+      path: dir,
+      cleanup: () => {
+        try {
+          fs.removeSync(dir);
+        } catch (e) {
+          // Swallow errors in removing temporary folder.
```
rwaldron: In what cases do you encounter errors?

lib/install/rust.js
```diff
@@ -0,0 +1,424 @@
...
+  }
+}
+
+var SDK_PATHS = {
+  sdk: path.join(osenv.home(), '.tessel/sdk'),
+  rustlib: path.join(osenv.home(), '.tessel/rust'),
+};
+
+var SDK_URLS = {
+  macos: 'https://builds.tessel.io/t2/sdk/t2-sdk-macos-x86_64.tar.bz2',
```
rwaldron: Lines 23-26 are unnecessary if you just rename "macos" => "darwin", otherwise it's unnecessary complexity

lib/install/rust.js
```diff
@@ -0,0 +1,424 @@
...
+    checked: false,
+    path: dir,
+  }));
+};
+
+module.exports.installSdk = () => {
+  var url = SDK_URLS[getPlatform()];
+  var checksumVerify = null;
+
+  return downloadString(url + '.sha256')
```
rwaldron: Use template string

tcr:
Alright @rwaldron, I think this is good for another review:

* The unit tests are failing; I'll be fixing these and adding new tests tomorrow.
* `t2 sdk` changed to `t2 install rust-sdk` (see my comments in #engineering on Slack) 
* Added support for optional `--rustcc` flag, with `--rustcc=localhost:xxxx` override. If the SDK isn't installed, user is prompted to either install it or use the remote compilation flag.
* Changed the command to `t2 rust <bin-name>`, where bin-name is the name you would use for `cargo build --bin <bin-name>`. For the blinky example, this is `t2 rust blinky`. I want to evaluate how well this works in practice. It's nagging that it isn't the path to a Rust file, but I am not sure how to improve it.
* Added a `cargo-tessel` script which we can use for the cross-compilation server to generate tarballs. You'll note that if you try the cross-compilation server that it probably fails; the real fix is to integrate `cargo-tessel` into the server once this PR is accepted. The unit tests should still confirm the server code is functional.

tcr:
Actually ready for review:

* The unit tests are failing; will work on these after next review.
* Tessel CLI installs two commands, `t2` and `cargo-tessel`.
* If you don't have Rust installed, `t2 run` tells you to install Rust or use --rustcc.
* If you do have Rust installed, `t2 run` suggests you run `cargo tessel sdk install` or use --rustcc.
* `cargo tessel build --bin {name}` is available for the cross-compilation server.
* `t2 run` has the optional `--rustcc` flag, with `--rustcc=localhost:xxxx` override. If the SDK isn't installed, user is prompted to either install it or use the remote compilation flag. (The cross-compilation will fail but that is unrelated to any code in our CLI.)

Same caveat as above about trying out `t2 rust <bin-name>`, where bin-name is the name you would use for `cargo build --bin <bin-name>`. For the blinky example, this is `t2 rust blinky`. If you type `t2 run Cargo.toml` it suggests a list of binaries. Again, I don't know if this is the right solution, but it feels Less Wrong than the other ones. Let's see how this goes.

tcr:
(One minor thing: maybe `--rustcc` should just be a flag argument, and `RUST_COMPILATION_SERVER` an env flag. It's weird for `--rustcc` to only optionally take an argument, nomnom doesn't really support optional values, and most users won't change the server path anyway.)

bin/cargo-tessel.js
```diff
@@ -0,0 +1,61 @@
...
+var parser = require('nomnom');
+
+// Internal
+var log = require('../lib/log');
+var rust = require('../lib/install/rust');
+
+function die(err) {
+  log.error(err.stack);
+  process.exit(1);
+}
```
rwaldron: We should refactor this to use `closeSuccessfulCommand` and `closeFailedCommand`, but I won't block this PR on that.

bin/cargo-tessel.js
```diff
@@ -0,0 +1,61 @@
...
+
+parser.command('build')
+  .callback(function(opts) {
```
rwaldron: Use arrow function (for consistency throughout this file)

lib/install/rust.js
```diff
@@ -0,0 +1,523 @@
+// System Objects
+var path = require('path');
+var spawn = require('child_process').spawn;
+var spawnSync = require('child_process').spawnSync;
```
rwaldron:
```js
var cp = require('child_process');
```

Then `cp.spawn` and `cp.spawnSync` throughout the file to be consistent throughout the project.

lib/install/rust.js
```diff
@@ -0,0 +1,523 @@
...
+var Transform = require('stream').Transform;
```
rwaldron: Same as above. Here's an example from usb-connection.js:

```js
// System Objects
var util = require('util');
var stream = require('stream');
var events = require('events');

var Duplex = stream.Duplex;
var Emitter = events.EventEmitter;
```


Please follow conventions—it reduces cognitive load for incoming contributors and future readers/maintainers of the code base.

lib/install/rust.js
rwaldron: @tcr ping.

lib/install/rust.js
```diff
@@ -0,0 +1,523 @@
...
+
+  return {
+    name: out.name,
+    path: dest
+  };
+};
+
+module.exports.runBuild = (matchBin) => {
+  return new Promise((resolve) => {
+      exports.checkRust(true);
```
rwaldron: `checkRust` really needs to be a promise returning function that resolves in safe-to-proceed conditions and otherwise rejects. Since this is in the `t2 run ...` path, it must not subvert the closing process, which is must be followed to ensure the connection to board is completely closed, whether there is a complete success or complete failure.

lib/install/rust.js
```diff
@@ -0,0 +1,523 @@
...
+    } else {
+      log.error(`No binary target "${matchBin}" exists for this Rust crate.\nMake sure you specify a valid binary using --bin.`);
+    }
+    if (bins.length > 0) {
+      log.error('Available targets:');
+      bins.forEach(bin => {
+        log.error(' $ cargo tessel build --bin', bin.name);
+      });
+    }
+    process.exit(1);
```
rwaldron: Since this is in the `t2 run ...` path, `checkConfiguration` must return a promise as explained in https://github.com/tessel/t2-cli/pull/1014/files#r82441981

tcr:
@rwaldron checkRust and check configuration can't be async if they're called in the lang meta.checkConfiguration function. Should I make that function return a promise?

lib/tessel/deployment/rust.js
```diff
@@ -1,21 +1,20 @@
...
-var path = require('path');
 var url = require('url');
 
 // Third Party Dependencies
 var fs = require('fs-extra');
 var tags = require('common-tags');
 var tar = require('tar');
-var toml = require('toml');
 var Reader = require('fstream').Reader;
+var tags = require('common-tags');
```
rwaldron: Now there are two `var tags = require('common-tags');`, so let's just clean up the whole section, change to: 

```js
// Third Party Dependencies
var fs = require('fs-extra');
var fstream = require('fstream');
var tags = require('common-tags');
var tar = require('tar');

var Reader = fstream.Reader;
```

rwaldron:
Calling @HipsterBrown! I need you as a co-reviewer

rwaldron:
@tcr 

> checkRust and check configuration can't be async if they're called in the lang meta.checkConfiguration function. Should I make that function return a promise?

`lang.meta.checkConfiguration` is only supposed to be "checking the configuration file defined by the property `configuration`". It's only used in an edge case where no target project root directory could be found by `findProject`—that's the edge case you've been exploiting when you use `t2 run Cargo.toml`.

rwaldron:
Re: 

> "checking the configuration file defined by the property configuration"

https://github.com/tessel/t2-cli/pull/1014#issuecomment-252329476

tcr:
Review feedback addressed! This should also pass nodeunit tests again. @rwaldron @HipsterBrown

HipsterBrown:
@tcr I'll check it out tonight.

lib/install/rust.js
```diff
@@ -477,13 +477,13 @@ module.exports.cli = {
...
   return exports.getBuildConfig()
-    .catch(e => {
+    .catch(e => new Promise((resolve, reject) => { // jshint ignore:line
       log.error('Could not find all the components for cross-compiling Rust:');
       log.error(e.message);
       log.info('Please run "cargo tessel sdk install" and try again.');
       log.info('To instead use the remote Rust compiler, use `t2 run <target> --rustcc`.');
-      process.exit(1);
-    });
+      reject();
```
rwaldron: `reject(...)` is never called with no arguments. 

To reject with a verbose message, compose the contents and pass it along to `reject(...)`, something like this:  

```js
module.exports.checkSdk = () => {
  return new Promise((resolve, reject)) {
    exports.getBuildConfig().then(resolve).catch(error => {
      var message = tags.stripIndent`
        Could not find all the components for cross-compiling Rust:
        ${error.message}
        Please run "cargo tessel sdk install" and try again.
        To instead use the remote Rust compiler, use "t2 run <target> --rustcc".`;

      reject(message);
    });
  });
};
```

lib/install/rust.js
rwaldron: ?

rwaldron:
To figure out the error on travis, I did `npm prune` and it looks like you didn't remove the `toml` module from the test bootstrapping, which is resulting in this error: 

```
$ grunt nodeunit:files:deploy
Running "nodeunit:files:deploy" (nodeunit:files) task

Running "nodeunit:file:deploy" (nodeunit:file) task

Fatal error: Cannot find module 'toml'
```

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+module.exports.checkBinaryName = (matchBin, destdir) => {
+  return exports.cargoMetadata(destdir)
+    .then(metadata => {
+      return new Promise((resolve, reject) => {
+        // Get first package.
+        var pkg = metadata.packages.pop();
+        var bins = pkg.targets.filter(target => target.kind.indexOf('bin') > -1);
+
+        // Filter by --bin argument.
+        var validbins = bins.filter(bin => bin.name === matchBin);
```
HipsterBrown: Should be camelCase -> `validBins`.

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+            message = 'Please specify a valid binary target with --bin.';
+          } else {
+            message = tags.stripIndent `
+              No binary target "${matchBin}" exists for this Rust crate.
+              Make sure you specify a valid binary using --bin.`;
+          }
+          if (bins.length > 0) {
+            message += '\nAvailable targets:';
+            bins.forEach(bin => {
+              message += `\n $ cargo tessel build --bin ${bin.name}`;
```
HipsterBrown: Folks might also copy and paste the `$` in this line.

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+            });
+          }
+          reject(message);
```
HipsterBrown: I think you need to `return reject(message)` here so that execution ends.

lib/tessel/deployment/rust.js
```diff
@@ -126,9 +127,15 @@ exportables.remoteRustCompilation = (opts) => {
...
         // When the post completes, resolve with the executable
         .on('end', function() {
-
           // Parse the incoming data as JSON
-          var result = JSON.parse(Buffer.concat(buffers));
+          var result;
+          try {
+            result = JSON.parse(Buffer.concat(buffers));
+          } catch (e) {
+            log.error('Invalid JSON response received from cross-compilation server:');
```
HipsterBrown: What does this mean to the person using this tool? How do they fix it?

lib/tessel/deployment/rust.js
```diff
@@ -126,9 +127,15 @@ exportables.remoteRustCompilation = (opts) => {
...
+            log.error('    ', Buffer.concat(buffers).toString());
+            process.exit(1);
```
HipsterBrown: Doesn't this happen when the command Promise is rejected?

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+  return new Promise((resolve, reject) => {
+    var sdkPlatformPath = path.join(SDK_PATHS.sdk, getPlatform());
+    var values = fs.readdirSync(sdkPlatformPath);
+
+    for (var i = 0; i < values.length; i++) {
+      if (values[i].match(/^toolchain\-/)) {
+        return resolve(path.join(sdkPlatformPath, values[i]));
+      }
+    }
+    return reject(new Error('No toolchain found.'));
```
tcr: Any function which invokes of this promise, if rejected, will instruct the user to run `cargo tessel sdk install`. This function just diagnoses the specific problem.

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+    var rustcOut = [];
+    rustc.stdout.on('data', (data) => {
+      rustcOut.push(data);
+    });
+    rustc.on('error', reject);
+    rustc.on('close', (status) => {
+      var out = Buffer.concat(rustcOut).toString();
+      var matches = out.match(/^rustc\s+(\S+)/);
+      if (status !== 0 || !matches) {
+        return reject(new Error('Could not find a locally installed version of `rustc`.'));
```
tcr: The error (go to rustup.sh) is displayed by consumers of rustVersion. This just identifies a specific issue.

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+      if (!check.exists) {
+        throw new Error('SDK not installed.');
+      }
+      config.stagingDir = check.path;
+
+      return exports.checkRustlib(config.rustv);
+    })
+    .then(check => {
+      if (!check.exists) {
+        throw new Error(`MIPS libstd v${config.rustv} not installed.`);
```
tcr: Instructions to install it are displayed by consumers of getBuildConfig().

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+
+  return exports.rustVersion()
+    .then(rustv => {
+      config.rustv = rustv;
+
+      return exports.checkTools();
+    })
+    .then(check => {
+      if (!check.exists) {
+        throw new Error('SDK not installed.');
```
tcr: Instructions to install it are displayed by consumers of getBuildConfig().

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+module.exports.installRustlib = () => {
+  return exports.rustVersion()
+    .then((rustv) => {
+      var pkgname = `MIPS libstd v${rustv}`;
+      var url = RUST_LIB_TGZ_URL.replace('VERSION', rustv);
+      var checksumVerify;
+
+      return downloadString(url + '.sha256')
+        .catch(() => {
+          throw new Error(`Could not find a MIPS libstd matching your current Rust version (${rustv}).`);
```
tcr: It's obtuse, but the only consumer of `installRustlib` (module.exports.cargo.install) explains that the next step is to change to a stable Rust and re-run the sdk install step.

tcr:
Whew, thanks @HipsterBrown.

lib/install/rust.js
tcr: Even considering the comment on line 36?

bin/cargo-tessel.js
```diff
@@ -12,26 +12,30 @@ var parser = require('nomnom');
...
 
 parser.command('build')
   .callback(options => {
-    rust.runBuild(options.bin)
+    rust.runBuild(false, options.bin)
       .then(tarball => {
+        // This is the direct invocation of `cargo tessel build ...` (not
+        // through t2 run). As a command line tool, it only writes to stdout
+        // the path of the cross-compiled tarball it generates. This allows This
+        // to be integrated into CLI toolchains by mapping its stdout.
```
HipsterBrown: Thank you for clarifying :+1:

lib/install/rust.js
```diff
@@ -265,7 +264,11 @@ function extract(checksumVerify, filename, sdkStream, root, strip, name, decompr
           .on('finish', () => {
             // Check sum.
             if (checksum !== checksumVerify) {
-              return reject(new Error(`Checksum for downloaded ${name} does not match!`));
+              return reject(tags.stripIndent `
+                Please file an issue on https://github.com/tessel/t2-cli with the following:
+                The downloaded file ${name} is invalid (wrong checksum).
+                  expected: ${checksumVerify}
+                  got:      ${checksum}`);
```
HipsterBrown: This is great! :clap:

lib/tessel/deployment/rust.js
```diff
@@ -132,9 +132,10 @@ exportables.remoteRustCompilation = (opts) => {
           try {
             result = JSON.parse(Buffer.concat(buffers));
           } catch (e) {
-            log.error('Invalid JSON response received from cross-compilation server:');
-            log.error('    ', Buffer.concat(buffers).toString());
-            process.exit(1);
+            return reject(tags.stripIndent `
+              Please file an issue on https://github.com/tessel/t2-cli with the following:
+              You received an invalid JSON response from the cross-compilation server.
+                  ${Buffer.concat(buffers).toString()}`);
```
HipsterBrown: This is great feedback for the user but I'm curious if the crash-reporter will also submit an issue if this happens.

tcr: This isn't a crash, but we could make it crash I suppose ;)

HipsterBrown: Nah, that's fine. Just wanted to confirm that. :+1:

bin/cargo-tessel.js
```diff
@@ -0,0 +1,56 @@
...
+  .help('Manage the SDK for cross-compiling Rust binaries.');
+
+if (require.main === module) {
+  if (process.argv[2] === 'tessel') {
+    // 'cargo tessel' invocation
+    parser.parse(process.argv.slice(3));
+  } else {
+    // 'cargo-tessel' invocation
+    parser.parse(process.argv.slice(2));
+  }
```
rwaldron: @HipsterBrown I also was confused, but it looks like third part developers are encouraged to make `cargo` extensions: https://github.com/rust-lang/cargo/wiki/Third-party-cargo-subcommands

lib/install/rust.js
rwaldron: Note: don't use back ticks in strings, ie. 'Could not find a locally installed version of `rustc`.'. There is no formatting done to these strings to make those ticks useful, and they can be potentially confusing since we're using template strings throughout the code. Since we're already using single quotes for all other strings, I've been trying to convert any that I come across to double quotes for literal things. This approach is adapted directly from `git`, eg. from `git help`: 


```

'git help -a' and 'git help -g' list available subcommands and some
concept guides. See 'git help <command>' or 'git help <concept>'
to read about a specific subcommand or concept.
```

lib/install/rust.js
```diff
@@ -0,0 +1,559 @@
...
+      stdio: ['ignore', 'inherit', 'inherit'],
+    });
+
+    cargo.on('error', (error) => {
+      log.error(`${error.stack}`);
+    });
+
+    cargo.on('close', (code) => {
+      if (code !== 0) {
+        process.exit(code);
```
rwaldron: @HipsterBrown this and any other non-standard exit will have to be refactored to use the same "safely close all tessels" logic used elsewhere.

lib/install/rust.js
rwaldron: I guess I missed that. Seems unfortunate to have to have all this additional plumbing :\

lib/install/rust.js
```diff
@@ -0,0 +1,557 @@
...
+
+// Checks if CHECKSUM file in our SDK equals our expected checksum.
+// This will resolve with checking that the SDK exists and matches the checksum.
+module.exports.checkTools = (checksumVerify) => {
+  var dir = path.join(SDK_PATHS.sdk, getPlatform());
+  return new Promise((resolve) => {
+    var checksum = fs.readFileSync(path.join(dir, 'CHECKSUM'), 'utf-8');
+    resolve({
+      exists: true,
+      checked: checksumVerify === checksum,
```
rwaldron: If the operation is "verification" and whether or not the tools are verified, then the property should probably be `isVerified`, vs. `checked` (checked doesn't indicate success or failure, ie. I could check something, but the result could've been failure)

lib/install/rust.js
```diff
@@ -0,0 +1,557 @@
...
+  }));
+};
+
+module.exports.checkRustlib = (rustv, checksumVerify) => {
+  var dir = path.join(SDK_PATHS.rustlib, rustv);
+  return new Promise((resolve) => {
+    var checksum = fs.readFileSync(path.join(dir, 'CHECKSUM'), 'utf-8');
+    resolve({
+      exists: true,
+      checked: checksumVerify === checksum,
```
rwaldron: Same notes as above

lib/install/rust.js
```diff
@@ -0,0 +1,557 @@
...
+      path: dir,
+    });
+  }).catch(() => ({
+    exists: false,
+    checked: false,
```
rwaldron: And again.

rwaldron: Actually, this is a nice example: this thing _was_ "checked", it just failed for whatever reason.

lib/install/rust.js
```diff
@@ -0,0 +1,557 @@
...
+module.exports.installRustlib = () => {
+  return exports.rustVersion()
+    .then((rustv) => {
+      var pkgname = `MIPS libstd v${rustv}`;
+      var url = RUST_LIB_TGZ_URL.replace('VERSION', rustv);
+      var checksumVerify;
+
+      return downloadString(url + '.sha256')
+        .catch(() => Promise.reject(tags.stripIndent `
+          Could not find a MIPS libstd matching your current Rust version (${rustv}).`))
```
rwaldron:
```js
 .catch(() => Promise.reject(`Could not find a MIPS libstd matching your current Rust version (${rustv}).`)
```

lib/install/rust.js
tcr: Agreed.

lib/install/rust.js
rwaldron: 👍

lib/install/rust.js
tcr: @rwaldron This is changed in a later commit.

lib/install/rust.js
```diff
@@ -0,0 +1,557 @@
...
+    var dir = fsTemp.template('t2-sdk-%s').mkdirSync();
+    resolve({
+      path: dir,
+      cleanup: () => {
+        try {
+          fs.removeSync(dir);
+        } catch (e) {
+          // Swallow errors in removing temporary folder. If the folder was
+          // successfully or unsuccessfully moved, it may not exist at its
+          // location in the temporary directory, but this isn't fatal.
```
rwaldron: When does this happen?

tcr: @rwaldron We unpack the tarball into a temporary folder, then move it to ~/.tessel/sdk/{os}. If anything goes wrong in either step, we still want to remove the very large temporary folder, which may or may not be there depending on how far we got.

tcr: (Note that the move *should* be atomic, but fs.move allows us gracefully to move across disjoint filesystems, and I assume that is a non-atomic, possibly copy operation.)

rwaldron:
Things remaining: 

- end-to-end smoke testing instructions
  - complete steps for primary use cases
  - @HipsterBrown can you think of specific smoke tests?
- unit tests

HipsterBrown:
@rwaldron 

> can you think of specific smoke tests?

I think @tcr's "Testing" description should be good to cover the usage of this build.

tcr:
@rwaldron @HipsterBrown I renamed the "testing" section to "Usage". The "Testing" section has a series of step by step instructions to test all permutations of the SDK usage locally, though it's more excessive than what this needs to be merged.

test/unit/deployment/rust.js
```diff
@@ -213,22 +213,26 @@ exports['deploy.rust'] = {
   // name of the binary instead any of the code files of the project
   rustPreBundle: function(test) {
 
-    test.expect(1);
-
-    // Gather the project info for the project we'll target
-    var cargoToml = toml.parse(fs.readFileSync(path.join(DEPLOY_DIR_RS, 'Cargo.toml'), 'utf8'));
+    test.expect(0);
```
rwaldron: ?

rwaldron:
Is this really still a thing? `t2 sdk install`?

rwaldron:
These are probably in the wrong order:

```
$ cargo-tessel sdk install
INFO Installing Tessel build tools...
INFO Downloading Tessel build tools...
     [========            ] 40% 19.6s remaining
```

lib/install/rust.js
rwaldron: All rustlib components are named as I assumed they'd be, using "darwin": 

```
rustup: installing toolchain for 'nightly'
rustup: extracting installer
install: uninstalling component 'rustc'
install: uninstalling component 'rust-std-x86_64-apple-darwin'
install: uninstalling component 'rust-docs'
install: uninstalling component 'cargo'
install: creating uninstall script at /usr/local/lib/rustlib/uninstall.sh
install: installing component 'rustc'
install: installing component 'rust-std-x86_64-apple-darwin'
install: installing component 'rust-docs'
install: installing component 'cargo'
```

tcr:
`t2 sdk install` is not and is removed. `cargo tessel sdk install` is a thing.

tcr:
Switched to 

```
$ cargo-tessel sdk install
INFO Installing Tessel build tools...
INFO Installing MIPS libstd v1.12.0...
INFO SDK installed.
```

It's cleaner.

rwaldron:
Some smoke testing...

```
rwaldron in ~/robotz/tessel-rust
$ cargo-tessel sdk install
INFO Installing Tessel build tools...
INFO Downloading Tessel build tools...
INFO Installing MIPS libstd v1.12.0...
INFO Downloading MIPS libstd...
INFO SDK installed.
rwaldron in ~/robotz/tessel-rust
$ cargo-tessel sdk install
INFO Latest Tessel build tools already installed.
INFO Latest MIPS libstd v1.12.0 already installed.
INFO SDK installed.
rwaldron in ~/robotz/tessel-rust
$ cargo-tessel sdk uninstall
INFO Tessel SDK uninstalled.
rwaldron in ~/robotz/tessel-rust
$ cargo-tessel sdk install
INFO Installing Tessel build tools...
INFO Downloading Tessel build tools...
INFO Installing MIPS libstd v1.12.0...
INFO Downloading MIPS libstd...
INFO SDK installed.
rwaldron in ~/robotz/tessel-rust
$ t2 init --lang=rust
INFO Initialized Cargo project...
INFO Wrote "Hello World" to /Users/rwaldron/robotz/tessel-rust/src/main.rs
rwaldron in ~/robotz/tessel-rust
$ ls
.
├── [drwxr-xr-x rwaldron staff     102 Oct 12 14:00]  src/
└── [-rw-r--r-- rwaldron staff     119 Oct 12 14:00]  Cargo.toml

1 directory, 1 file
rwaldron in ~/robotz/tessel-rust
$ cat Cargo.toml
[package]
name = "blinky"
version = "0.0.1"
authors = ["Your Name <you@example.com>"]

[dependencies]
tessel = "0.2.0"
rwaldron in ~/robotz/tessel-rust
$ cat src/main.rs
//! A blinky example for Tessel

// Import the tessel library
extern crate tessel;
// Import the Tessel API
use tessel::Tessel;
// Import sleep from the standard lib
use std::thread::sleep;
// Import durations from the standard lib
use std::time::Duration;

fn main() {
    // Create a new Tessel
    let mut tessel = Tessel::new();

    // Turn on one of the LEDs
    tessel.led[2].on().unwrap();

    println!("I'm blinking! (Press CTRL + C to stop)");

    // Loop forever
    loop {
        // Toggle each LED
        tessel.led[2].toggle().unwrap();
        tessel.led[3].toggle().unwrap();
        // Re-execute the loop after sleeping for 100ms
        sleep(Duration::from_millis(100));
    }
}
rwaldron in ~/robotz/tessel-rust
$ t2 list
  LAN sonny
  USB sonny
rwaldron in ~/robotz/tessel-rust
$ t2 run blinky
INFO Connected to sonny.
    Updating registry `https://github.com/rust-lang/crates.io-index`
 Downloading tessel v0.2.0
 Downloading unix_socket v0.5.0
 Downloading libc v0.2.16
 Downloading cfg-if v0.1.0
   Compiling libc v0.2.16
   Compiling cfg-if v0.1.0
   Compiling unix_socket v0.5.0
   Compiling tessel v0.2.0
   Compiling blinky v0.0.1 (file:///Users/rwaldron/robotz/tessel-rust)
    Finished release [optimized] target(s) in 2.81 secs
INFO Writing project to RAM on sonny (338.432 kB)...
INFO Deployed.
INFO Running blinky...
^C
```

...snip...

> Run t2 run blinky --rustcc when rustc is not available
> Run t2 run blinky when rustc is available.

Both cases have same result: Finds tessel, then explodes. This produces more `__stack_chk_guard` errors than my terminal scrollback can display, but ultimately ends with: 

```
ERR!
ERR!
ERR! error: aborting due to previous error
ERR!
ERR! error: Could not compile `blinky`.
ERR!
ERR! To learn more, run the command again with --verbose.
```

tcr:
@rwaldron Everything looks good there, including the `__stack_chk_guard` mess. The `rustcc` server needs it config file updated, though it should be timed with the next CLI release (more or less)

lib/tessel/deployment/rust.js
```diff
@@ -53,7 +53,7 @@ exportables.preBundle = function(opts) {
     return Promise.resolve();
   } else {
     // Check that rust, the sdk, and the binary name are valid.
-    return rust.checkRust()
+    return rust.checkRust(true)
```
rwaldron: Oof—we eliminated all of the "boolean trap" parameters! Because `rust.checkRust(true)` means nothing useful :\ Does `rust.checkRust(false)` _not_ "check rust"?

lib/install/rust.js
```diff
@@ -488,41 +490,42 @@ module.exports.checkSdk = () => {
...
 };
 
-module.exports.checkRust = (isCli) => {
+module.exports.checkRust = (opts) => {
   return exports.rustVersion()
     .catch(() => Promise.reject(tags.stripIndent `
         "rustc" and "cargo" are required to cross-compile for Tessel.
         Please install Rust on your machine: https://rustup.rs/` +
-      (isCli ? '\nTo instead use the remote Rust compiler, use "t2 run <target> --rustcc".' : '')
+      (opts.isCli ? '\nTo instead use the remote Rust compiler, use "t2 run <target> --rustcc".' : '')
```
rwaldron: No need to break the string like this, you can put expressions in the `${}`: 

```js
  return exports.rustVersion()
    .catch(() => Promise.reject(tags.stripIndent `
        "rustc" and "cargo" are required to cross-compile for Tessel.
        Please install Rust on your machine: https://rustup.rs/
        ${opts.isCli ? To instead use the remote Rust compiler, use "t2 run <target> --rustcc". : ''}`)
    ));        
```

lib/install/rust.js
```diff
@@ -543,9 +546,12 @@ module.exports.checkBinaryName = (isCli, matchBin, destdir) => {
...
 
-module.exports.runBuild = (isCli, matchBin, destdir) => {
-  return exports.checkRust(isCli)
-    .then(() => exports.checkBinaryName(isCli, matchBin, destdir))
+// opts matches { isCli: boolean, binary: String[, path: String] }
+module.exports.runBuild = (opts) => {
+  return exports.checkRust({
+      isCli: opts.isCli
+    })
+    .then(() => exports.checkBinaryName(opts))
```
rwaldron: 👍

rwaldron:
@tcr I don't understand what the change in the last commit does w/r to its commit message:

> Fixes tests when Rust isn't installed.


Rust shouldn't need to be installed for the tests to run

tcr:
@rwaldron Right, the commit adds a stub for checkRust:

```
this.checkRust = sandbox.stub(rust, 'checkRust', () => Promise.resolve());
```

rwaldron:
@HipsterBrown please review my commits. Ping me when done—thanks!!

HipsterBrown:
I pushed an update to fix the `jshint`errors. I ran @rwaldron's tests locally and passed. They look good with about 75% of coverage from the report. 

I think `lib/install/rust.js` will need more tests in a followup PR but this build looks good to me right now. :+1: 

All that is left should be the conflicting `package.json`.

rwaldron:
Landed!

PR #1030, opened 2016-10-11
Updates package.json, changes supported Node versions to >=4.2.0

tcr:
Additionally, this allows users to install `t2-cli` with `yarn global add t2-cli`. `npm install -g` doesn't need the "engines" value to match your current Node version, but `yarn` enforces it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "technical",
     "machine"
   ],
-  "author": "Technical Machine",
+  "author": "Tessel Project",
   "license": "MIT",
   "bugs": {
     "url": "https://github.com/tessel/t2-cli/issues"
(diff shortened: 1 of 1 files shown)


tcr:
The limiting factor before was that Node 6.x wasn't precompiling builds for `usb`; we will be able to easily support each new node version for `usb` going forward so I don't (personally) see the need for an upper bound like `>=4.2.0 <7.0`.

rwaldron:
The mac builds on travis are failing, but not because of anything this project can control (that I know of). See: https://github.com/tessel/t2-cli/pull/1022

johnnyman727:
:clap:

PR #1043, opened 2016-10-15
Hotfixes to Rust compilation.

tcr:
This PR makes the CLI work with the cross compilation server (now live). It does the following:

* Fixes rust SDK installation on Linux, which was incorrect.
* Fixes a bug where the Linux toolchain couldn't be found due to a generous regex.
* Passes the `target` name to the Rust cross-compilation server.
* Removes the port from `rustcc.tessel.io` and defaults to port 80.

This is ready to merge whenever and a new release should be cut after that, enabling the `--rustcc` switch to cross-compile Rust binaries.

### DIFF
--- a/lib/install/rust.js
+++ b/lib/install/rust.js
@@ -139,7 +139,7 @@ module.exports.toolchainPath = () => {
     var values = fs.readdirSync(sdkPlatformPath);
 
     for (var i = 0; i < values.length; i++) {
-      if (values[i].match(/^toolchain\-mipsel/)) {
+      if (values[i].startsWith('toolchain-mipsel')) {
         return resolve(path.join(sdkPlatformPath, values[i]));
       }
     }
--- a/lib/install/rust.js
+++ b/lib/install/rust.js
@@ -139,7 +139,7 @@ module.exports.toolchainPath = () => {
     var values = fs.readdirSync(sdkPlatformPath);
 
     for (var i = 0; i < values.length; i++) {
-      if (values[i].match(/^toolchain\-/)) {
+      if (values[i].match(/^toolchain\-mipsel/)) {
         return resolve(path.join(sdkPlatformPath, values[i]));
       }
     }
@@ -304,7 +304,7 @@ function extract(checksumVerify, filename, sdkStream, root, strip, name, decompr
 }
 
 function extractTools(checksumVerify, filename, sdkStream) {
-  var root = path.join(SDK_PATHS.sdk, 'macos');
+  var root = path.join(SDK_PATHS.sdk, getPlatform());
   return extract(checksumVerify, filename, sdkStream, root, 2, 'Tessel build tools', bz2());
 }
 
--- a/lib/tessel/deployment/rust.js
+++ b/lib/tessel/deployment/rust.js
@@ -1,5 +1,6 @@
 // System Objects
 var http = require('http');
+var querystring = require('querystring');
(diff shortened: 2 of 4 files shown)


HipsterBrown:
@tcr Is there a good smoke test for this? I can do a release tomorrow once I can confirm it works locally.

tcr:
@HipsterBrown Ah, running the rust blinky example with rustcc will do it:

```
mkdir blinky
cd blinky
t2 init --lang=rust
t2 run blinky --rustcc
```

It will compile using the cross-compilation server and succeed.

HipsterBrown:
Ok thanks @tcr. Maybe @rwaldron can do a smoke test as well. I'll check that out during lunch anyway. 

I did some investigating into those false CI failures since all the tests pass locally and still have not been able to discover a fix. But I would like to see if we can resolve those before we do much more feature development.

HipsterBrown:
Now that the build is green, I ran the smoke test with the following results:

```
nickhehr@Nicks-MacBook-Pro-2 blinky $ t2 init --lang=rust
INFO Initialized Cargo project...
INFO Wrote "Hello World" to /Users/nickhehr/Projects/blinky/src/main.rs
nickhehr@Nicks-MacBook-Pro-2 blinky $ t2 run blinky --rustcc
INFO Compiling Rust code remotely (--rustcc)...
INFO Updating registry `https://github.com/rust-lang/crates.io-index`
INFO  Downloading libc v0.2.17
INFO    Compiling cfg-if v0.1.0
INFO    Compiling libc v0.2.17
INFO    Compiling unix_socket v0.5.0
INFO    Compiling tessel v0.2.0
INFO    Compiling blinky v0.0.1 (file:///tmp/tessel-rust-compile116919-1615-zfeajx.k0nr0be29/blinky)
INFO     Finished release [optimized] target(s) in 6.33 secs
INFO INFO Tessel bundle written out to:
INFO /tmp/tessel-rust-compile116919-1615-zfeajx.k0nr0be29/blinky/target/tessel2/release/tessel-bundle.tar
INFO Writing project to RAM on tessel-router (337.92 kB)...
INFO Deployed.
INFO Running blinky...
  nickhehr@Nicks-MacBook-Pro-2 blinky $
```

Looks like a success to me. :+1:

lib/install/rust.js
```diff
@@ -139,7 +139,7 @@ module.exports.toolchainPath = () => {
     var values = fs.readdirSync(sdkPlatformPath);
 
     for (var i = 0; i < values.length; i++) {
-      if (values[i].match(/^toolchain\-/)) {
+      if (values[i].match(/^toolchain\-mipsel/)) {
```
rwaldron: Clarifying q: this is just checking that the string in `values[i]` includes `"toolchain-mipsel/"`?

lib/tessel/deployment/rust.js
```diff
@@ -113,7 +114,7 @@ exportables.remoteRustCompilation = (opts) => {
     var options = {
       host: destination.hostname,
       port: destination.port,
-      path: '/rust-compile',
+      path: '/rust-compile?target=' + querystring.escape(opts.resolvedEntryPoint),
```
rwaldron: Use template string: 

```js
`/rust-compile?target=${querystring.escape(opts.resolvedEntryPoint)}`
```

rwaldron:
@tcr please address the two notes before merging

Issue #850, opened 2016-09-02
2 Tessels plugged in, CLI can find both but one of them hangs


Frijol:
My left USB port Tessel just hangs trying to execute. Sample output:

```sh
➜  t2-cli git:(adhoc) t2 wifi --off
INFO Looking for your Tessel...
INFO Connected to Mariano.
INFO Connected to trendelmame.
/ (spinning)
```
indefinitely.

Works ok if I disconnect one of the two. I believe the one that hangs is always the one that connected second.

@HipsterBrown says he ran into this as well

johnnyman727:
Probably a duplicate of #721.

PR #1045, opened 2016-10-17
Multiple usb fix

HipsterBrown:
Fixes https://github.com/tessel/t2-cli/issues/850

Smoke Test:

- Connect 2 Tessels to computer via USB
- Run a command for Tessel, like `t2 wifi`
- Should see a prompt to pick one of the 2 Tessels connected to your computer

```
hipsterbrown:t2-cli (multiple-usb-fix) $ t2 wifi
INFO Which Tessel do want to use?
  	USB	Tessel-02A3B3DCF2D8
  	USB	tessel-router
| 	Exit \
```

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -223,7 +223,8 @@ Tessel.get = function(opts) {
                       choices: tessels.map(function(tessel, i) {
                         var isLAN = !!tessel.lanConnection;
                         var isAuthorized = isLAN && tessel.lanConnection.authorized;
-                        var authorization = isAuthorized ? '' : '(not authorized)';
+                        // show "(not authorized)" if there is a lanConnection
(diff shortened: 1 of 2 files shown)


lib/controller.js
```diff
@@ -274,11 +275,6 @@ Tessel.get = function(opts) {
       if (opts.name && opts.name === tessel.name) {
         finishSearchEarly(tessel);
       }
-      // If we just found a USB connection and should prefer it
-      else if (!opts.name && tessel.usbConnection !== undefined && !opts.lanPrefer) {
-        // Finish early with this Tessel
-        finishSearchEarly(tessel);
-      }
```
rwaldron: Shouldn't this condition should stay intact, only skipping `finishSearchEarly(tessel);` if there are more than one attached?

rwaldron:
Smoke test passed!

PR #1047, opened 2016-10-17
tests: close all TesselSimulator objects in teardowns

rwaldron:


rwaldron:
About to re-build to make sure we're not seeing false positives

HipsterBrown:
LGTM :+1:

Thanks for digging into this issue @rwaldron. It will be nice to see a green build badge again. Merge when ready.

PR #1059, opened 2016-10-24
Fixes broken call to runBuild for Rust code.

tcr:
I missed an invocation of this method when changing the function signature.

I long for strong typing. :whine:


HipsterBrown:
Should there be a test we can add that will catch that error in the future?

rwaldron:
@HipsterBrown I just pushed some tests for this. Everything looks good, can land when green

HipsterBrown:
CI is green, now merging.

Issue #1057, opened 2016-10-22
Crash report: TypeError: Cannot read property 'stop' of undefined



tessel-crashbot:
```
TypeError: Cannot read property 'stop' of undefined
    at LAN.Scanner.stop (t2-cli/lib/lan-connection.js:206:15)
    at Object.stopScan (t2-cli/lib/lan-connection.js:159:13)
    at TesselSeeker.stop (t2-cli/lib/discover.js:125:24)
    at Promise.all.then (t2-cli/lib/discover.js:114:16)
```

Full report is at [0x223008ed](http://crash-reporter.tessel.io/crashes?fingerprint=0x223008ed)

rwaldron:
Duplicate: https://github.com/tessel/t2-cli/issues/1024

Issue #1024, opened 2016-10-07
Crash report: TypeError: Cannot read property 'stop' of undefined



tessel-crashbot:
```
TypeError: Cannot read property 'stop' of undefined
    at LAN.Scanner.stop (t2-cli/lib/lan-connection.js:206:15)
    at Object.stopScan (t2-cli/lib/lan-connection.js:159:13)
    at TesselSeeker.stop (t2-cli/lib/discover.js:125:24)
    at process.<anonymous> (t2-cli/lib/controller.js:156:16)
    at process.g (events.js:286:16)
    at emitNone (events.js:91:20)
    at process.emit (events.js:185:7)
    at processEmit (t2-cli/node_modules/signal-exit/index.js:146:32)
    at processEmit [as emit] (t2-cli/node_modules/usb/node_modules/node-pre-gyp/node_modules/npmlog/node_modules/gauge/node_modules/signal-exit/index.js:146:32)
    at Signal.wrap.onsignal (internal/process.js:201:44)
```

Full report is at [-0xc884c6e](http://crash-reporter.tessel.io/crashes?fingerprint=-0xc884c6e)

PR #1060, opened 2016-10-24
tests: make LAN.Scanner.prototype.stop calls safer. Fixes gh-1024, Fixes gh-1057

rwaldron:


HipsterBrown:
Thanks for this quick fix @rwaldron. Fixes https://github.com/tessel/t2-cli/issues/1024

PR #1077, opened 2016-11-02
Use "t2-progress" which is a throwaway fork of "progress" that applies the patch to terminate. Fixes gh-1073

rwaldron:
@tcr @HipsterBrown I've forked "node-progress" into "t2-progress" (with a bunch of nasty warnings), applied the patch necessary to fix the issue and published to npm. No new tests, because its the same usage on our end.


HipsterBrown:
Hoping your patch gets resolved by `node-progress` but thank you for the temporary solution in the meantime. LGTM :+1:

rwaldron:
Will merge when this is green

rwaldron:
Thanks @HipsterBrown

PR #1086, opened 2016-11-11
adding check to make sure d exists before trimming,

### LINKED ISSUE #771: t2 init -i will throw an error
The interactive option for initializing a JavaScript project (`t2 init -i`) throws an error:

```
➜  t2-demo t2 init -i
INFO Initializing new Tessel project for JavaScript...
INFO Created package.json.
ERR! [TypeError: Cannot read property 'trim' of undefined]
```

The error is being thrown [by this line](https://github.com/tessel/t2-cli/blob/55978f15d4ea21ddb95286e14ef8d33f14b263d5/resources/init-config.js#L28). Adding another check for a `falsy` value of `d` allows you to interactively create a package.json but the spinner awkwardly blocks the text being typed.

The fix for this bug would be to fix that check on `d` and fix the spinner blocking stdin text.

ducks:
stops spinner while interactivity initing to keep it from loading over text

fixes #771

### DIFF
--- a/lib/init/javascript.js
+++ b/lib/init/javascript.js
@@ -49,6 +49,9 @@ exportables.buildJSON = (npmConfig) => {
       promzardConfig = path.join(resources, 'init-config.js');
     }
 
+    // Stop spinner from being in our way while entering package.json data
+    log.spinner.stop();
+
     // Init promozard with appropriate config.
     var pz = new PZ(promzardConfig, ctx);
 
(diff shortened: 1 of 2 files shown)


rwaldron:
Since the init operations still have relatively low coverage, I don't think the coverage decrease should block this. There is also no clear way to add a simple test for this (which is actually due to the previous statement).

ducks:
Sweet. Thank you.

I thought I left a comment asking if there was any test to add but I guess I never added it.

Thanks again.

rwaldron:
No worries, I'm going to write some basic tests for this and push them up shortly. This way you can observe and have a good idea for next time :)

rwaldron:
Oh, actually... I have a better idea. I will add basic tests that should've existed and then you can make a PR that adds assertions for the changes!

ducks:
sounds great to me :)

Issue #1136, opened 2017-01-17
t2 push always uses rust remote compilation for rust project


mzgoddard:
In a rust project I can `t2 run <bin>` and compile locally. But when I try `t2 push <bin>` it compiles without the `--rustcc` flag both run and push have.

rwaldron:
I wonder if the options isn't getting forwarded?

rwaldron:
I found this, it was just an oversight when local compilation was added.

rwaldron:
@mzgoddard are you using t2-cli from an npm install? If so, then please `npm uninstall -g t2-cli`, clone this repo and do `npm install && npm link`, then checkout the branch I've pushed to verify the fix—thanks!

PR #1137, opened 2017-01-18
t2 push: Remove default rustcc url. Fixes gh-1136

rwaldron:


mzgoddard:
I can confirm this fixes `t2 run` for local rust builds. Thanks @rwaldron. (Also checked that --rustcc still remote compiles.)

rwaldron:
@mzgoddard thanks for the review!

PR #1139, opened 2017-01-19
Resolve access point creation with password

### LINKED ISSUE #1004: Access point doesn't work with password
I tried `t2 ap -n myssid -p mypassword` and myssid doesn't show up at all in access points list in my computer/mobile phone wifi list. If I just used `t2 ap -n otherssid' then otherssid is visible in wifi list and lets my computer connect to it.
How to make tessel 2 and cli work for an ap with password?
Environment: 

```
INFO Tessel [myt2] CLI version: 0.0.29
INFO Tessel [myt2] Firmware version: 0.0.16
INFO Tessel [myt2] Node version: 4.5.0
```

HipsterBrown:
Fixes #1004 

It turns out that the `t2 ap` command options changed in `bin/tessel-2.js` but `lib/tessel/access-point.js` was never updated to use `opts.password` instead of `opts.pass`. 

There was also a frequent occurrence of an error, `Error: udhcpc: sendto: Network is down`, that would cause an empty `wifi-iface` to be added to `etc/config/wireless`. This has been resolved by removing unnecessary `reconnectDnsmasq` and `reconnectDhcp` commands within the `commitAndClose` function. Those were added naively when this feature was originally built by me. I've confirmed that existing and factory-reset Tessels work without those commands. 

## Smoke Test:

- clone this repo
- `npm link path/to/clone`
- `t2 ap --ssid "ThisIsMyTessel" --password "HelloWorld"`

You should get the following result:

```
hipsterbrown:tessel-test $ t2 ap --ssid "ThisIsMyTessel" --password "HelloWorld"
INFO Connected to newtwo.
INFO Updated Access Point successfully. SSID:  ThisIsMyTessel, password HelloWorld, security mode: psk2
```

### DIFF
--- a/lib/tessel/access-point.js
+++ b/lib/tessel/access-point.js
@@ -14,14 +14,8 @@ function commitAndClose(tessel, status, resolve, reject) {
 
   var reconnectWifi = () => tessel.simpleExec(commands.reconnectWifi());
 
-  var reconnectDnsmasq = () => tessel.simpleExec(commands.reconnectDnsmasq());
-
-  var reconnectDhcp = () => tessel.simpleExec(commands.reconnectDhcp());
-
   return tessel.simpleExec(commands.commitWirelessCredentials())
     .then(reconnectWifi)
-    .then(reconnectDnsmasq)
-    .then(reconnectDhcp)
     .then((str) => log.info(status, str))
     .then(resolve)
     .catch(reject);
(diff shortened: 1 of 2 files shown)


rwaldron:
LGTM—merge when ready. This should go out in the patch release today.

rwaldron:
I think something is wrong with coveralls 0_o

HipsterBrown:
Yea, that's a bit odd. Merging now that Travis has passed.

PR #1167, opened 2017-03-13
Update error message for USB controller when the version of node is c…

crvineeth97:
…hanged.


lib/usb-connection.js
```diff
@@ -29,7 +29,7 @@ try {
   // var VENDOR_REQ_IN  = usb.LIBUSB_REQUEST_TYPE_VENDOR | usb.LIBUSB_RECIPIENT_DEVICE | usb.LIBUSB_ENDPOINT_IN;
 } catch (e) {
   haveusb = false;
-  log.error('WARNING: No usb controller found on this system.');
```
rwaldron: We should get rid of the word "WARNING". Now is a good time to do that

PR #1185, opened 2017-04-13
Modernize initial JS script

rwaldron:


rwaldron:
Also, it would be awesome if we updated all of the docs code examples as well.

PR #1191, opened 2017-04-17
Remove crash handlers for testing

rwaldron:
This will prevent development-time errors from being reported to the crash reporter. 


tikurahul:
One other option could be turning off the crash reporter using `t2 crash-reporter --off` before the beginning of a test run.

rwaldron:
@tikurahul that would mostly work: the obligatory "possibly a remote process error" logging in `CrashReporter.onerror` would still occur. I'm closing this anyway, because I want to think more about the bigger picture implications of toggling our global handlers.

Issue #1219, opened 2017-06-01
Crash report: Error: EACCES: permission denied, open '/.tessel/id_rsa'



tessel-crashbot:
```
Error: EACCES: permission denied, open '/.tessel/id_rsa'
    at Error (native)
    at Object.fs.openSync (fs.js:641:18)
    at Object.fs.readFileSync (fs.js:509:33)
    at new LAN.Connection (t2-cli/lib/lan-connection.js:51:50)
    at browser.on.err (t2-cli/lib/lan-connection.js:197:33)
    at emitOne (events.js:96:13)
    at emit (events.js:188:7)
    at internal.onMessage (t2-cli/node_modules/mdns-js/lib/browser.js:108:12)
    at emitThree (events.js:116:13)
    at emit (events.js:194:7)
```

Full report is at [0x53917aca](http://crash-reporter.tessel.io/crashes?fingerprint=0x53917aca)

rwaldron:
I suspect the user does not have permissions set correctly.

rwaldron:
I'm going to explore adding explicit handling for this

PR #1222, opened 2017-06-01
Detect id_rsa permissions problems and bail out early. Fixes gh-1219

rwaldron:

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -1,6 +1,7 @@
 'use strict';
 
 // System Objects
+var fs = require('fs');
 var cp = require('child_process');
 var util = require('util');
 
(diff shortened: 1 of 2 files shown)


lib/lan-connection.js
```diff
@@ -48,7 +47,7 @@ LAN.Connection = function(opts) {
   this.ssh = undefined;
 
   if (Tessel.isProvisioned()) {
-    this.auth.privateKey = opts.privateKey || fs.readFileSync(path.join(Tessel.LOCAL_AUTH_KEY));
+    this.auth.privateKey = opts.privateKey || fs.readFileSync(Tessel.LOCAL_AUTH_KEY);
```
HarleyKwyn: Should we write a common module that reads the auth key and throws the EACCESS error even in this case?. Then we can use it in both `lib/controller.js` and here.

rwaldron: If the permissions problem exists, no operation will ever reach this code.

lib/controller.js
```diff
@@ -28,6 +29,17 @@ var responses = {
   auth: 'No Tessels Found.'
 };
 
+// Alert the user that they have a permissions problem.
+try {
+  fs.readFileSync(Tessel.LOCAL_AUTH_KEY);
```
LinusU: Could probably be `fs.accessSync` which wouldn't read the file into memory

rwaldron: Oh, good call. Let me try that

PR #1228, opened 2017-06-05
cli: update display color of all occurences of tessel.name.

rwaldron:


rwaldron:
It's impossible to stub any of the functions on `colors` object, as they are all defined as getter accessors, with `configurable` set to `false`.

```
{ get: [Function: get],
  set: undefined,
  enumerable: false,
  configurable: false }
```

After this patch, I want to eliminate `colors` and use `chalk` instead.

PR #1260, opened 2017-06-26
Check version and sha specifically, instead of build list string

HipsterBrown:
This PR should resolve testing the output from `controller.printAvailableUpdates()`, which uses `Date().toLocaleString` internally and is affected by the timezone of the CI environment. Instead, we now test that the version number is in the beginning and the correct SHA is included.

### DIFF
--- a/test/unit/update.js
+++ b/test/unit/update.js
@@ -33,7 +33,8 @@ exports['controller.update'] = {
       return Promise.resolve('9a85c84f5a03c715908921baaaa9e7397985bc7f');
     });
 
-    this.requestBuildList = this.sandbox.stub(updates, 'requestBuildList', () => Promise.resolve(builds));
+    // resolve requestBuildList with a copy of the stubbed data, otherwise the original stub is mutated when sorted
+    this.requestBuildList = this.sandbox.stub(updates, 'requestBuildList', () => Promise.resolve(builds.slice()));
 
     this.updateTesselWithVersion = this.sandbox.spy(controller, 'updateTesselWithVersion');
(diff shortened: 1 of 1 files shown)


tessel-highfive:
r? @hipsterbrown

test/unit/update.js
```diff
@@ -33,7 +33,8 @@ exports['controller.update'] = {
       return Promise.resolve('9a85c84f5a03c715908921baaaa9e7397985bc7f');
     });
 
-    this.requestBuildList = this.sandbox.stub(updates, 'requestBuildList', () => Promise.resolve(builds));
+    // resolve requestBuildList with a copy of the stubbed data, otherwise the original stub is mutated when sorted
+    this.requestBuildList = this.sandbox.stub(updates, 'requestBuildList', () => Promise.resolve(builds.slice()));
```
rwaldron: Nice

test/unit/update.js
```diff
@@ -65,14 +66,14 @@ exports['controller.update'] = {
...
-        );
-        test.equal(
-          this.basic.getCall(1).args[0],
-          'Version: 0.0.0\nPublished: 5/17/2017, 10:21:57 PM\nac4d8d8a5bfd671f7f174c2eaa258856bd82fe29\n'
-        );
+        const firstCall = this.basic.getCall(0).args[0];
+        test.equal(firstCall.startsWith('Version: 0.0.1'), true);
+        test.equal(firstCall.includes(builds[1].sha), true);
+
+        const secondCall = this.basic.getCall(1).args[0];
```
rwaldron: Technically, this is not the "secondCall", it's the second call's args. I'm not going to block on that, so I'll leave it up to you if you want to change it.

PR #1272, opened 2017-07-14
t2 version should return at least the CLI version

HipsterBrown:
Right now when `t2 version` is run with no Tessels connected via USB or LAN, the command will end with `WARN No Tessels Found.` or `WARN No Authorizes Tessels Found.`. This patch allows the CLI version to be displayed if no Tessels are found. 

New output:

```
hipsterbrown:t2-cli (t2-version-cli-only) $ t2 version
INFO Tessel Environment Versions:
INFO t2-cli: 0.1.5
INFO No Authorized Tessels Found.
hipsterbrown:t2-cli (t2-version-cli-only) $
```

### DIFF
--- a/test/unit/controller.js
+++ b/test/unit/controller.js
@@ -593,6 +593,42 @@ exports['controller.tesselEnvVersions'] = {
       });
   },
 
+  requestVersionWithoutTessel(test) {
+    test.expect(7);
+
+    const message = 'No Tessels Found.';
+    
+    this.standardTesselCommand.restore();
+    this.standardTesselCommand = this.sandbox.stub(controller, 'standardTesselCommand', () => {
+      return Promise.reject(message);
+    });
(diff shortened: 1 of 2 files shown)


tessel-highfive:
r? @tcr

test/unit/controller.js
```diff
@@ -593,6 +593,42 @@ exports['controller.tesselEnvVersions'] = {
...
 
+  requestVersionWithoutTessel(test) {
+    test.expect(7);
+
+    const message = 'No Tessels Found.';
+    
+    this.standardTesselCommand.restore();
+    this.standardTesselCommand = this.sandbox.stub(controller, 'standardTesselCommand', () => {
+      return Promise.reject(message);
+    });
```
rwaldron: Nice

PR #1297, opened 2017-08-03
Call Tessel.setWiFiState with updated arguments

HipsterBrown:
Currently, running `t2 wifi --off` returns the following output:

```
hipsterbrown:tessel-test (master) $ t2 wifi --off
INFO Connected to j5.
ERR! Error: Missing Wifi State: (empty).
```

This is due to the controller command not calling `tessel.setWiFiState` with the recently updated parameters.


rwaldron:
Merged ddba290f86e80104121a276ea93ca2ea8c3a03f3

Issue #1071, opened 2016-10-30
latest t2-cli results in ERR! WARNING: No usb controller found on this system.


sbromberger:
Environment: OSX 10.12.1, node.js 7.0.0

To reproduce: clone the repo, then execute `npm link --local`

Results:

```

> t2-cli@0.1.1 postinstall /Users/me/dev/tessel/t2-cli
> t2 install drivers --loglevel=error || true; t2 install homedir --loglevel=error || true;

ERR! WARNING: No usb controller found on this system.
```

To recover, `kill -9` the process.

HipsterBrown:
@sbromberger We currently support the [Node.js LTS version](https://nodejs.org/en/download/), which currently 4.x.x and the 6.x.x, for `t2-cli`. I'm not sure `node-usb` has support for the latest version of Node yet. When switching versions, be sure to reinstall and rebuild `t2-cli` due to the native bindings required for `node-usb`.

sbromberger:
Thanks - I figured it'd be a 7.0.0 thing - but the maddening thing is that even though I'm trying to access it via the LAN interface, it's still hanging on USB. Is there a way to say "ignore USB, use LAN"?

Frijol:
@sbromberger yes, `--lan` specifies LAN usage. But I'm not sure if it will keep you from hitting the error message anyway.

@HipsterBrown I think the "unsupported USB" message is the most common question we get. Do you know if it's possible to build a more helpful error message for this eventuality into our CLI? Have we investigated this?

sbromberger:
@Frijol one hits the error message anyway, which causes a spike to 100% CPU (and fans), and a nonresponsive terminal - the only way to stop the runaway node process is to `kill -9` it.

HipsterBrown:
@Frijol Yea the message comes from [`lib/usb-connection.js`](https://github.com/tessel/t2-cli/blob/2b08c0a5671308ba9559f1003f95056065100ea7/lib/usb-connection.js#L32). Any ideas on a better error message?

rwaldron:
I have an idea of how we might get around this and allow `--lan` commands to be truly free of USB. I will post back when I've learned more.

rwaldron:
```
rwaldron in ~/clonez/t2-cli on 1071*
$ t2 list
INFO Searching for nearby Tessels...
ERR! WARNING: No usb controller found on this system.
ERR! TypeError: Cannot read property 'on' of undefined
[...]

PR #1301, opened 2017-08-07
Updates warning with instructions to rebuild usb drivers without reinstalling t2-cli

HipsterBrown:
This PR proposes a new solution to the "WARNING: Please run npm install -g t2-cli to compile USB drivers for the newer version of node" issue. (More details [here](https://github.com/tessel/t2-cli/issues/1071))

The error message will now show: **"ERR! Node version mismatch for USB drivers. Please run "npm rebuild --update-binary usb" to rebuild the USB drivers for your version of node"**

<del> `t2 rebuild` is new command that programmatically calls `npm rebuild usb --update-binary` so users don't have to completely reinstall the CLI. </del>

Edit: After a conversation with @rwaldron, I learned that `npm rebuild --update-binary usb` can be called from anywhere. So instead of adding this responsibility to this CLI, we can simply inform the user to call the `npm rebuild` command manually. 

Preview:

```
hipsterbrown:tessel-test $ t2 list
ERR! Node version mismatch for USB drivers.
INFO To correct this issue, please run the following command:
INFO
INFO   npm rebuild --update-binary usb
INFO
INFO This will rebuild the USB drivers for your version of Node.js
hipsterbrown:tessel-test $ npm rebuild --update-binary usb

> usb@1.2.0 install /Users/hipsterbrown/Projects/tessel-test/node_modules/t2-cli/node_modules/usb
> node-pre-gyp install --fallback-to-build

[usb] Success: "/Users/hipsterbrown/Projects/t2-cli/node_modules/usb/src/binding/usb_bindings.node" is installed via remote
usb@1.2.0 /Users/hipsterbrown/Projects/tessel-test/node_modules/t2-cli/node_modules/usb
hipsterbrown:tessel-test $ t2 list
WARN No Tessels Found.
hipsterbrown:tessel-test $
```

### DIFF
--- a/lib/controller.js
+++ b/lib/controller.js
@@ -12,14 +12,13 @@ var semver = require('semver');
 
 // Internal
 var CrashReporter = require('./crash-reporter');
-var discover = require('./discover');
 var init = require('./init');
 var installer = require('./installer');
 var log = require('./log');
 var Menu = require('./menu');
 var updates = require('./update-fetch');
-var provision = require('./tessel/provision');
 var Tessel = require('./tessel/tessel');
+var provision = require('./tessel/provision');
 
 var controller = {
   // This will be assigned with the Tessel that is found or selected.
@@ -86,6 +85,8 @@ controller.setupLocal = function(opts) {
 
 
 Tessel.list = function(opts) {
+  var discover = require('./discover');
+
   return new Promise(function(resolve, reject) {
     // Grab all attached Tessels
     log.info('Searching for nearby Tessels...');
@@ -181,6 +182,8 @@ Tessel.list = function(opts) {
 };
 
 Tessel.get = function(options) {
+  var discover = require('./discover');
+
   return new Promise((resolve, reject) => {
     log.info('Looking for your Tessel...');
     // Collection variable as more Tessels are found
--- a/lib/usb-connection.js
+++ b/lib/usb-connection.js
@@ -7,7 +7,7 @@ var Duplex = stream.Duplex;
 var Emitter = events.EventEmitter;
 
 // Third Party Dependencies
-// ...
+const tags = require('common-tags');
 
 // Internal
 var DFU = require('./dfu');
@@ -22,14 +22,26 @@ function debugCommands(message) {
   log.debug(`(commands:usb) ${message}`);
 }
 
-var haveusb = true;
+var isUSBAvailable = true;
 try {
   var usb = require('usb');
   var VENDOR_REQ_OUT = usb.LIBUSB_REQUEST_TYPE_VENDOR | usb.LIBUSB_RECIPIENT_DEVICE | usb.LIBUSB_ENDPOINT_OUT;
   // var VENDOR_REQ_IN  = usb.LIBUSB_REQUEST_TYPE_VENDOR | usb.LIBUSB_RECIPIENT_DEVICE | usb.LIBUSB_ENDPOINT_IN;
-} catch (e) {
-  haveusb = false;
-  log.error('WARNING: No USB controller found on this system. Please run npm install -g t2-cli to compile USB drivers for your version of node');
+} catch (error) {
+  isUSBAvailable = false;
+
+  // do not exit the process during tests because usb is not needed to run them
+  if (!global.IS_TEST_ENV) {
+    log.error('Node version mismatch for USB drivers.');
+    log.info(tags.stripIndent`
+      To correct this issue, please run the following command:
+
+        npm rebuild --update-binary usb
+
+      This will rebuild the USB drivers for your version of Node.js
+    `);
+    process.exit(1);
+  }
 }
 
 var Daemon = require('./usb/usb-daemon');
@@ -408,7 +420,7 @@ USB.Scanner.prototype.start = function() {
     }
   };
 
-  if (haveusb) {
+  if (isUSBAvailable) {
     usb.getDeviceList().forEach(deviceInspector);
 
     usb.on('attach', deviceInspector);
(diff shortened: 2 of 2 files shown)


Frijol:
Cool! I'd suggest a pretty explicit command name like "usb-driver-rebuild" since it's used infrequently and otherwise might be confused with rebuilding a tessel project or similar

HipsterBrown:
@Frijol I've updated the description with a new direction for this PR.

lib/usb-connection.js
```diff
@@ -22,14 +22,19 @@ function debugCommands(message) {
...
   // var VENDOR_REQ_IN  = usb.LIBUSB_REQUEST_TYPE_VENDOR | usb.LIBUSB_RECIPIENT_DEVICE | usb.LIBUSB_ENDPOINT_IN;
-} catch (e) {
-  haveusb = false;
-  log.error('WARNING: No USB controller found on this system. Please run npm install -g t2-cli to compile USB drivers for your version of node');
+} catch (error) {
+  isUSBAvailable = false;
+
+  // do not exit the process during tests because usb is not needed to run them
+  if (!global.IS_TEST_ENV) {
+    log.error('Node version mismatch for USB drivers. Please run "npm rebuild --update-binary usb" to rebuild the USB drivers for your version of node');
```
rwaldron: Let's take this opportunity to improve this for our users. Something like this: 

```js
const tags = require('common-tags');
```

```js
log.error('Node version mismatch for USB drivers.');
log.info(tags.stripIndent `
  To correct this issue, run the following command: 

  npm rebuild --update-binary usb

  This will rebuild the USB drivers for your version of Node.js
`);
```

lib/controller.js
```diff
@@ -181,6 +182,8 @@ Tessel.list = function(opts) {
 };
 
 Tessel.get = function(options) {
+  var discover = require('./discover');
+
```
rwaldron: Are these changes still necessary?

lib/init/javascript.js
```diff
@@ -21,10 +21,11 @@ exportables.meta = {
   keywords: ['javascript', 'js']
 };
 
-exportables.loadNpm = () => {
+exportables.loadNpm = (options) => {
+  options = options || {};
   // You have to load npm in order to use it programatically
   return new Promise((resolve, reject) => {
-    NPM.load((error, npm) => {
+    NPM.load(options, (error, npm) => {
```
rwaldron: Are these changes still necessary?

HipsterBrown: This part is no longer necessary. I can revert those changes.

lib/tessel/deployment/javascript.js
```diff
@@ -34,7 +34,7 @@ var Tessel = require('../tessel');
 var binaryModulesUsed = new Map();
 
 const BINARY_SERVER_ROOT = `http://${remote.PACKAGES_HOSTNAME}/npm/`;
-const BINARY_CACHE_PATH = path.join(Tessel.LOCAL_AUTH_PATH, 'binaries');
+const BINARY_CACHE_PATH = path.join(Tessel.LOCAL_AUTH_PATH || '', 'binaries');
```
rwaldron: Are these changes still necessary?

test/unit/init.js
```diff
@@ -444,7 +444,7 @@ exports['init --lang=javascript'] = {
     test.expect(2);
 
     this.npmLoad.restore();
-    this.npmLoad = this.sandbox.stub(npm, 'load', (callback) => {
+    this.npmLoad = this.sandbox.stub(npm, 'load', (options, callback) => {
```
rwaldron: Are these changes still necessary?

lib/controller.js
HipsterBrown: I believe so, yes. When the `discover` module is always required by this file, any CLI action can be blocked by the USB driver error, including things like `t2 init`. Requiring the `discover` only when needed, _should_ lead to a nicer user experience.

lib/tessel/deployment/javascript.js
HipsterBrown: I don't believe so. I'll revert those after I run the tests.

test/unit/init.js
HipsterBrown: Nope, I'll revert that.

lib/controller.js
rwaldron: Ok, but I think we should revisit this some in the future

lib/controller.js
```diff
@@ -12,14 +12,13 @@ var semver = require('semver');
...
 var CrashReporter = require('./crash-reporter');
-var discover = require('./discover');
 var init = require('./init');
 var installer = require('./installer');
 var log = require('./log');
 var Menu = require('./menu');
 var updates = require('./update-fetch');
-var provision = require('./tessel/provision');
 var Tessel = require('./tessel/tessel');
+var provision = require('./tessel/provision');
```
HipsterBrown: Reordered this so that certain constants, i.e. `Tessel.LOCAL_AUTH_PATH`, can be added correctly.

HipsterBrown:
The description has been updated to preview the new formatted error and info logs.

lib/usb-connection.js
```diff
@@ -22,14 +22,26 @@ function debugCommands(message) {
...
+    log.error('Node version mismatch for USB drivers.');
+    log.info(tags.stripIndent`
+      To correct this issue, please run the following command:
+
+        npm rebuild --update-binary usb
+
+      This will rebuild the USB drivers for your version of Node.js
+    `);
+    process.exit(1);
+  }
```
rwaldron: 👍

PR #1307, opened 2017-08-10
t2 install homedir no overwrite

rwaldron:
This has a hard dependency on [greenkeeper/initial](https://github.com/tessel/t2-cli/pull/1305/files) and will require rebasing once that lands.


package.json
```diff
@@ -45,49 +45,54 @@
...
+    "glob": "^7.1.2",
+    "inquirer": "^3.2.1",
     "is-root": "^1.0.0",
     "mdns-js": "^0.5.0",
     "minimatch": "^3.0.0",
-    "node-rsa": "^0.2.26",
+    "node-rsa": "^0.4.2",
     "nomnom": "^1.8.1",
-    "npm": "^3.10.10",
+    "npm": "^5.3.0",
```
huseyinkozan: Can this cause problems ?

rwaldron: It may, but I haven't finished the greenkeeper testing so I have no idea yet :)

PR #1348, opened 2017-10-05
Add alternative compression with uglify-js 3

rwaldron:
Similar to the way we used to do it with different compressors, but now using uglify-es and uglify-js


tessel-highfive:
r? @hipsterbrown

lib/tessel/deployment/lists/javascript.js
```diff
@@ -54,7 +54,14 @@ module.exports = {
     },
   },
   compressionBlacklist: {
-    bindings: true,
-    base64id: true,
```
HipsterBrown: Do we still need this in the blacklist?

rwaldron: Yes, it's being used for "supports-color" right now

rwaldron: Oh, hrm... I think maybe it doesn't need to?

Issue #377, opened 2015-10-17
Somtimes during update, the file download stalls without exiting


aricearice:
When I try to update, sometimes the download stalls without exiting the process.

```
$ t2 update -f 
INFO Looking for your Tessel...
INFO Connected to oogie-boogie over LAN
INFO Updating oogie-boogie to latest version (0.0.4)...
INFO Beginning update download. This could take a couple minutes..
  downloading [==============      ] 70% 0.3s remaining
```

**Expected**
The T2 should error out of the update if the download stalls for too long.

johnnyman727:
Ah yes, we definitely should. Good find!

johnnyman727:
What we need for this issue is a timeout that gets reset each time we get a packet of data from the build server and cleared when the update finishes downloading. I think a stall time of ~10 seconds is reasonable.

Frijol:
This file looks like the right place to start: https://github.com/tessel/t2-cli/blob/master/lib/update-fetch.js

PR #1354, opened 2017-10-21
WIP set a timer for each packet of data #377

alwasa0b:
I am not entirely sure this would fix the issue so need some feedback. I also not sure what should happen when the timeout is reached.

### DIFF
--- a/lib/update-fetch.js
+++ b/lib/update-fetch.js
@@ -163,9 +163,14 @@ exportables.downloadTgz = function(tgzUrl, fileMap) {
       return resolve(files);
     });
 
-
     remote.ifReachable(remote.BUILDS_HOSTNAME).then(() => {
+
       var req = request.get(tgzUrl);
+
+      var startTimeout = () => setTimeout(() => {
+          reject(new Error('The attempt to download a build has timed out. Check your network connection and try again.'));
+        }, 10000);
+      var timeout = startTimeout();
 
       // When we receive the response
       req.on('response', (res) => {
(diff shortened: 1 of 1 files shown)


tessel-highfive:
Thanks for the pull request, and welcome! The Rust team is excited to review your changes, and you should hear from @tcr (or someone else) soon.

If any changes to this PR are deemed necessary, please add them as extra commits. This ensures that the reviewer can see what has changed since they last reviewed the code. Due to the way GitHub handles out-of-date commits, this should also make it reasonably obvious what issues have or haven't been addressed. Large or tricky changes may require several passes of review and changes.

Please see [the contribution instructions](https://github.com/tessel/project/blob/master/CONTRIBUTING.md) for more information.

HipsterBrown:
Thanks for opening a PR @alwasa0b. To what issue are you referring? Can you link to it?

alwasa0b:
I thought I put the right issue# on the PR. issue #337 [https://github.com/tessel/t2-cli/issues/377](url)

lib/update-fetch.js
```diff
@@ -163,10 +163,13 @@ exportables.downloadTgz = function(tgzUrl, fileMap) {
       return resolve(files);
     });
 
-
     remote.ifReachable(remote.BUILDS_HOSTNAME).then(() => {
+
       var req = request.get(tgzUrl);
 
+      var startTimeout = () => setTimeout(() => {reject("what should happen on timeout reached");}, 10000);
+      var timeout = startTimeout();
```
rwaldron: Two minor fixes: 

- Travis is telling us that you need to use single quotes ;)
- Reject with: 
  ```js
  reject(new Error('The attempt to download a build has timed out. Check your network connection and try again.'));
  ```

rwaldron:
This approach looks good to me, just needs to have the rejection argument fixed up. We can work on tests here or in a follow up.

rwaldron:
I'm not going to press hard for tests on this patch because testing this appropriately is not immediately clear to me. I also suspect that's why there are no tests included? If so, then that's ok, making this better for the end users is more important than grinding out a complicated test with several layers of async mocking :)

PR #1373, opened 2017-11-26
Upgrade uglify-es

rexxars:
I'm running into a bunch of issues when trying to run my app on the tessel. After a bunch of digging, I found out that my code was being run through uglify-es. 

Unfortunately, the version that is bundled with the t2-cli does weird things to my code. For instance:
```js
const contentLength = response.headers['content-length'];
const length = contentLength && Number(contentLength);
```

turns into

```js
var u = (t.headers["content-length"], contentLength && +contentLength);
```

...which results in `ReferenceError: contentLength is not defined`

I upgraded `uglify-es` to the latest version, and all the build problems went away.

I could of course pass `--compress=false`  on every build, but as a library author, it would be nice not having to ask the users to do so.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "tar": "^2.1.1",
     "tar-fs": "^1.13.2",
     "tar-stream": "^1.3.0",
-    "uglify-es": "3.1.3",
+    "uglify-es": "3.2.0",
     "uglify-js": "3.1.3",
     "unbzip2-stream": "^1.0.10",
     "update-notifier": "^1.0.2",
--- a/test/unit/deployment/javascript.js
+++ b/test/unit/deployment/javascript.js
@@ -1279,7 +1279,7 @@ exports['deployment.js.tarBundle'] = {
 
       const minified = this.compress.lastCall.returnValue;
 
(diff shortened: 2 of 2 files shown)


tessel-highfive:
Thanks for the pull request, and welcome! The Rust team is excited to review your changes, and you should hear from @tcr (or someone else) soon.

If any changes to this PR are deemed necessary, please add them as extra commits. This ensures that the reviewer can see what has changed since they last reviewed the code. Due to the way GitHub handles out-of-date commits, this should also make it reasonably obvious what issues have or haven't been addressed. Large or tricky changes may require several passes of review and changes.

Please see [the contribution instructions](https://github.com/tessel/project/blob/master/CONTRIBUTING.md) for more information.

rexxars:
Hum. Looks like the new version uses template literals when minifying, which breaks the test.

rexxars:
Oh, my bad, it looks like it's the other way around - it's changing template literals to regular quotes. I've updated the failing test to reflect the new uglify behavior, please have another look :smiley:

rwaldron:
@HipsterBrown can you do a release? Thanks!!

HipsterBrown:
Fix released as `t2-cli@v0.1.17` https://github.com/tessel/t2-cli/releases/tag/v0.1.17

Thanks for contributing @rexxars! :clap:

PR #1508, opened 2018-07-24
Tessel: Make "displayName" property for wrapping colorized output of Tessel name

rwaldron:


lib/controller.js
```diff
@@ -252,7 +252,7 @@ Tessel.get = function(options) {
                         // show "(not authorized)" if there is a lanConnection
                         /* istanbul ignore next */
                         const authorization = isAuthorized || !!tessel.usbConnection ? '' : '(not authorized)';
-                        const display = `\t${tessel.connection.connectionType}\t${colors.magenta(tessel.name)}\t${authorization}`;
+                        const display = `\t${tessel.connection.connectionType}\t${tessel.name}\t${authorization}`;
```
LinusU: Did you mean to change to `.displayName` here as well?

rwaldron: Hm... yep! I'll fix that.

rwaldron:
@LinusU thanks! ❤️

PR #1509, opened 2018-07-24
Controller: exit process immediately after successful update.

rwaldron:
Blocked by #1508


rwaldron:
@LinusU thanks again! ❤️

PR #1579, opened 2018-11-27
Deploy: ignore errors produced by "dd" when creating the remote shell script for push deployment

rwaldron:
For some reason that I cannot figure out, creating the shell script with `openStdinToFile(...)` was failing to "close". Digging in I discovered that this output was being produced: 

```
0+1 records in
0+1 records out
```

Despite that, the file was being created and written to as expected. 

Other places that openStdinToFile is used:

- t2 rename: works as expected
- t2 update: works as expected

To verify, do a `t2 push <file>` of some project that is known to work correctly with `t2 run <file>`

### DIFF
--- a/lib/tessel/deploy.js
+++ b/lib/tessel/deploy.js
@@ -445,8 +445,13 @@ exportables.createShellScript = function(tessel, opts) {
       if (error) {
         return reject(error);
       }
-      // When the remote process finishes
-      remoteProcess.once('close', () => {
+
+      tessel.receive(remoteProcess, (error, received) => {
+        /* istanbul ignore if */
+        if (error && !error.message.includes('records in')) {
+          return reject(error);
+        }
+
         // Set the perimissions on the file to be executable
         tessel.connection.exec(commands.chmod('+x', PUSH_START_SH_SCRIPT), (error, remoteProcess) => {
           /* istanbul ignore if */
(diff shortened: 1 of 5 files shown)


lib/tessel/deploy.js
```diff
@@ -445,16 +445,21 @@ exportables.createShellScript = function(tessel, opts) {
       if (error) {
         return reject(error);
       }
-      // When the remote process finishes
-      remoteProcess.once('close', () => {
+
+      tessel.receive(remoteProcess, (error, received) => {
+        /* istanbul ignore if */
+        if (error && !error.message.includes('records in')) {
```
rwaldron: Yes, I know this is awful.

rwaldron:
image

HipsterBrown:
Looks like there are some linting errors. Looks good otherwise. :+1:

rwaldron:
> Does it set an exit code?

I thought to myself "ya know... I'm not sure, let's find out...". Turns out SSH2 does offer an exitCode if such exists, however it appears to be `undefined` in this case :|

rwaldron:
I'm still working on revising this

rwaldron:
@reconbot @HipsterBrown I'm using an exit code now. If there's an error, but an exit code = 0, then we'll ignore it.

Issue #541, opened 2016-02-02
[Feature] Enable 3G Dongles


johnnyman727:
To get 3G Dongles working, we'll need to update some dependencies on the device, add a udev rule, and finally add logic to the CLI to write the appropriate options to the `/etc/config/network` file. This issue is about that final piece which needs to be in t2-cli.

To see how to do this, check out https://wiki.openwrt.org/doc/recipes/3gdongle.

johnnyman727:
This would still be a great feature and we can't use 3G dongles smoothly until it's added. If anyone is interested in contributing and needs more help, feel free to ping me here.

shaochuancs:
Hi, @johnnyman727 , this would be a great feature. Looking forward that we can use 3G dongle automatically.

I opened [https://wiki.openwrt.org/doc/recipes/3gdongle](https://wiki.openwrt.org/doc/recipes/3gdongle) but feel completely confused (I'm a web developer and not familiar with OpenWRT/hardware/system programming). Does it mean we need to ssh to Tessel machine and run several commands there?

Thanks

johnnyman727:
Hi @shaochuancs, thanks for getting involved!

> I opened https://wiki.openwrt.org/doc/recipes/3gdongle but feel completely confused (I'm a web developer and not familiar with OpenWRT/hardware/system programming)

No worries - I felt the same way when I first started perusing through OpenWRT docs (and still do sometimes). Feel free to ask me any questions that come up.

> Does it mean we need to ssh to Tessel machine and run several commands there?

Yes, that's exactly right. If your Tessel is connected to your wifi, you can get a shell with `t2 root`. That's the preferred way to work with Tessel's console (it uses `ssh` internally). If not, you can still get a shell over USB but you'll need to use a tool like `dterm` or `screen` (you'd use the command like `screen /dev/usbmodem1422`).

Typically, the way we build things for the CLI, is we'll use the shell to figure out what commands we need to run to get some feature to work. Once we have a list of commands to run, it becomes relatively simple to automate those commands with the CLI. 

As an example, [here is how we enable access points](https://github.com/tessel/t2-cli/blob/7d95332175d3a5d26f5f246d9166d913d1523eac/lib/tessel/access-point.js#L122-L143). You can see that we have a lot of commands that use:
```
tessel.simpleExec
```
This is basically a wrapper over:
```
tessel.connection.exec(SOME_COMMAND, () => {});
```
[...]

PR #1597, opened 2019-01-06
Implemented 3G USB dongle setup CLI command

bart:


bart:
Resolves #541

lib/tessel/setup-3g.js
```diff
@@ -0,0 +1,119 @@
+var Tessel = require('./tessel');
```
HipsterBrown: Can you update the use of `var` in this file to `const`? Those variables are never reassigned so we can optimize with the use of `const`.

lib/tessel/setup-3g.js
```diff
@@ -0,0 +1,119 @@
...
+  var status = options.status;
+
+  if (config) {
+	return this.simpleExec(['uci', 'get', 'network.wan']).then(() => {
+		// If dongle already exists remove it and store new configuration
+		return this.simpleExec(['uci', 'delete', 'network.wan']).then(() => {
+			return this.configure3gDongle(apn, dialnumber, username, password);
+		});
+	}).catch(() => {
+		return this.configure3gDongle(apn. dialnumber, username, password);
```
HipsterBrown: There appears to be a typo here. 

```suggestion
		return this.configure3gDongle(apn, dialnumber, username, password);
```

lib/tessel/setup-3g.js
```diff
@@ -0,0 +1,119 @@
...
+  var apn = options.apn;
+  var dialnumber = options.dialnumber;
+  var username = options.username;
+  var password = options.password;
+  var config = options.config;
+  var on = options.on;
+  var off = options.off;
+  var status = options.status;
+
+  if (config) {
```
HipsterBrown: Similar to the `wifi` and `ap` commands, I would expect configuration to be the default mode of the `3g` command. For example:

```
t2 3g --apn="" --dialnumber="" --username="" --password=""
```

I would expect that command alone to configure the Tessel to connect to that 3g network and be enabled. 

If none of those configuration options are passed in, i.e. `t2 3g`, I would expect to see the current configuration and status of 3g mode. Just like if you use `t2 wifi`, the cli will return the current details of the Tessel's wifi connection (if it is connected).

